PR #1, opened 2020-06-23
Added fixes for building Dagster RDS instance

blarghmatey:
- Fixed stack exports and component data access
- Fixed parameter passing for RDS config instance
- Fixed bug in RDS helper function


src/ol_infrastructure/components/aws/database.py
```diff
@@ -9,30 +9,64 @@
...
-from typing import Text, Dict
+from enum import Enum
+from typing import Dict, List, Optional, Text, Union
 
 import pulumi
 from pulumi_aws import rds
-from pydantic import validator
+from pydantic import BaseModel, PositiveInt, SecretStr, conint, validator
 
+from ol_infrastructure.lib.aws.rds_helper import (
```
shaidar, line 19: Why the parentheses?

src/ol_infrastructure/components/aws/database.py
```diff
@@ -9,30 +9,64 @@
...
+)
 from ol_infrastructure.lib.ol_types import AWSBase
-from ol_infrastructure.lib.aws.rds_helper import db_engines, parameter_group_family
 
-# ensure db subnet group
-# create parameter group
 # manage backup policy
 # generate or retrieve password in config
 # create DB security group
 # storage encrypted
```
shaidar, line 28: Those comments feel a bit out of place

src/ol_infrastructure/components/aws/database.py
blarghmatey: Because of line length. I'm using isort to format import statements.

src/ol_infrastructure/components/aws/database.py
blarghmatey: Yup, those are out of date. Removing...

src/ol_infrastructure/components/aws/database.py
```diff
@@ -9,30 +9,64 @@
...
+    subnet_group_name: Union[Text, pulumi.Output[str]]
+    security_groups: List[rds.SecurityGroup]
+    backup_days: conint(ge=0, le=35, strict=True) = 30
+    db_name: Optional[Text] = None  # The name of the database schema to create
+    instance_size: Text = 'db.m5.large'
+    is_public: bool = False
+    max_storage: Optional[PositiveInt] = None  # Set to allow for storage autoscaling
     multi_az: bool = True
+    prevent_delete: bool = True
     public_access: bool = False
```
shaidar, line 61: This depends on what the db will be used by, no?

src/ol_infrastructure/components/aws/database.py
```diff
@@ -9,30 +9,64 @@
...
+    read_replica: Optional[Dict] = None
+    storage: PositiveInt = PositiveInt(50)
+    storage_type: StorageType = StorageType.ssd
+    username: Text = 'oldevops'
+    read_replica: Optional[OLReplicaDBConfig] = None
```
shaidar, line 66: Does this allow having the same key twice? (line 62)

src/ol_infrastructure/components/aws/database.py
```diff
@@ -52,7 +85,100 @@ def is_valid_version(cls: 'OLDBConfig', engine_version: Text, values: Dict) -> T
...
+        {'name': 'timezone', 'value': 'UTC'},
+        {'name': 'rds.force_ssl', 'value': 1}
+    ]
+
+
+class OLMariaDBConfig(OLDBConfig):
+    engine: Text = 'mariadb'
+    engine_version: Text = '10.4.8'
+    port: PositiveInt = PositiveInt(3306)
+    parameter_overrides: List[Dict[Text, Union[Text, bool, int, float]]] = [
```
shaidar, line 105: Is ssl on by default in here unlike Postgres?

src/ol_infrastructure/components/aws/database.py
```diff
@@ -52,7 +85,100 @@ def is_valid_version(cls: 'OLDBConfig', engine_version: Text, values: Dict) -> T
...
+        {'name': 'character_set_server', 'value': 'utf8mb4'},
+        {'name': 'time_zone', 'value': 'UTC'}
+    ]
+
+
+class OLAmazonDB(pulumi.ComponentResource):
 
     def __init__(self, db_config: OLDBConfig, opts: pulumi.ResourceOptions = None):
-        super().__init__('ol:infrastructure:AmazonPostgresDB', db_config.db_name, None, opts)
+        super().__init__('ol:infrastructure:aws:rds:PostgresDB', db_config.instance_name, None, opts)
```
shaidar, line 119: Where is `PostgresDB` defined?

src/ol_infrastructure/components/aws/database.py
blarghmatey: Yes, but it can be overridden. This is just a default.

src/ol_infrastructure/components/aws/database.py
blarghmatey: Apparently it does, but that's a mistake. Removing the duplicate.

src/ol_infrastructure/lib/aws/rds_helper.py
```diff
@@ -41,4 +41,4 @@ def parameter_group_family(engine: Text, engine_version: Text) -> Text:
     :rtype: Text
     """
     engine_details = rds_client.describe_db_engine_versions(Engine=engine, EngineVersion=engine_version)
-    return engine_details['DBParameterGroupFamily']
+    return engine_details['DBEngineVersions'][0]['DBParameterGroupFamily']
```
shaidar, line 44: Why are we only interested in the first item?

src/ol_infrastructure/components/aws/database.py
blarghmatey: So, whether TLS is used looks to be a property of how the client connects and there doesn't seem to be an option to force TLS.

src/ol_infrastructure/components/aws/database.py
blarghmatey: It's not, this is just an arbitrary string but thanks for calling that out. Changing to reflect the generic use of this class.

src/ol_infrastructure/lib/aws/rds_helper.py
blarghmatey: Because the combination of engine name and version should only result in one result.

shaidar: Even when u have multiple dbs with the same engine and version?

shaidar: Nevermind

shaidar:
👍

PR #2, opened 2020-07-09
Adding module for handling standalone s3 buckets

blarghmatey:
There are some cases where we manage S3 buckets that are not associated with any application. This module is intended to handle those cases.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -89,10 +89,10 @@ description = "The AWS SDK for Python"
 name = "boto3"
 optional = false
 python-versions = "*"
-version = "1.14.9"
+version = "1.14.19"
 
 [package.dependencies]
-botocore = ">=1.17.9,<1.18.0"
+botocore = ">=1.17.19,<1.18.0"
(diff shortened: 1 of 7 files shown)


blarghmatey:
One thing I'm debating is whether it makes any sense to have a QA stack for this module.

shaidar:
Do those two file names kinda feel redundant given their place in the directory structure?
`Pulumu.infrastructure.aws.s3.Production.yaml`
`Pulumu.infrastructure.aws.s3.QA.yaml`

blarghmatey:
Those files are required. That's how the stacks are defined in Pulumi.

shaidar:
I understand they're required, but what I was referring to is that they're already in a folder called infra/aws/s3 .. so isn't kinda redundant to name them that way as well?

blarghmatey:
No, because the file name is how it defines the stack name and the stack name needs to be fully namespaced to allow for cross-"project" stack references.

shaidar:
👍

PR #3, opened 2020-07-14
Added component resource for building an S3 serverless static site

blarghmatey:
- Added component resource to build a serverless site with S3, ACM, and Cloudfront
- Updated s3 infrastructure module to build a serverless site to host the certificates

### DIFF
--- a/src/ol_infrastructure/components/aws/s3_cloudfront_site.py
+++ b/src/ol_infrastructure/components/aws/s3_cloudfront_site.py
@@ -1,0 +1,118 @@
+"""Module for creating and managing static websites hosted in S3 and delivered through Cloudfront."""
+from enum import Enum
+from typing import List, Text, Optional
+
+from pydantic import validator
+from pulumi import ComponentResource, ResourceOptions
+from pulumi_aws import acm, cloudfront, s3
+
+from ol_infrastructure.lib.ol_types import AWSBase
+
+
+class CloudfrontPriceClass(str, Enum):  # noqa: WPS600
+    """Valid price classes for CloudFront to control tradeoffs of price vs. latency for global visitors."""
+
+    # For more details on price class refer to below link and search for PriceClass
+    # https://docs.aws.amazon.com/cloudfront/latest/APIReference/API_DistributionConfig.html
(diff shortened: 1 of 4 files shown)


src/ol_infrastructure/components/aws/s3_cloudfront_site.py
```diff
@@ -0,0 +1,118 @@
...
+
+
+class S3ServerlessSiteConfig(AWSBase):
+    """Configuration object for customizing a static site hosted with S3 and Cloudfront."""
+
+    site_name: Text
+    domains: List[Text]
+    bucket_name: Text
+    site_index: Text = 'index.html'
+    cloudfront_price_class: CloudfrontPriceClass = CloudfrontPriceClass.us_eu
```
shaidar, line 30: Should that price class be hardcoded or is there a way to override it?

src/ol_infrastructure/components/aws/s3_cloudfront_site.py
```diff
@@ -0,0 +1,118 @@
...
+            validation_method='DNS',
+            subject_alternative_names=site_config.domains[1:]
+        )
+
+        s3_origin_id = f'{site_config.site_name}-s3-origin'
+
+        self.cloudfront_distribution = cloudfront.Distribution(
+            f'{site_config.site_name}-cloudfront-distribution',
+            aliases=site_config.domains,
+            comment='Cloudfront distribution for serving legacy xPRO certificates',
```
shaidar, line 80: Shouldn't that be coming from `site_config.comment` instead?

src/ol_infrastructure/components/aws/s3_cloudfront_site.py
blarghmatey: It's not hardcoded, that's just a default. This acts as a data class which gets instantiated for the specific instances.

src/ol_infrastructure/components/aws/s3_cloudfront_site.py
blarghmatey: Yes it should

shaidar:
👍

PR #5, opened 2020-07-28
Adding component for Vault database backends

blarghmatey:
----
Added a component resource to encapsulate creating a mount point, database secret backend connection, and role definitions for Postgres and MySQL/MariaDB databases.

### DIFF
--- a/src/ol_infrastructure/components/services/vault.py
+++ b/src/ol_infrastructure/components/services/vault.py
@@ -1,5 +1,81 @@
-import pulumi
-from pulumi_vault import database
+"""
+This module controls mounting and configuration of database secret backends in Vault.
+
+This includes:
+- Mount a database backend at the configured mount point
+- Set the configuration according to the requirements of the backend type
+- Control the lease TTL settings
+- Define the base set of roles according to our established best practices
+"""
+from enum import Enum
+from pulumi import ComponentResource, ResourceOptions
+from pulumi_vault import database, Mount
+from pydantic import BaseModel, SecretStr
+from typing import Text
+
+DEFAULT_PORT_POSTGRES = 5432
+DEFAULT_PORT_MYSQL = 3307
+DEFAULT_POT_MONGODB = 27017
+
+class DBEngines(str, Enum):
+    """Constraints for valid engine types that are supported by this component."""
+    postgres = 'postgresql'
+    mariadb = 'mysql'
+    mysql = 'mysql'
+    mysql_rds = 'mysql_rds'
+    mongodb = 'mongodb'
+
+class OLVaultDatabaseConfig(BaseModel):
+    db_name: Text
+    mount_point: Text
+    db_admin_username: Text
+    db_admin_password: SecretStr
+    verify_connection: bool = True
+    db_host: Text
 
 
-class VaultDatabaseRole(pulumi.ComponentResource):
+class OLVaultPostgresDatabaseConfig(OLVaultDatabaseConfig):
+    db_port: Int = DEFAULT_PORT_POSTGRES
+    db_connection: Text = 'postgresql://{{db_user}}:{{db_pass}}@{{db_host}}:{{db_port}}/{{db_name}}'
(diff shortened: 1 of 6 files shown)


shaidar:
Should we have the non-vault related changes as a separate PR?

src/ol_infrastructure/components/services/vault.py
```diff
@@ -1,5 +1,104 @@
...
+from pulumi_vault import Mount, database
+from pydantic import BaseModel, SecretStr
+
+from ol_infrastructure.lib.vault import (
+    mysql_sql_statements,
+    postgres_sql_statements
+)
+
+DEFAULT_PORT_POSTGRES = 5432
+DEFAULT_PORT_MYSQL = 3307
```
shaidar, line 23: Isn't that 3306?

src/ol_infrastructure/components/services/vault.py
```diff
@@ -1,5 +1,104 @@
...
+    db_connection: Text = 'postgresql://{{db_user}}:{{db_pass}}@{{db_host}}:{{db_port}}/{{db_name}}'
+    db_type: Text = DBEngines.postgres.value
+    role_statements: Dict[Text, Text] = postgresql_sql_statements
+
+
+class OLVaultMysqlDatabaseConfig(OLVaultDatabaseConfig):
+    db_port: Int = DEFAULT_PORT_MYSQL
+    db_connection: Text = '{{db_user}}:{{db_pass}}@tcp({{db_host}}:{{db_port}})/'
+    db_type: Text = DBEngines.mysql_rds.value
+    role_statements: Dict[Text, Text] = postgresql_sql_statements
```
shaidar, line 57: Shouldn't this be `mysql_sql_statements`?

src/ol_infrastructure/components/services/vault.py
```diff
@@ -1,5 +1,104 @@
...
+
+
+class OLVaultDatabaseBackend(ComponentResource):
+
+    def __init__(self, db_config: OLVaultDatabaseConfig, opts: ResourceOptions = None):
+        super().__init__('ol:services:VaultDatabaseBackend', db_config.db_name, None, opts)
```
shaidar, line 63: Not sure I follow where this is coming from

src/ol_infrastructure/components/services/vault.py
```diff
@@ -1,5 +1,104 @@
...
+
+        resource_opts = ResourceOptions(parent=self).merge(opts)
+
+        self.db_mount = Mount(
+            f'{db_config.db_name}-mount-point',
```
shaidar, line 68: Looks like in the class definition the [name](https://github.com/mitodl/ol-infrastructure/blob/main/src/ol_infrastructure/components/aws/database.py#L57) is optional. Do we want to change that in order to avoid having a db_mount that doesn't have a db_name?

src/ol_infrastructure/components/services/vault.py
blarghmatey: That's part of the pattern for a component resource. It requires a call to super https://www.pulumi.com/docs/intro/concepts/programming-model/#authoring-a-new-component

src/ol_infrastructure/components/services/vault.py
blarghmatey: This is a different db_config object. The `OLVaultDatabaseConfig` has a `db_name` attribute that's required. For the RDS component, the db_name is for the schema to be added after creation of the instance.

shaidar: So we're setting up a separate vault mountpoint for each db as opposed to how we currently have some of the mount points defined where we use the same mountpoint but use different paths to split envs. Why did we choose to go that route and do we foresee any issues with this approach?

blarghmatey: We've always had different mount points for the DBs. The generic secret mounts are where we have different paths per env.

shaidar:
👍

PR #7, opened 2020-08-04
Adding Dockerfile definition for vdist build container

blarghmatey:
Using the ol-infrastructure repository as the default home for dockerfiles to be used to build utility containers. I'm still not totally convinced that this is the right solution, but wanted to get this somewhere other than my local disk.

### DIFF
--- a/dockerfiles/vdist_debian
+++ b/dockerfiles/vdist_debian
@@ -1,0 +1,16 @@
+#-*- mode: dockerfile -*-
+FROM debian:buster
+LABEL maintainer="MIT Open Learning (odl-devops@mit.edu)"
+LABEL description="Image to be used by vdist to package python applications into debian deb packages."
+# Abort on error.
+RUN set -e
+# Sometimes deb-src repositories comes disabled by default, we're activating it
+# to use them for apt-get build-dep.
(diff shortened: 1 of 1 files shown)


dockerfiles/vdist_debian
```diff
@@ -0,0 +1,15 @@
+FROM debian:stretch
```
shaidar, line 1: Should we go with `buster` instead?

dockerfiles/vdist_debian
```diff
@@ -0,0 +1,16 @@
+#-*- mode: dockerfile -*-
+FROM debian:buster
+LABEL maintainer="MIT Open Learning (odl-devops@mit.edu)"
+LABEL description="Image to be used by vdist to package python applications into debian deb packages."
+# Abort on error.
+RUN set -e
+# Sometimes deb-src repositories comes disabled by default, we're activating it
+# to use them for apt-get build-dep.
+RUN printf "deb-src http://deb.debian.org/debian stretch main \ndeb-src http://deb.debian.org/debian stretch-updates main\ndeb-src http://security.debian.org stretch/updates main\n" >> /etc/apt/sources.list
```
shaidar, line 9: Should `stretch` be changed to buster?

shaidar:
👍

PR #8, opened 2020-08-12
Adding VPC peering between operations and data networks

blarghmatey:
- Added conditional importing of network resources defined in OLVPC component
- Added import of operations VPC
- Added peering between data and operations VPCs with associated routing rules

### DIFF
--- a/src/ol_infrastructure/lib/aws/ec2_helper.py
+++ b/src/ol_infrastructure/lib/aws/ec2_helper.py
@@ -1,3 +1,5 @@
+"""Helper functions for working with EC2 resources."""
+
 from functools import lru_cache
 from ipaddress import IPv4Network
 from types import FunctionType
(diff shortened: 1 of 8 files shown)


src/ol_infrastructure/lib/aws/ec2_helper.py
```diff
@@ -18,7 +21,7 @@ def aws_regions() -> List[Text]:
 
 
 @lru_cache
-def availability_zones(region: Text) -> List[Text]:
+def availability_zones(region: Text) -> Tuple[Text]:
```
shaidar, line 24: Shouldn't `rtype` in doc string change as well?

shaidar:
@blarghmatey Is the `pre-commit-config` intentional?

blarghmatey:
Yes. I want it to be in the repo, but I can back it out and commit it directly to main if you prefer.

shaidar:
👍

PR #9, opened 2020-08-18
Adding initial implementation of dynamic resource provider for Salt

blarghmatey:
In order to connect salt minions with the salt master we need to pre-create the minion keys. This resource provider uses the Salt API to generate a minion keypair and accept it on the master, so that we can pass the key to the user-data input of an EC2 instance at creation time.


src/ol_infrastructure/providers/salt/minion.py
```diff
@@ -0,0 +1,104 @@
...
+from pulumi import Input, Output, ResourceOptions
+from pulumi.dynamic import CreateResult, ReadResult, Resource, ResourceProvider
+from pydantic import BaseModel, SecretStr
+
+
+class OLSaltStackInputs(BaseModel):
+    minion_id: Input[Text]
+    salt_api_url: Input[Text]
+    salt_user: Input[Text]
+    salt_password: Input[SecretStr]
```
shaidar, line 19: How are we going to be managing/rotating those creds?

blarghmatey: That will be handled using the Pulumi stack configs, so it's out of band of Pulumi itself. I actually needed to check with you on the current status of the Salt API on QA and production.

src/ol_infrastructure/providers/salt/minion.py
```diff
@@ -0,0 +1,104 @@
...
+    salt_auth_method: Input[Text] = 'pam'
+
+
+class _OLSaltStackProviderInputs(BaseModel):
```
shaidar, line 23: Does this need to be its own class or can those attributes be part of the `OLSaltStackProvider` one?

blarghmatey: It needs to be its own class, because these are the 'unwrapped' values, where the other class will be receiving the data as a Pulumi `Input` type. https://www.pulumi.com/docs/intro/concepts/programming-model/#resource-provider-interface

shaidar:
👍

PR #12, opened 2020-09-10
Adding the remaining VPCs to the network project

blarghmatey:
Added the definitions for the remaining VPCs to import and peer them


src/ol_infrastructure/infrastructure/aws/network/__main__.py
```diff
@@ -34,9 +42,25 @@ def vpc_exports(vpc: OLVPC) -> Dict:
 stack = get_stack()
 stack_name = stack.split('.')[-1]
 env_suffix = stack_name.lower()
+
+apps_config = Config('apps_vpc')
```
shaidar, line 46: nit-pick: Should be stick to either calling it `applications` or `apps` and sticking to that or is there something I'm missing?

shaidar:
👍

src/ol_infrastructure/infrastructure/aws/network/__main__.py
blarghmatey: No, you're not missing anything, just me being a lazy typist. The one thing to note is that the `apps` references are local to the Pulumi code, and the `applications` references are what get reflected/deployed on AWS

PR #13, opened 2020-09-14
Added handling of cross-VPC Consul security groups

blarghmatey:
Added handling of discovering VPC peers to allow for specifying the network segments that the security group should allow.


shaidar:
Might be useful to add to the README what `environment:services` will be used for

shaidar:
👍

PR #14, opened 2020-09-17
Adding more precise filtering for VPC and subnet imports

blarghmatey:
There are some VPCs with overlapping CIDR ranges, so more precise filtering is needed to ensure the proper import. Also removed references to the OCW VPC due to the fact that it is a legacy environment and doesn't match our requirements for CIDR ranges.


shaidar:
👍

PR #15, opened 2020-09-29
Added KMS config to create CMKs keys

shaidar:
This creates symmetric keys that are auto-rotated on a yearly basis to be used for encrypting EBS volumes on EC2 instances. In addition to having `root` as one of the principals to ensure that accidental deletion of an account doesn't lead to complete loss of access to KMS, I statically added users as a group can't be added which would've been preferable.


src/ol_infrastructure/infrastructure/aws/kms/__main__.py
```diff
@@ -0,0 +1,84 @@
...
+                'kms:Decrypt',
+                'kms:ReEncrypt*',
+                'kms:GenerateDataKey*',
+                'kms:CreateGrant',
+                'kms:DescribeKey'
+            ],
+            'Resource': '*',
+            'Condition': {
+                'StringEquals': {
+                    'kms:ViaService': 'ec2.us-east-1.amazonaws.com',
```
blarghmatey, line 36: Not currently an issue, as we only use us-east-1 currently, but should we make this applicable to any ec2 region?

src/ol_infrastructure/infrastructure/aws/kms/__main__.py
```diff
@@ -0,0 +1,84 @@
...
+                'AWS': [
+                  f'arn:aws:iam::{owner}:root',
+                  f'arn:aws:iam::{owner}:user/mbreedlove',
+                  f'arn:aws:iam::{owner}:user/shaidar',
+                  f'arn:aws:iam::{owner}:user/tmacey',
+                ]
+            },
+            'Action': [
+                 'kms:Create*',
+	             'kms:Describe*',
```
blarghmatey, line 54: indentation

src/ol_infrastructure/infrastructure/aws/kms/__main__.py
```diff
@@ -0,0 +1,84 @@
...
+	description=f"Key to encrypt/decrypt {env_suffix} EBS volumes",
+	enable_key_rotation=True,
+	is_enabled=True,
+	key_usage='ENCRYPT_DECRYPT',
+	policy=json.dumps(kms_ec2_ebs_encryption_policy),
+	tags={
+	    'OU': 'operations',
+	    'Environment': f'operations-{env_suffix}'
+	},
+)
```
blarghmatey, line 84: Don't forget about exporting the key ID, and whatever other relevant informaiton so that we can reference it in other modules.

blarghmatey:
Overall looks pretty good. Don't forget to also create a stack definition for production (and maybe Dev?)

blarghmatey:
:+1:

PR #16, opened 2020-10-05
Add OCW Build server

markbreedlove:


src/ol_infrastructure/applications/ocw_build/__main__.py
```diff
@@ -0,0 +1,220 @@
...
+        salt_user=get_config('saltstack:api_user'),
+        salt_password=get_config('saltstack:api_password')
+    )
+)
+
+cloud_init_userdata = build_userdata(
+    instance_name=ocw_build_minion_id,
+    minion_keys=salt_minion,
+    minion_roles=['ocw-build'],
+    minion_environment=ocw_next_build_environment,
```
blarghmatey, line 153: This will need to match the `environment` grain that we are using in the salt code. For the apps VPCs we'll likely want to do some migration to align naming between projects. For this particular server, assuming that there are no pillar or state values that rely on the environment grain then we should probably keep this value as-is.

src/ol_infrastructure/applications/ocw_build/__main__.py
```diff
@@ -0,0 +1,220 @@
...
+    salt_host=f'salt-{env_suffix}.private.odl.mit.edu'
+)
+
+iam_image = get_ami(
```
blarghmatey, line 157: This invocation of get_ami is trying to retrieve an image named `ocw-build` which we aren't currently building. We _can_ do that but would need a Packer configuration in the salt-ops repository. For now, this should just import the `debian_10_ami` object from the `ol_infrastructure.lib.aws.ec2_helper` module and pass that to the `ec2.Instance` call.

src/ol_infrastructure/applications/ocw_build/__main__.py
```diff
@@ -0,0 +1,220 @@
...
+    f'ocw-build-{env_suffix}-service-domain-v6',
+    name=get_config('ocw_build:domain'),
+    type='AAAA',
+    ttl=fifteen_minutes,
+    records=[ec2_instance.ipv6_addresses],
+    zone_id=mitodl_zone_id,
+    opts=ResourceOptions(depends_on=[ec2_instance])
+)
+
+export('dagster_app', {
```
blarghmatey, line 216: Copy pasta

src/ol_infrastructure/applications/ocw_build/__main__.py
```diff
@@ -0,0 +1,220 @@
...
+    records=[ec2_instance.public_ip],
+    zone_id=mitodl_zone_id,
+    opts=ResourceOptions(depends_on=[ec2_instance])
+)
+route53_domain_v6 = route53.Record(
+    f'ocw-build-{env_suffix}-service-domain-v6',
+    name=get_config('ocw_build:domain'),
+    type='AAAA',
+    ttl=fifteen_minutes,
+    records=[ec2_instance.ipv6_addresses],
```
blarghmatey, line 211: The `ipv6_addresses` attribute is already a list

src/ol_infrastructure/applications/ocw_build/__main__.py
```diff
@@ -0,0 +1,220 @@
...
+    subnet_id=apps_vpc['subnet_ids'][0],
+    key_name=get_config('saltstack:key_name'),
+    root_block_device=ec2.InstanceRootBlockDeviceArgs(
+        volume_type='gp2',
+        volume_size=100
+    ),
+    vpc_security_group_ids=[
+        apps_vpc['security_groups']['default'],
+        apps_vpc['security_groups']['web'],
+        apps_vpc['security_groups']['salt_minion'],
```
blarghmatey, line 189: These security groups aren't currently being exported in the Network project, so that file will need to be modified to include code as shown [here](https://github.com/mitodl/ol-infrastructure/blob/main/src/ol_infrastructure/infrastructure/aws/network/__main__.py#L149-L171)

blarghmatey (changes requested):

src/ol_infrastructure/applications/ocw_build/__main__.py
markbreedlove: Fixed in f37603e

src/ol_infrastructure/applications/ocw_build/__main__.py
markbreedlove: Fixed in 3c5bbc2

src/ol_infrastructure/applications/ocw_build/__main__.py
markbreedlove: Fixed in b259de5

src/ol_infrastructure/applications/ocw_build/__main__.py
markbreedlove: Is the new name property OK?

src/ol_infrastructure/applications/ocw_build/__main__.py
blarghmatey: This instance is getting deployed in the apps VPCs, so the name used in Salt currently is `rc-apps` and `production-apps`, vs `applications-qa` and `applications-production` in Pulumi. As long as the salt code for the OCW build machine isn't relying on the environment grain anywhere then I think we should leave it as is and we should plan to migrate the naming in the Salt code to use the format that we defined here.

src/ol_infrastructure/applications/ocw_build/__main__.py
markbreedlove: I've added something in 1495063. Does it look right?

blarghmatey:
This all looks good. There's some order of operations things that we need to handle before we add automated Consul support, so we should be able to merge this as is.

:+1:

blarghmatey (approved):

PR #18, opened 2020-10-08
Adding fix for conditional resource imports to avoid loops

blarghmatey:
In the current implementation the imported resource doesn't get the `pulumi_managed` tag applied so that subsequent runs will try to import the same resource again. This adds a step to use boto3 to apply that tag after generating the ResourceOptions that specify the import but only when the resources are actually being manipulated.


markbreedlove:
Is this intended to fix the error I was getting with `pulumi up` on `infrastructure.aws.network.Production`?

blarghmatey:
This will prevent that situation from occurring in the future.

markbreedlove (approved):

PR #20, opened 2020-10-20
Deploy Redash application

blarghmatey:
Deploy the full stack of Redash services using Pulumi and launching the Redash AMI

### DIFF
--- a/src/ol_infrastructure/applications/redash/__main__.py
+++ b/src/ol_infrastructure/applications/redash/__main__.py
@@ -3,15 +3,30 @@
 - Create an RDS PostgreSQL instance for storing Redash's configuration data and intermediate query results
 - Mount a Vault database backend and provision role definitions for the Redash RDS database
 - Create an IAM role for Redash instances to allow access to S3 and other AWS resources
+- Create a Redis cluster in Elasticache
 - Register a minion ID and key pair with the appropriate SaltStack master instance
-- Provision an EC2 instance from a pre-built AMI with the configuration and code for Redash
+- Provision a set of EC2 instances from a pre-built AMI with the configuration and code for Redash
+- Provision an AWS load balancer and connect the deployed EC2 instances
+- Create a DNS record for the deployed load balancer
 """
+import json
+
 from itertools import chain
 
 from pulumi import Config, ResourceOptions, StackReference, export, get_stack
 from pulumi.config import get_config
-from pulumi_aws import ec2, get_ami, get_caller_identity, iam, route53
+from pulumi_aws import (
+    ec2,
+    get_ami,
+    get_caller_identity,
+    iam,
+    route53
+)
 
+from ol_infrastructure.components.aws.cache import (
+    OLAmazonCache,
+    OLAmazonRedisConfig
+)
 from ol_infrastructure.components.aws.database import (
     OLAmazonDB,
     OLPostgresDBConfig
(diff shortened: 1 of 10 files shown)


src/ol_infrastructure/components/aws/cache.py
```diff
@@ -0,0 +1,211 @@
...
+        engines_map = cache_engines()
+        if engine_version not in engines_map.get(engine, []):
+            raise ValueError(f'The specified version of the {engine} engine is not supported in AWS.')
+        return engine_version
+
+
+class OLAmazonRedisConfig(OLAmazonCacheConfig):
+    cluster_mode_enabled: bool = True
+    encrypt_transit: bool = True
+    auth_token: Optional[Text] = None
```
shaidar, line 69: Wouldn't this throw a ValueError since its value is `None` and `encrypt_transit` is `True`?

blarghmatey: If you use the defaults, then yes. You can solve that situation by either setting `encrypt_transit` to False or by setting the `auth_token` value.

src/ol_infrastructure/components/aws/cache.py
```diff
@@ -0,0 +1,211 @@
...
+        if clustered_redis:
+            cache_parameters.append(
+                elasticache.ParameterGroupParameterArgs(
+                    name='cluster-enabled', value='yes'
+                )
+            )
+            if cache_config.engine_version.startswith('6'):
+                cache_options = resource_options.merge(
+                    pulumi.ResourceOptions(ignore_changes=['engine_version'])
+                )  # type: ignore
```
shaidar, line 149: Not sure I follow what this is supposed to do and also am hesitant about having the Redis version hardcoded here.

blarghmatey: It's not hardcoded, it's just set to a sane default. As to what's happening, the issue is that with releases of Redis >= 6 AWS refers to that version as `6.x`, but when it gets deployed that resolves to a concrete version (e.g. 6.0.5), so that when you attempt to re-run the `up` it shows an attribute mismatch, so this tells Pulumi to ignore that difference.

src/ol_infrastructure/applications/redash/__main__.py
```diff
@@ -0,0 +1,281 @@
...
+redash_db_vault_backend = OLVaultDatabaseBackend(redash_db_vault_backend_config)
+
+# Set up Redis instance for Redash in Elasticache
+redis_config = Config('redis')
+redash_redis_config = OLAmazonRedisConfig(
+    encrypt_transit=True,
+    auth_token=redis_config.require('auth_token'),
+    engine_version='6.x',
+    node_group_count=3,
+    replica_count=3,
```
shaidar, line 157: I don't see either `node_group_count` or `replica_count` defined in components.aws.cache

src/ol_infrastructure/applications/redash/__main__.py
```diff
@@ -0,0 +1,281 @@
...
+    auto_upgrade=True,
+    cluster_description='Redis cluster for Redash tasks and caching',
+    cluster_name=f'redash-redis-{redash_environment}',
+    num_instances=3,
+    security_groups=['dummy-group'],
```
shaidar, line 162: ?

blarghmatey: I put that there because I needed to be able to instantiate the object for referencing it in the next block where I create the security group and then override it in the config object before passing it through to the component for creation.

src/ol_infrastructure/applications/redash/__main__.py
blarghmatey: That's because I changed the names of those attributes after creating them. Fixing now...

shaidar:
Don't see any obvious issues so 👍

PR #21, opened 2020-10-29
Adding support for linting IAM policies

blarghmatey:
In order to avoid issues with poorly scoped access or bugs in policies I added integration with the Parliament library for validating IAM policy documents at runtime before resources are deployed.

### DIFF
--- a/src/ol_infrastructure/applications/ocw_build/__main__.py
+++ b/src/ol_infrastructure/applications/ocw_build/__main__.py
@@ -13,12 +13,12 @@ push web content to the OCW site's S3 buckets.
 
 TODO: consul cloud autojoin functionality
 '''
-
 from pulumi import ResourceOptions, StackReference, export, get_stack
 from pulumi.config import get_config
 from pulumi_aws import ec2, iam, route53, s3
 
 from ol_infrastructure.lib.aws.ec2_helper import build_userdata, debian_10_ami
+from ol_infrastructure.lib.aws.iam_helper import lint_iam_policy
 from ol_infrastructure.lib.ol_types import AWSBase
 from ol_infrastructure.providers.salt.minion import (
     OLSaltStackMinion,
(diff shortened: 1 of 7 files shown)


src/ol_infrastructure/infrastructure/aws/policies/__main__.py
```diff
@@ -13,7 +14,9 @@
         }
     ]
 }
-
+describe_instance_policy_lint = analyze_policy_string(json.dumps(describe_instance_policy_document))
```
shaidar, line 17: Can't we use the helper function in this file?

src/ol_infrastructure/applications/ocw_build/__main__.py
```diff
@@ -95,6 +95,7 @@
         }
     ]
 }
+lint_iam_policy(ocw_next_instance_policy_data)
```
shaidar, line 98: Not related specifically to this PR, but do we want to standardize on how we refer to app policies? Is it `<app_name>_policy` or `<app_name>_policy_data` or something else ...

src/ol_infrastructure/infrastructure/aws/policies/__main__.py
```diff
@@ -13,38 +13,39 @@
         }
     ]
 }
-
 describe_instance_policy = iam.Policy(
     'describe-ec2-instances-policy',
     name='describe-ec2-instances-policy',
     path='/ol-operations/describe-ec2-instances-policy/',
-    policy=json.dumps(describe_instance_policy_document),
+    policy=lint_iam_policy(describe_instance_policy_document, stringify=True),
```
shaidar, line 20: Why are we running the linter here and not when the policy is originally defined similar to what you've done below with the cloudwatch policy?

shaidar:
👍

PR #22, opened 2020-11-01
Added Vault backends and roles

shaidar:
Added ability to create Vault AWS secrets engine and approle along with the ability to pass policies and creates those vault policies as part of enabling the vault backends.

### DIFF
--- a/src/ol_infrastructure/components/services/vault.py
+++ b/src/ol_infrastructure/components/services/vault.py
@@ -74,7 +74,7 @@ class OLVaultDatabaseBackend(ComponentResource):
             db_config: Union[OLVaultMysqlDatabaseConfig, OLVaultPostgresDatabaseConfig],
             opts: ResourceOptions = None
     ):
-        super().__init__('ol:services:VaultDatabaseBackend', db_config.db_name, None, opts)
+        super().__init__('ol:services:Vault:DatabaseBackend', db_config.db_name, None, opts)
 
         resource_opts = ResourceOptions.merge(ResourceOptions(parent=self), opts)  # type: ignore
 
@@ -146,15 +146,15 @@ class OLVaultAWSSecretsEngineConfig(BaseModel):
     max_lease_ttl_seconds: int = SIX_MONTHS
     description: Text
     aws_secret_key: Text
-    path: Text
+    vault_backend_path: Text
     policy_documents: Dict[Text, Text]
     credential_type: Text = 'iam_user'
 
-    @validator('path')
-    def is_valid_path(cls: 'OLVaultAWSSecretsEngineConfig', path: Text) -> Text:
-        if path.startswith('/') or path.endswith('/'):
-            raise ValueError(f'The specified path value {path} can not start or end with a slash')
-        return path
+    @validator('vault_backend_path')
+    def is_valid_path(cls: 'OLVaultAWSSecretsEngineConfig', vault_backend_path: Text) -> Text:
+        if vault_backend_path.startswith('/') or vault_backend_path.endswith('/'):
+            raise ValueError(f'The specified path value {vault_backend_path} can not start or end with a slash')
+        return vault_backend_path
 
 
 class OLVaultAWSSecretsEngine(ComponentResource):
@@ -164,7 +164,7 @@ class OLVaultAWSSecretsEngine(ComponentResource):
             engine_config: OLVaultAWSSecretsEngineConfig,
             opts: ResourceOptions = None
     ):
-        super().__init__('ol:services:VaultAWSSecretsEngine', engine_config.app_name, None, opts)
+        super().__init__('ol:services:Vault:AWSSecretsEngine', engine_config.app_name, None, opts)
 
         resource_options = ResourceOptions(parent=self).merge(opts)  # type: ignore
 
@@ -173,7 +173,7 @@ class OLVaultAWSSecretsEngine(ComponentResource):
             f'aws-{engine_config.app_name}',
             access_key=engine_config.aws_access_key,
             secret_key=engine_config.aws_secret_key,
-            path=f'aws-{engine_config.app_name}',
+            path=f'aws-{engine_config.vault_backend_path}',
             opts=resource_options,
         )
 
@@ -210,7 +210,7 @@ class OLVaultAppRoleAuthBackend(ComponentResource):
             approle_config: OLVaultAppRoleAuthBackendConfig,
             opts: ResourceOptions = None
     ):
-        super().__init__('ol:services:VaultAppRoleAuthBackend', approle_config.authbackend_name, None, opts)
+        super().__init__('ol:services:Vault:AppRoleAuthBackend', approle_config.authbackend_name, None, opts)
 
         resource_options = ResourceOptions(parent=self).merge(opts)  # type: ignore
 
--- a/src/ol_infrastructure/components/services/vault.py
+++ b/src/ol_infrastructure/components/services/vault.py
@@ -7,13 +7,16 @@ This includes:
 - Control the lease TTL settings
 - Define the base set of roles according to our established best practices
 """
+import json
+
 from enum import Enum
-from typing import Dict, Text, Union
+from typing import Dict, List, Text, Union
 
 from pulumi import ComponentResource, Output, ResourceOptions
-from pulumi_vault import Mount, database
+from pulumi_vault import AuthBackend, Mount, Policy, approle, aws, database
 from pydantic import BaseModel
 
+from ol_infrastructure.lib.ol_types import Apps
 from ol_infrastructure.lib.vault import (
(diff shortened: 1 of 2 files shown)


src/ol_infrastructure/components/services/vault.py
```diff
@@ -134,3 +137,81 @@ def __init__(
...
+
+class OLVaultAWSSecretsEngineConfig(BaseModel):
+    app_name: Text
+    access_key: Text
+    default_lease_ttl_seconds: int = SIX_MONTHS
+    max_lease_ttl_seconds: int = SIX_MONTHS
+    description: Text
+    secret_key: Text
+    path: Text
+    policy_documents: Dict
```
blarghmatey, line 149: What kind of `Dict`? Would be good to specify the valid subtypes for the dictionary. e.g. `Dict[Text, Any]`, `Dict[Text, Text]`, etc. Would also be good to add a comment or validation function to specify what the constraints are.

src/ol_infrastructure/components/services/vault.py
```diff
@@ -134,3 +137,81 @@ def __init__(
...
+    credential_type: str = 'iam_user'
```
blarghmatey, line 150: Should be `Text` to be consistent with other type hints.

src/ol_infrastructure/components/services/vault.py
```diff
@@ -134,3 +137,81 @@ def __init__(
...
+            f'aws-{engine_config.app_name}',
+            access_key=engine_config.access_key,
+            secret_key=engine_config.secret_key,
+            path=f'aws-{engine_config.app_name}'
+        )
+
+        for role_name, policy in engine_config.policy_documents.items():
+            aws.SecretBackendRole(
+                role_name,
+                backend=f'aws-{engine_config.app_name}',
```
blarghmatey, line 173: Can we use the name or path attribute of the secret backend created on line 162? That creates an explicit dependency chain between the resources.

src/ol_infrastructure/components/services/vault.py
```diff
@@ -134,3 +137,81 @@ def __init__(
...
+                credential_type='iam_user',
```
blarghmatey, line 174: Should this be coming from the engine_config object?

src/ol_infrastructure/components/services/vault.py
```diff
@@ -134,3 +137,81 @@ def __init__(
...
+                name=role_name,
+                policy_document=json.dumps(policy),
```
blarghmatey, line 176: Are these going to be IAM policies? If so, should we use the iam linting that we just merged?

src/ol_infrastructure/components/services/vault.py
```diff
@@ -134,3 +137,81 @@ def __init__(
...
+                opts=ResourceOptions(depends_on=[self.aws_secrets_engine])
```
blarghmatey, line 177: This class should have a baseline `opts` object that sets `parent=self` and gets passed to all resources created within the scope of the component resource.

src/ol_infrastructure/components/services/vault.py
```diff
@@ -134,3 +137,81 @@ def __init__(
...
+            )
+
+        self.register_outputs({})
+
+
+class OLVaultAppRoleAuthBackendConfig(BaseModel):
+    name: Text
```
blarghmatey, line 184: What is this the name of? It's better to have a more descriptive attribute name.

src/ol_infrastructure/components/services/vault.py
```diff
@@ -134,3 +137,81 @@ def __init__(
...
+    type: str = 'approle'
+    description: Text
+    backend: Text
```
blarghmatey, line 187: Maybe a more descriptive attribute name, or a comment as to where to determine what the value of `backend` should be?

src/ol_infrastructure/components/services/vault.py
```diff
@@ -134,3 +137,81 @@ def __init__(
...
+        super().__init__('ol:services:VaultAppRoleAuthBackend', approle_config.name, None, opts)
+
+        self.approle_backend = AuthBackend(
+            approle_config.name,
+            description=approle_config.description,
+            path=approle_config,
+            type=approle_config.type)
+
+        for policy in approle_config.token_policies:
+            self.approle_policy = Policy(f'{approle_config.name}', policy=policy)
```
blarghmatey, line 208: This gets overwritten on each iteration through the loop. If you want to preserve it then you should assign a list container to `self` along the lines of `self.approle_policies` or `self.role_policies`

src/ol_infrastructure/components/services/vault.py
```diff
@@ -134,3 +137,81 @@ def __init__(
...
+
+        self.approle_backend_role = approle.AuthBackendRole(
+            approle_config.name,
+            backend=self.approle_backend.path,
+            role_name=approle_config.role_name,
+            token_policies=approle_config.token_policies,
```
blarghmatey, line 214: Might be good to populate a list of policies in the loop above and then pass that list here so that there is a direct dependency of the order of resource creation.

src/ol_infrastructure/lib/ol_types.py
```diff
@@ -25,6 +25,21 @@ class BusinessUnit(str, Enum):  # noqa: WPS600
...
+    """Canonical source of truth for defining apps."""
+
+    bootcamps = 'bootcamps'
+    micromasters = 'micromasters'
+    ocw_build = 'ocw-build'
+    redash = 'redash'
+    dagster = 'dagster'
+    mitxpro_edx = 'xpro-edx'
+    mitx_edx = 'mitx-edx'
+    odl_video_service = 'ovs'
```
blarghmatey, line 39: Should we have the value of this be `odl-video`?

shaidar: It can be, but seems to be internally identified now as `ovs`

src/ol_infrastructure/components/services/vault.py
shaidar: Don't think I fully understand your argument

src/ol_infrastructure/components/services/vault.py
```diff
@@ -134,3 +137,81 @@ def __init__(
...
         self.register_outputs({})
+
+class OLVaultAWSSecretsEngineConfig(BaseModel):
+    app_name: Text
+    access_key: Text
+    default_lease_ttl_seconds: int = SIX_MONTHS
+    max_lease_ttl_seconds: int = SIX_MONTHS
+    description: Text
+    secret_key: Text
+    path: Text
```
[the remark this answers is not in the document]
blarghmatey: I think it would be good to make the `path` attribute more descriptive to indicate what type of path it is supposed to be. e.g. `mount_path` or `vault_backend_path`, etc.

src/ol_infrastructure/components/services/vault.py
blarghmatey: So, the pattern I have been using is to create a top level `opts` object at the beginning of the `__init__` method (e.g. [here](https://github.com/mitodl/ol-infrastructure/blob/sar%2Fvault_backend/src/ol_infrastructure/components/aws/database.py#L143))

And then further down for each child resource pass that top level opts object like [here](https://github.com/mitodl/ol-infrastructure/blob/sar%2Fvault_backend/src/ol_infrastructure/components/aws/database.py#L150)

This creates a dependency structure between the component object and the child resources to ensure that everything gets created and cleaned up as a unit.

src/ol_infrastructure/components/services/vault.py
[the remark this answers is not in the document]
shaidar: I'm partially of the mindset that if you're creating a secrets engine mount and you're not aware of what `path` is then you probably should spend some time reading the vault docs prior to using this. Additionally, feel like it's easier to debug when the values we're using map to the same ones in pulumi and hopefully in vault.

src/ol_infrastructure/components/services/vault.py
```diff
@@ -199,24 +210,29 @@ def __init__(
             approle_config: OLVaultAppRoleAuthBackendConfig,
             opts: ResourceOptions = None
     ):
-        super().__init__('ol:services:VaultAppRoleAuthBackend', approle_config.name, None, opts)
+        super().__init__('ol:services:VaultAppRoleAuthBackend', approle_config.authbackend_name, None, opts)
```
blarghmatey, line 213: Just a nitpick, but since we have a number of components for Vault we should probably change the namespacing to e.g. `ol:services:Vault:AppRoleAuthBackend`

blarghmatey:
Aside from the merge conflict, I think the only remaining issue is to set the `opts` object at the top level of the class to set all of the created resources as children of the component.

blarghmatey:
:+1:

PR #24, opened 2020-11-12
Adding Black as precommit hook and applying it to existing code

blarghmatey:
- Adding Black formatter
- Updating Flake8 config to match with Black
- Fixing Flake8 pre-commit hook to use wemake-python-styleguide plugin

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,6 +1,7 @@
 [flake8]
-max_line_length = 120
-ignore = C812,WPS319,WPS305,WPS318,DAR202,WPS326,DAR301,WPS306,WPS226,WPS432
+max_line_length = 88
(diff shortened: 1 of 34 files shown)


setup.cfg
```diff
@@ -1,6 +1,7 @@
 [flake8]
-max_line_length = 120
-ignore = C812,WPS319,WPS305,WPS318,DAR202,WPS326,DAR301,WPS306,WPS226,WPS432
+max_line_length = 88
+ignore = WPS305,ANN101,C812,I001,I002,I003,I004,E203,E501,W503 #,WPS319,WPS318,WPS326,WPS306,WPS226,WPS432,W503
```
shaidar, line 3: Is `#` accidental?

shaidar:
Apart from comment above, I skimmed through the changes and looks good.

setup.cfg
blarghmatey: No, that's on purpose. Just trying to make sure I'm not ignoring things that can be easily addressed through auto-formatting.

PR #27, opened 2020-12-14
Vault PKI & Roles

shaidar:
Create the vault intermediate CA along with PKI intermediate backends in specified environments. Additionally, config to create roles for said intermediate PKI's


src/ol_infrastructure/components/services/vault.py
```diff
@@ -250,3 +267,232 @@ def __init__(
...
+                    f"{VAULT_API_URL}/backend_confg.environment_name/ca"
+                ],
+            )
+        )
+
+        self.register_outputs({})
+
+
+class OLVaultPKIIntermediateRoleConfig(BaseModel):
+    pki_intermediate_backend: Text
```
blarghmatey, line 434: Should we set this to be an instance of the actual backend object to create a dependency chain from the role to the PKI mount? That will allow us to ensure that we are pointing at a valid backend rather than relying on a stringly typed field.

src/ol_infrastructure/components/services/vault.py
```diff
@@ -250,3 +267,232 @@ def __init__(
...
+    role_name: Text
+    key_type: Text = "rsa"
```
blarghmatey, line 436: Might be good to create a helper object to constrain the options and the corresponding key bits. e.g. rsa/4096 bits, ecdh/256 bits, etc.

src/ol_infrastructure/components/services/vault.py
```diff
@@ -250,3 +267,232 @@ def __init__(
...
+
+        self.register_outputs(
+            {"pki_intermediate_ca": self.pki_intermediate_ca_backend.id}
+        )
+
+
+class OLVaultPKIIntermediateEnvBackendConfig(BaseModel):
+    environment_name: Text  # e.g. mitx-qa
+    max_ttl: int = TWELVE_MONTHS
+    default_ttl: int = TWELVE_MONTHS
```
blarghmatey, line 351: Same as above, should we set the default_ttl lower than the max TTL?

src/ol_infrastructure/components/services/vault.py
```diff
@@ -250,3 +267,232 @@ def __init__(
         )
 
         self.register_outputs({})
+
+
+class OLVaultPKIIntermediateCABackendConfig(BaseModel):
+    max_ttl: int = TWELVE_MONTHS
+    default_ttl: int = TWELVE_MONTHS
```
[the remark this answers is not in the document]
shaidar: Ideally yes, but I was somewhat concerned that we currently don't have a reliable way to renew this stuff, but also think the more pertinent one is the ttl set for roles and that's six months. We can always tune the mount whenever we want

src/ol_infrastructure/components/services/vault.py
```diff
@@ -250,3 +261,232 @@ def __init__(
...
+        )
+
+        self.register_outputs({})
+
+
+class OLVaultPKIIntermediateRoleConfig(BaseModel):
+    pki_intermediate_backend: Text
+    role_name: Text
+    key_type: Text = VaultPKIKeyTypeBits.rsa.name
+    key_bits: int = VaultPKIKeyTypeBits.rsa
```
blarghmatey, line 431: This should just be condensed into a single line like `key_config: VaultPKIKeyTypeBits` and then in the body of the logic just use `key_type = key_config.name, key_bits = key_config.value` or whatever the proper enum syntax is.

blarghmatey: I might also generalize it to just by `crypto_key_defaults` or something like that.

src/ol_infrastructure/components/services/vault.py
```diff
@@ -473,8 +472,8 @@ def __init__(
             backend=role_config.pki_intermediate_backend,
             allowed_domains=role_config.allowed_domains,
             allow_glob_domains=True,
-            key_type=role_config.key_type,
-            key_bits=role_config.key_bits,
+            key_type=VaultPKIKeyTypeBits.rsa.name,
+            key_bits=VaultPKIKeyTypeBits.rsa,
```
blarghmatey, line 476: This is hardcoding it to use RSA. You shoudl be able to just use `key_config.name, key_config.value`

blarghmatey:
:+1:

PR #28, opened 2020-12-18
Added some mitxpro_edxapp resources

shaidar:
Mainly to create an RDS instance for xPRO edxapp and have that external service register itself with consul. In order to do that, this PR does the following:
- Adds pulumi-consul library
- Creates AWS secgroups that for RDS and have access to RDS
- Register service with consul


src/ol_infrastructure/applications/mitxpro/__main__.py
```diff
@@ -0,0 +1,216 @@
...
+            from_port=18040,
+            to_port=18040,
+            description="Xqueue access",
+        ),
+    ],
+    tags=aws_config.tags,
+    vpc_id=mitxpro_vpc["id"],
+)
+
+mitxpro_edxapp_security_group = ec2.SecurityGroup(
```
blarghmatey, line 81: Is there a reason that this is duplicated?

src/ol_infrastructure/applications/mitxpro/__main__.py
```diff
@@ -0,0 +1,216 @@
...
+    f"mitxpro-edx_worker-access-{env_suffix}",
+    name=f"mitxpro-edx_worker-access-{env_suffix}",
+    description="Access control to mitxpro_edx_worker",
+    ingress=[
+        ec2.SecurityGroupIngressArgs(
+            security_groups=[mitxpro_edxapp_security_group.id],
+            protocol="tcp",
+            from_port=18040,  # noqa: WPS432
+            to_port=18040,  # noqa: WPS432
+            description="Xqueue access",
```
blarghmatey, line 141: The worker doesn't run xqueue so there's no need to have this rule present.

src/ol_infrastructure/applications/mitxpro/__main__.py
```diff
@@ -0,0 +1,216 @@
...
+from pulumi_aws import ec2
+from pulumi_consul import Node, Service
+
+from ol_infrastructure.components.aws.database import OLAmazonDB, OLMariaDBConfig
+from ol_infrastructure.components.services.vault import (
+    OLVaultDatabaseBackend,
+    OLVaultMariaDatabaseConfig,
+)
+from ol_infrastructure.infrastructure.operations.consul import (
+    consul_server_security_group,
```
blarghmatey, line 12: Importing the security group definition won't work for using as a reference. We need to pass the ID of the resource generated, so we'll need to rely on stack outputs from the operations stack.

src/ol_infrastructure/applications/mitxpro/__main__.py
```diff
@@ -0,0 +1,216 @@
...
+mitxpro_edxapp_db_security_group = ec2.SecurityGroup(
+    f"mitxpro-edxapp-db-access-{env_suffix}",
+    name=f"mitxpro-edxapp-db-access-{env_suffix}",
+    description="Access from the mitxpro VPC to the mitxpro edxapp database",
+    ingress=[
+        ec2.SecurityGroupIngressArgs(
+            security_groups=[
+                mitxpro_edxapp_security_group.id,
+                mitxpro_edx_worker_security_group.id,
+                consul_server_security_group.id,
```
blarghmatey, line 157: As mentioned above, this will need to rely on retrieving the security group ID from the stack output of the `ol_infrastructure.infrastructure.operations` project for the appropriate stack.

src/ol_infrastructure/applications/mitxpro/__main__.py
```diff
@@ -0,0 +1,216 @@
...
+namespace = stack.rsplit(".", 1)[0]
+env_suffix = stack_name.lower()
+network_stack = StackReference(f"infrastructure.aws.network.{stack_name}")
+dns_stack = StackReference("infrastructure.aws.dns")
+mitodl_zone_id = dns_stack.require_output("odl_zone_id")
+mitxpro_vpc = network_stack.require_output("mitxpro_vpc")
+operations_vpc = network_stack.require_output("operations_vpc")
+mitxpro_environment = f"mitxpro-{env_suffix}"
+aws_config = AWSBase(
+    tags={"OU": "mitxpro", "Environment": mitxpro_environment},
```
blarghmatey, line 28: Might be good to add an `application` tag to this as well.

src/ol_infrastructure/applications/mitxpro/__main__.py
```diff
@@ -0,0 +1,216 @@
...
+                mitxpro_edxapp_security_group.id,
+                mitxpro_edx_worker_security_group.id,
+                consul_server_security_group.id,
+            ],
+            protocol="tcp",
+            from_port=3306,  # noqa: WPS432
+            to_port=3306,  # noqa: WPS432
+        )
+    ],
+    tags=aws_config.tags,
```
blarghmatey, line 164: We should add a `Name` tag as well using the `merged_tags` method.

`tags=aws_config.merged_tags({'Name': 'database-name-tag-value'})`

src/ol_infrastructure/applications/mitxpro/__main__.py
```diff
@@ -0,0 +1,216 @@
...
+    vpc_id=mitxpro_vpc["id"],
+)
+
+mitxpro_edxapp_db_config = OLMariaDBConfig(
+    instance_name=f"ol-mitxpro-edxapp-db-{env_suffix}",
+    password=get_config("mitxpro_edxapp:db_password"),
```
blarghmatey, line 170: The `get_config` method will happily allow you to not specify the corresponding stack value. We should get in the habit of using the `Config` class object and the associated methods such as `require_secret`, `require_object`, etc.

So for this it would by:
```python
from pulumi import Config

xpro_config = Config('mitxpro_edxapp')
...
password=mitxpro_config.require_secret('db_password')
...
db_purpose=mitxpro_config.require('db_purpose')
```

src/ol_infrastructure/applications/mitxpro/__main__.py
```diff
@@ -0,0 +1,216 @@
...
+    ),
+    **defaults(stack)["rds"],
+)
+mitxpro_edxapp_db = OLAmazonDB(mitxpro_edxapp_db_config)
+
+mitxpro_edxapp_db_vault_backend_config = OLVaultMariaDatabaseConfig(
+    db_name=mitxpro_edxapp_db_config.db_name,
+    mount_point=f"{mitxpro_edxapp_db_config.engine}-mitxpro-edxapp-{mitxpro_environment}",
+    db_admin_username=mitxpro_edxapp_db_config.username,
+    db_admin_password=get_config("mitxpro_edxapp:db_password"),
```
blarghmatey, line 185: Same as above using the `require_secret` method

src/ol_infrastructure/applications/mitxpro/__main__.py
```diff
@@ -0,0 +1,216 @@
...
+    tags=aws_config.tags,
+    vpc_id=mitxpro_vpc["id"],
+)
+
+mitxpro_edxapp_db_security_group = ec2.SecurityGroup(
+    f"mitxpro-edxapp-db-access-{env_suffix}",
+    name=f"mitxpro-edxapp-db-access-{env_suffix}",
+    description="Access from the mitxpro VPC to the mitxpro edxapp database",
+    ingress=[
+        ec2.SecurityGroupIngressArgs(
```
blarghmatey, line 153: This will also require an ingress rule to allow for the Vault cluster to communicate with the DB for generating secrets.

src/ol_infrastructure/applications/mitxpro/__main__.py
```diff
@@ -0,0 +1,216 @@
...
+    name="mysql",
+    port=3306,
+    checks=[
+        {
+            "check_id": "mitxpro_edxapp_db",
+            "interval": "10s",
+            "name": "mitxpro_edxapp_db",
+            "timeout": "60s",
+            "status": "passing",
+        }
```
blarghmatey, line 212: This check isn't actually doing any validation of the service. It should be using a TCP check type.

src/ol_infrastructure/applications/mitxpro/__main__.py
```diff
@@ -0,0 +1,216 @@
...
+    ],
+    tags=["rds", "mitxpro", "mitxpro_edxapp", mitxpro_environment],
+)
+
+export(
+    "mitxpro_edxapp",
+    {"rds_host": mitxpro_edxapp_db.db_instance.address},
```
blarghmatey, line 219: We should also export the security group IDs as part of the stack output.

src/ol_infrastructure/applications/mitxpro/Pulumi.applications.mitxpro.Dev.yaml
```diff
@@ -0,0 +1,13 @@
...
+config:
+  aws:region: us-east-1
+  consul:address: https://consul-mitxpro-qa.odl.mit.edu
+  consul:httpAuth:
+    secure: v1:OhuBxw6C/oJd3b6W:sfBXy+WtqpFD3x2ORyN3v+6ckliOM1JQCOIMUK/iyPQ0tbeHnMmMnPHPyaEDffGOrsXa8WlCBw==
+  consul:scheme: https
+  mitxpro_edxapp:db_password:
+    secure: v1:Ecp0f4H29kIL1EfA:lwHSpgyJJZb0b4Hazho8Amav7CpLBE01y5UDlaHRT1Vr5lBjP5PM23cmmm6yt5P+
+  mitxpro_edxapp:db_purpose: xpro_qa
+  vault:address: active.vault.service.consul
```
blarghmatey, line 12: We should leave this out of the stack configs in favor of using the environment variable so that it's easier to do things like SSH tunneling.

src/ol_infrastructure/applications/mitxpro/__main__.py
```diff
@@ -0,0 +1,220 @@
...
+    node="mysql",
+    name="mysql",
+    port=3306,
+    meta={
+        "external-node": True,
+        "external-probe": True,
+    },
+    checks=[
+        {
+            "check_id": "mitxpro_edxapp_db",
```
blarghmatey, line 207: It would be good to pass this as an instance of `ServiceCheckArgs` to allow for type enforcement.

blarghmatey (changes requested):

src/ol_infrastructure/applications/mitxpro/__main__.py
shaidar: What's the purpose of this?

blarghmatey: It adds the `Name` tag to the already defined tags so that we can add arbitrary tags in addition to the baseline values and return the merged dictionary.

src/ol_infrastructure/applications/mitxpro/__main__.py
shaidar: I call the secgroup id [here](https://github.com/mitodl/ol-infrastructure/blob/4aa2c65b3cdee4fcf1c95a6c477a3bf29eb756ae/src/ol_infrastructure/applications/mitxpro/__main__.py#L157). Wouldn't that work?

blarghmatey: No, because that will cause it to try to execute that whole module and then return the output of the `id` object. Despite it being "just Python" we can't import between projects like that. The only interface for data exchange across concrete projects is via the stack references.

blarghmatey: Essentially, by importing from the Consul module it will try to deploy all of the resources defined within it.

shaidar: Isn't [this](https://github.com/mitodl/ol-infrastructure/blob/95aca39329813d610dbfd1efc7e51c4dd06105c1/src/ol_infrastructure/applications/edx/__main__.py#L57) though basically doing the same?

blarghmatey: The difference is that for that import it's within the same project and it is intended to deploy all of those resources. If we import it from a different project then it will cause all of those resources to be deployed again but with the wrong stack information.

shaidar: Makes sense, but the stack that we're trying to get the secgroup.id from (consul in this case) has to be exporting the `consul_server_security_group`, right?

blarghmatey: Correct, which it already does. https://github.com/mitodl/ol-infrastructure/blob/main/src/ol_infrastructure/infrastructure/operations/__main__.py#L16

blarghmatey: https://github.com/mitodl/ol-infrastructure/blob/main/src/ol_infrastructure/infrastructure/operations/consul.py#L166-L169

src/ol_infrastructure/applications/mitxpro/__main__.py
```diff
@@ -8,74 +7,27 @@
...
-            to_port=18040,
-            description="Xqueue access",
-        ),
-    ],
-    tags=aws_config.tags,
-    vpc_id=mitxpro_vpc["id"],
+    tags={
+        "OU": "mitxpro",
+        "Environment": mitxpro_environment,
+        "application": Apps.mitxpro_edx.name,
```
blarghmatey, line 29: I'm pretty sure this should be `.value` if you want the string. Otherwise I believe it will return `"mitxpro_edx"`

src/ol_infrastructure/applications/mitxpro/__main__.py
```diff
@@ -132,47 +84,45 @@
...
 )
 
 mitxpro_edxapp_db_config = OLMariaDBConfig(
     instance_name=f"ol-mitxpro-edxapp-db-{env_suffix}",
-    password=get_config("mitxpro_edxapp:db_password"),
+    password=mitxpro_config.require_secret("db_password"),
     subnet_group_name=mitxpro_vpc["rds_subnet"],
     security_groups=[mitxpro_edxapp_db_security_group],
-    tags=aws_config.tags,
+    tags=aws_config.merged_tags({"Name": "database-name-tag-value"}),
```
blarghmatey, line 123: Copy :spaghetti: 

I had just put in the value as a placeholder :)

src/ol_infrastructure/applications/mitxpro/__main__.py
```diff
@@ -8,74 +7,27 @@
...
-from ol_infrastructure.lib.ol_types import AWSBase
+from ol_infrastructure.lib.ol_types import Apps, AWSBase
 from ol_infrastructure.lib.stack_defaults import defaults
 
 stack = get_stack()
 stack_name = stack.split(".")[-1]
 namespace = stack.rsplit(".", 1)[0]
 env_suffix = stack_name.lower()
 network_stack = StackReference(f"infrastructure.aws.network.{stack_name}")
+consul_stack = StackReference("infrastructure.operations.consul")
```
blarghmatey, line 18: There isn't a specific Consul stack as it's part of the `infrastructure.operations` project. It should be `operations_stack = StackReference("infrastructure.operations.{environment}.{stack_name}")` and define the proper environment to match the available stacks.

src/ol_infrastructure/applications/mitxpro/__main__.py
```diff
@@ -94,11 +95,8 @@
     description="Access from the mitxpro and operations VPC to the mitxpro edxapp database",
     ingress=[
         ec2.SecurityGroupIngressArgs(
-            security_groups=[
-                mitxpro_edxapp_security_group.id,
-                mitxpro_edx_worker_security_group.id,
```
blarghmatey: We can maintain the security groups block, we just want to add the operations CIDR block for allowing traffic from Vault.

src/ol_infrastructure/applications/mitxpro/__main__.py
```diff
@@ -128,12 +126,38 @@
 )
 mitxpro_edxapp_db = OLAmazonDB(mitxpro_edxapp_db_config)
 
+hyphenated_db_purpose = (mitxpro_config.require("db_purpose")).replace("_", "-")
+
+edx_role_statments = mysql_sql_statements.update(
+    {
+        f"edxapp-{hyphenated_db_purpose}": {
+            "create": "CREATE USER '{{{{name}}}}'@'%' IDENTIFIED BY '{{{{password}}}}';"
+            "GRANT SELECT, INSERT, UPDATE, DELETE, CREATE, INDEX, DROP, ALTER, REFERENCES"  # noqa: Q000
```
blarghmatey, line 135: Shouldn't this be scoped to the specific DB name?

src/ol_infrastructure/applications/mitxpro/__main__.py
```diff
@@ -128,12 +126,38 @@
...
+            "CREATE TEMPORARY TABLES, LOCK TABLES ON {{{{ app }}}}.* TO '{{{{name}}}}'@'%';",
+            "revoke": "DROP USER '{{{{name}}}}';",
+        },
+        f"edxapp-{hyphenated_db_purpose}": {
```
blarghmatey, line 139: This is a duplicate of the previous statement. Is this intended to be for the CSMH database?

src/ol_infrastructure/applications/mitxpro/__main__.py
```diff
@@ -128,12 +126,38 @@
...
+            "create": "CREATE USER '{{{{name}}}}'@'%' IDENTIFIED BY '{{{{password}}}}';"
+            "GRANT SELECT, INSERT, UPDATE, DELETE, CREATE, INDEX, DROP, ALTER, REFERENCES"  # noqa: Q000
+            "CREATE TEMPORARY TABLES, LOCK TABLES ON {{{{ app }}}}.* TO '{{{{name}}}}'@'%';",
+            "revoke": "DROP USER '{{{{name}}}}';",
+        },
+        f"xqueue-{hyphenated_db_purpose}": {
+            "create": "CREATE USER '{{{{name}}}}'@'%' IDENTIFIED BY '{{{{password}}}}';"
+            "GRANT SELECT, INSERT, UPDATE, DELETE, CREATE, INDEX, DROP, ALTER, REFERENCES"  # noqa: Q000
```
blarghmatey, line 147: Should we scope this to the specific DB/schema? e.g. `ON xqueue_{ db_purpose }.*`

shaidar: This would be `xqueue_{ mitxpro_config.db_purpose }.*` given that `db_purpose` is defined as part of the mitxpro_config, right?

blarghmatey: It would be `xqueue_{mitxpro_config.require('db_purpose')}.*`

blarghmatey (approved):

blarghmatey:
Looks good aside from the merge conflict. :+1:

PR #30, opened 2020-12-22
Minor cleanup to Consul agent security groups

blarghmatey:
Taking advantage of new parameter to allow for reducing the boilerplate in defining the consul agent security group.


shaidar:
👍

PR #33, opened 2021-01-13
Extracting boilerplate stack parsing into a helper function

blarghmatey:
Extracted the boilerplate setup of stack name, env_suffix, etc. into a helper function and replaced all current uses of the boilerplate.


shaidar:
👍

shaidar (approved):

PR #35, opened 2021-01-15
Added Vault PKI project

shaidar:
Initial work to get intermediate Vault PKI configured for each env and business unit. The cert for those vault PKI's would be signed by an intermediate vault CA which is also part of this configuration and that intermediate CA is signed by an offline root CA.


src/ol_infrastructure/infrastructure/vault/pki/__main__.py
```diff
@@ -0,0 +1,84 @@
...
+# The files should be generated by the offline Root CA and copied locally prior to
+# running this. More information here - https://docs.odl.mit.edu/en/platform/security/vault-pki
+try:
+    with open("/tmp/int_ca_pe_bundle_file.pem") as int_ca_pem_bundle_file, open(
+        "/tmp/int_ca_cert_file"
+    ) as int_ca_cert_file:
+        pki_intermediate_ca_config = OLVaultPKIIntermediateCABackendConfig(
+            max_ttl=TWELVE_MONTHS,
+            default_ttl=TWELVE_MONTHS,
+            intermediate_ca_pem_bundle=int_ca_pem_bundle_file,
```
blarghmatey, line 31: Shouldn't this be doing a `.read()` to pass the contents of the file rather than just the file object?

src/ol_infrastructure/infrastructure/vault/pki/__main__.py
```diff
@@ -0,0 +1,84 @@
...
+business_unit = env_config.get("business_unit") or "operations"
+
+# The files should be generated by the offline Root CA and copied locally prior to
+# running this. More information here - https://docs.odl.mit.edu/en/platform/security/vault-pki
+try:
+    with open("/tmp/int_ca_pe_bundle_file.pem") as int_ca_pem_bundle_file, open(
+        "/tmp/int_ca_cert_file"
+    ) as int_ca_cert_file:
+        pki_intermediate_ca_config = OLVaultPKIIntermediateCABackendConfig(
+            max_ttl=TWELVE_MONTHS,
```
blarghmatey, line 29: Is this the TTL for the leases that it generates, or the TTL for the CA? If it's for the CA we should probably make it longer.

src/ol_infrastructure/infrastructure/vault/pki/__main__.py
```diff
@@ -0,0 +1,84 @@
...
+        default_ttl=SIX_MONTHS,
+        key_usages=["DigitalSignature", "KeyAgreement", "KeyEncipherment"],
+        allowed_domains=[f"fluentd.*.{business_unit.value}-{stack_info.env_suffix}"],
+        cert_type="client",
+    )
+
+    pki_fluentd_client_role = OLVaultPKIIntermediateRole(pki_fluentd_client_role_config)
+
+export("pki_intermediate_ca", pki_intermediate_ca.name)
+export("pki_intermediate_env", pki_intermediate_env.name)
```
blarghmatey, line 82: This is only exporting the last instance of the env that was generated, rather than each one that was created.

src/ol_infrastructure/infrastructure/vault/pki/__main__.py
```diff
@@ -0,0 +1,84 @@
...
+export("pki_fluentd_server_role", pki_fluentd_server_role.name)
+export("pki_fluentd_client_role", pki_fluentd_client_role.name)
```
blarghmatey, line 84: Same for these. Maybe just export a list of intermediate mounts created, and a list of role definitions?

src/ol_infrastructure/infrastructure/vault/pki/__main__.py
```diff
@@ -0,0 +1,84 @@
...
+from ol_infrastructure.lib.ol_types import BusinessUnit
+from ol_infrastructure.lib.pulumi_helper import parse_stack
+from ol_infrastructure.lib.vault import VaultPKIKeyTypeBits
+
+SIX_MONTHS = 60 * 60 * 24 * 30 * 6  # noqa: WPS432
+TWELVE_MONTHS = 60 * 60 * 24 * 30 * 12  # noqa: WPS432
+
+env_config = Config("environment")
+stack_info = parse_stack()
+business_unit = env_config.get("business_unit") or "operations"
```
blarghmatey, line 20: This doesn't appear to get used anywhere.

src/ol_infrastructure/infrastructure/vault/pki/__main__.py
```diff
@@ -0,0 +1,84 @@
...
+
+# The files should be generated by the offline Root CA and copied locally prior to
+# running this. More information here - https://docs.odl.mit.edu/en/platform/security/vault-pki
+try:
+    with open("/tmp/int_ca_pe_bundle_file.pem") as int_ca_pem_bundle_file, open(
```
blarghmatey, line 25: There's not really much benefit to a context manager here. You can just inline the `open(fname).read()` calls to reduce a bit of visual noise.

PR #38, opened 2021-02-16
Vault PKI implementation 

shaidar:
This fixes any issues with creating the pki mounts and getting certs in place. The one remaining issue that is unsolved is the problem where the app pki mount doesn't have the complete ca chain. When comparing the new mount with the old mounts that I've created manually, the old mounts had two certs (Root CA which is offline, and online intermediate CA). When doing this through Pulumi, the mount only seems to have the cert for the online intermediate CA and not the full chain. So, more work is needed to identify the problem and resolve it. I reached out to Pulumi folks through Slack, however didn't get much help there and I glanced over some of the Pulumi code, which uses the Pulumi Terraform bridge, and didn't run across anything obvious. My theory is that something's truncating the cert chain and only reading in one cert but have not been able to prove it so far.


src/ol_infrastructure/components/services/vault.py
```diff
@@ -300,30 +300,80 @@ def __init__(
...
         resource_options = ResourceOptions(parent=self).merge(opts)  # type: ignore
 
+        # Create the pki-intermediate-ca endpoint / mount
         self.pki_intermediate_ca_backend = Mount(
             "pki-intermediate-ca",
             opts=resource_options,
             path="pki-intermediate-ca",
             type="pki",
-            description="Backend to create certs for pki-intermediate-env backends",
+            description="Backend to create certs for pki-intermediate-{env} backends",
```
blarghmatey, line 309: Missing the `f` prefix if this is intended to be a formatted string

src/ol_infrastructure/components/services/vault.py
```diff
@@ -362,26 +416,34 @@ def __init__(
...
             "ol:services:Vault:PKI:IntermediateEnvBackendConfig",
-            backend_config.environment_name,
+            f"pki-intermediate-{backend_config.environment_name}",
             None,
             opts,
         )
 
-        resource_options = ResourceOptions(parent=self).merge(opts)  # type: ignore
+        resource_options = ResourceOptions(
+            parent=self,
```
blarghmatey, line 425: I'm curious what this is needed for? Not questioning the validity, just wondering how it manifests.

src/ol_infrastructure/components/services/vault.py
```diff
@@ -394,28 +456,59 @@ def __init__(
...
             )
         )
 
-        # Sign genereated CSR for pki-intermediate-{env} backend by pki-intermediate-ca
+        # Create a default issuer role for pki-intermediate-{env}
+        self.pki_intermediate_environment_default_issuer_config = OLVaultPKIIntermediateRoleConfig(  # noqa: E501
+            pki_intermediate_backend=self.pki_intermediate_environment_backend,
+            role_name="default-issuer",
+            key_config=VaultPKIKeyTypeBits.rsa,
+            allowed_domains=["mit.edu"],
```
blarghmatey, line 468: We may also want to allow `.service.consul` domains, and optionally `.service.<env_name>.consul`

src/ol_infrastructure/substructure/vault/pki/__main__.py
```diff
@@ -19,70 +17,35 @@
...
-    pki_intermediate_ca_config = OLVaultPKIIntermediateCABackendConfig(
-        max_ttl=TWELVE_MONTHS,
-        default_ttl=TWELVE_MONTHS,
-        intermediate_ca_pem_bundle=int_ca_pem_bundle_file,
-        intermediate_ca_cert=int_ca_cert_file,
-    )
-except OSError:
-    raise OSError("Could not find or read cert files")  # noqa: B904
+acmpca_stack = StackReference("infrastructure.aws.private_ca")
+root_ca_arn = acmpca_stack.require_output("root_ca").apply(lambda rca: f"{rca['arn']}")
```
blarghmatey, line 21: This shouldn't be necessary. It should be possible to do `root_ca = acmpca_stack.require_output("root_ca")` and then on line 39 do `acmpca_rootca_arn=root_ca["arn"]` (I think...), but if this works then :shrug:

src/ol_infrastructure/substructure/vault/pki/__main__.py
```diff
@@ -19,70 +17,35 @@
...
 
+if Config("vault_server").get("env_namespace"):
```
blarghmatey, line 23: This shouldn't need the conditional. We can just run the provider setup unconditionally. That pattern was a temporary accommodation while we migrated to the automatic provider.

blarghmatey (approved):
Some minor comments/questions but overall looks great! Good job @shaidar and @Ardiea !

src/ol_infrastructure/components/services/vault.py
Ardiea: It is not intended to be a f-string.

src/ol_infrastructure/components/services/vault.py
Ardiea: I don't know about the `parent=self` thing, that was there when I got here. But the dependency on the `pki_intermediate_ca_default_issuer` is important. Pulumi wasn't able to figure out that `pki_intermediate_ca` needs to be there and complete before it can start doing _anything_ with any of the `pki_intermediate_{env}` resources.  I guess technically it could do the backends and issue the CSRs but it definitely cannot signed w/o this dependency. Seems safer to just block all `pki_intermediate_{env}` activities until `pki_intermediate_ca` is complete.

src/ol_infrastructure/components/services/vault.py
```diff
@@ -298,30 +300,80 @@ def __init__(
...
             max_lease_ttl_seconds=backend_config.max_ttl,
             default_lease_ttl_seconds=backend_config.default_ttl,
         )
 
+        # Generate CSR for pki-intermediate-ca backend
+        self.pki_intermediate_ca_cert_request = (
+            pkisecret.SecretBackendIntermediateCertRequest(
+                "pki-intermediate-ca-csr",
+                backend=self.pki_intermediate_ca_backend.id,
+                common_name="pki-intermediate-ca " "Intermediate Authority",
```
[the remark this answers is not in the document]
Ardiea: 🤷

src/ol_infrastructure/components/services/vault.py
shaidar: I also think we might need to add another attribute if we were to allow subdomains - https://developer.hashicorp.com/vault/api-docs/secret/pki#allowed_domains

Ardiea: Ooo Good catch! 

But, I kind of figured the default-issuer wouldn't really be utilized beyond this `bootstrapping` of the PKI itself? Seems to make more sense to create a new `role` or `issuer` for the use cases as the appear. But I defer to wiser minds if you guys disagree.

src/ol_infrastructure/substructure/vault/pki/__main__.py
```diff
@@ -17,11 +17,10 @@
 stack_info = parse_stack()
 pki_intermediate_export = {}
 
-acmpca_stack = StackReference("infrastructure.aws.private_ca")
-root_ca_arn = acmpca_stack.require_output("root_ca").apply(lambda rca: f"{rca['arn']}")
+root_ca = StackReference("infrastcuture.aws.private_ca").require_output("root_ca")
```
blarghmatey, line 20:
```suggestion
root_ca = StackReference("infrastucture.aws.private_ca").require_output("root_ca")
```

src/ol_infrastructure/components/services/vault.py
blarghmatey: Yeah, I remembered after posting this that the reason that's there is so that all of the resources created inside the component get properly associated with the containing component resource.

PR #105, opened 2021-04-22
Fix elasticsearch mounts

shaidar:
This just fixes a few things that Mark had already been working on regarding the ES7 deployment.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -50,6 +50,7 @@ repos:
       args:
         - "--config=setup.cfg"
         - '--extend-ignore=D1'
+        - '--max-string-usages=10'
 - repo: https://github.com/pre-commit/mirrors-mypy
   rev: v0.812
   hooks:
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -50,7 +50,6 @@ repos:
       args:
         - "--config=setup.cfg"
         - '--extend-ignore=D1'
-        - '--max-string-usages=10'
 - repo: https://github.com/pre-commit/mirrors-mypy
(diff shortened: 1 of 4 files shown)


.pre-commit-config.yaml
```diff
@@ -50,6 +50,7 @@ repos:
       args:
         - "--config=setup.cfg"
         - '--extend-ignore=D1'
+        - '--max-string-usages=10'
```
blarghmatey, line 53: This should be removed. Any configuration settings for flake8 should be managed via setup.cfg

src/ol_infrastructure/infrastructure/elasticsearch/__main__.py
```diff
@@ -125,16 +127,16 @@
     ingress=[
         ec2.SecurityGroupIngressArgs(
             protocol="tcp",
-            from_port=9200,
-            to_port=9200,
+            from_port=9200,  # noqa: WPS432
```
blarghmatey, line 130: For these kinds of magic numbers I actually created a module (I think called magic_numbers) under utils to hold those values and then import and use the descriptive names in these locations. That's preferable to sprinkling `noqa` statements everywhere we use a number, at least for well-known ports, etc.

blarghmatey (changes requested):

src/ol_infrastructure/infrastructure/elasticsearch/__main__.py
shaidar: That util is in the configuration repo, so leaving this as is for now.

blarghmatey (approved):

blarghmatey:
:+1:

PR #119, opened 2021-04-29
Fixed issue with adding salt grains

shaidar:
Additional salt grains were being populated in salt config file instead of being added as a grain to the salt-minion. This should address that issue.


src/ol_infrastructure/lib/aws/ec2_helper.py
```diff
@@ -354,6 +358,9 @@ def _build_cloud_config_string(keys) -> str:  # noqa: WPS430
         salt_config["salt_minion"]["conf"].update(  # type: ignore
             additional_salt_config or {}
         )
+        salt_config["salt_minion"]["grains"].update(  # type: ignore
+            additional_salt_config or {}
```
blarghmatey, line 362: Copy :spaghetti:

PR #123, opened 2021-04-30
Allow access from VPC to forum service

shaidar:
Added rule to secgroup so that the worker nodes in the VPC can access the forum service running on the edxapp nodes.


blarghmatey:
:+1:

PR #126, opened 2021-05-03
Add OCW site project for website buckets

markbreedlove:


src/ol_infrastructure/applications/ocw_site/Pulumi.applications.ocw_site.Production.yaml
```diff
@@ -0,0 +1,4 @@
+secretsprovider: awskms://alias/infrastructure-secrets-qa  # pragma: allowlist secret
```
blarghmatey, line 1: This stack got instantiated with the wrong secrets provider. You'll likely need to destroy and regenerate the stack. It should be `infrastructure-secrets-production` for the `.Production` stack.

src/ol_infrastructure/applications/ocw_site/Pulumi.yaml
```diff
@@ -0,0 +1,5 @@
+name: ol-infrastructure-ocw_site-application
```
blarghmatey, line 1: For the name we should stick with `-` over `_`.

src/ol_infrastructure/applications/ocw_site/Pulumi.yaml
```diff
@@ -0,0 +1,5 @@
...
+runtime: python
+description: OCW next-generation site contents stored in S3 buckets and hosted on the CDN
```
blarghmatey, line 3: We should probably avoid using the `next-generation` terminology as much as possible.

src/ol_infrastructure/applications/ocw_site/__main__.py
```diff
@@ -0,0 +1,80 @@
...
+from ol_infrastructure.lib.aws.iam_helper import lint_iam_policy
+from ol_infrastructure.lib.ol_types import AWSBase
+from ol_infrastructure.lib.pulumi_helper import parse_stack
+
+ocw_site_config = Config("ocw_site")
+stack_info = parse_stack()
+aws_config = AWSBase(
+    tags={
+        "OU": "open-courseware",
+        "environment": f"applications_{stack_info.env_suffix}",
```
blarghmatey, line 14: The environment tag name should be capitalized.

src/ol_infrastructure/applications/ocw_site/__main__.py
```diff
@@ -0,0 +1,80 @@
...
+
+policy_description = (
+    "Access controls for the CDN to be able to read from the"
+    f"{stack_info.env_suffix} website buckets"
+)
+s3_bucket_iam_policy = iam.Policy(
+    f"ocw-site-{stack_info.env_suffix}-policy",
+    description=policy_description,
+    path=f"/ol-applications/ocw-site/{stack_info.env_suffix}/",
+    name_prefix="aws-permissions-",
```
blarghmatey, line 47: This should be more descriptive to make it easy to visually identify its association with OCW course content buckets

src/ol_infrastructure/applications/ocw_site/__main__.py
```diff
@@ -0,0 +1,80 @@
...
+                    ],
+                }
+            ],
+        },
+        stringify=True,
+    ),
+)
+
+role_name = f"ocw-site-{stack_info.env_suffix}"
+ocw_site_vault_backend_role = aws.SecretBackendRole(
```
blarghmatey, line 72: What's the motivation for registering the Vault role? Are we using this anywhere?

markbreedlove: I was just copying from the ocw_studio role. I just want to create an IAM role, however I do that -- or should I just stop after creating a policy?

src/ol_infrastructure/applications/ocw_site/__main__.py
markbreedlove: I was taking the ocw_studio role as an example for this. I'm not sure what the name_prefix signifies. Can you suggest something?

src/ol_infrastructure/applications/ocw_site/__main__.py
markbreedlove: Fixed in f5fc689

src/ol_infrastructure/applications/ocw_site/__main__.py
blarghmatey: That's the policy name that is created and is viewable in the IAM console. I would likely use something like`ocw-site-content-read-only-{stack_info.env_suffix}`

src/ol_infrastructure/applications/ocw_site/Pulumi.yaml
markbreedlove: Fixed in a3fae49

src/ol_infrastructure/applications/ocw_site/__main__.py
blarghmatey: If you want to create an IAM user then you would use https://www.pulumi.com/docs/reference/pkg/aws/iam/user/, for a role you would use https://www.pulumi.com/docs/reference/pkg/aws/iam/role/

The Vault role registers a template in Vault for generating an IAM user and isn't relevant for the site content since we won't be populating the Fastly config from Vault.

src/ol_infrastructure/applications/ocw_site/Pulumi.yaml
markbreedlove: I think this got filled in by the `copier` command. I'd put in an underscore because the other project directories under `applications` have underscores. Is there some way around having it fill it in that way and keep consistency with the other directory names?

src/ol_infrastructure/applications/ocw_site/Pulumi.applications.ocw_site.Production.yaml
markbreedlove: Oh, OK. I can recreate it. I was just reusing shell history to create the production environment.

markbreedlove: Hopefully fixed in d622b55

I recreated the stack.

src/ol_infrastructure/applications/ocw_site/Pulumi.yaml
blarghmatey: Yeah, that's fine, I just meant that we should edit the name attribute in the yaml. Not really significant, just a nit pick.

src/ol_infrastructure/applications/ocw_site/__main__.py
markbreedlove: OK. Is the one for OCW studio OK, then? Should we change it, too?  https://github.com/mitodl/ol-infrastructure/blob/d622b5572955943925d1b3dc1e90e381eae93815/src/ol_infrastructure/applications/ocw_studio/__main__.py#L61

blarghmatey: We should probably change that one as well, but we can tackle that later.

src/ol_infrastructure/applications/ocw_site/__main__.py
markbreedlove: Could I just leave out the creation of the user and role? We already have an IAM user for Fastly and all we need is for the policy that I'm creating to be attached to that user eventually.

blarghmatey: That's fine for now, though I'd prefer to have user/credential management in code rather than manually created and floating around relying on tribal knowledge.

src/ol_infrastructure/applications/ocw_site/__main__.py
markbreedlove: Name changed in 85bf03b (for this project)

src/ol_infrastructure/applications/ocw_site/Pulumi.yaml
markbreedlove: Sure, that's fine. I've updated the `name` property.

src/ol_infrastructure/applications/ocw_site/__main__.py
markbreedlove: OK. Commit 04fb3bf takes out the creation of the role. I was going to add a user and attach the policy to it, but I wasn't clear on a few things like the path, and thought maybe we want to revisit it because we have existing Fastly configurations with a particular access key and secret key and those don't have configuration management.

blarghmatey:
:+1:

PR #141, opened 2021-05-10
Automatically restart services when config files change

blarghmatey:
Adds a deploy function to install file watchers that restart a service when config
files change


src/bilder/components/baseline/templates/service_restarter.service.j2
```diff
@@ -0,0 +1,12 @@
+[Unit]
+Description=Restart {{ service_name }} when watched files change
+After=network.target
+StartLimitIntervalSec=10
+StartLimitBurst=5
+
+[Service]
+Type=oneshot
+ExecStart=/usr/bin/systemctl restart {{ service_name }}.service
```
shaidar, line 9: Would a reload work as opposed to restart?

blarghmatey: In some cases, but for a first pass this is sufficient. The problem with a reload is that if the service is dead then the reload will fail.

PR #147, opened 2021-05-13
Fixing Concourse web and worker configs

blarghmatey:
- Added more comprehensive set of model parameters for Concourse configuration
- Updated worker to use btrfs volume storage
- Fixed Consul healthcheck configuration for web instance

### DIFF
--- a/src/bilder/images/concourse/deploy.py
+++ b/src/bilder/images/concourse/deploy.py
@@ -36,6 +36,7 @@ from bilder.components.hashicorp.steps import (
 )
 from bilder.components.hashicorp.vault.models import (
     Vault,
+    VaultAgentCache,
     VaultAgentConfig,
     VaultAutoAuthAWS,
     VaultAutoAuthConfig,
@@ -97,10 +98,13 @@ concourse_config_map = {
             "{{ end }}"
         ),
         vault_url="http://localhost:8200",
+        vault_client_token="this-token-gets-overridden-by-the-vault-agent",
+        vault_path_prefix="/secret-concourse",
     ),
     CONCOURSE_WORKER_NODE_TYPE: partial(
         ConcourseWorkerConfig,
-        container_runtime="containerd",
+        container_runtime="guardian",
+        baggageclaim_driver="btrfs",
     ),
 }
 concourse_config = concourse_config_map[node_type]()
@@ -168,7 +172,7 @@ vault_template_map = {
 }
 
 # Install Concourse
-install_baseline_packages()
+install_baseline_packages(packages=["curl", "btrfs-progs"])
 concourse_install_changed = install_concourse(concourse_base_config)
 concourse_config_changed = configure_concourse(concourse_config)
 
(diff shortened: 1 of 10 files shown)


src/ol_infrastructure/applications/concourse/__main__.py
```diff
@@ -249,6 +246,17 @@
     vpc_id=ops_vpc_id,
 )
 
+ec2.SecurityGroupRule(
+    "concourse-worker-access-from-concourse-web",
+    security_group_id=concourse_worker_security_group.id,
+    source_security_group_id=concourse_web_security_group.id,
+    protocol="tcp",
+    from_port=0,
+    to_port=65535,
```
shaidar, line 255: Does access have to be that open?

src/bilder/images/concourse/deploy.py
```diff
@@ -168,7 +172,7 @@
 }
 
 # Install Concourse
-install_baseline_packages()
+install_baseline_packages(packages=["curl", "btrfs-progs"])
```
shaidar, line 175: Why btrfs?

src/bilder/images/concourse/deploy.py
```diff
@@ -96,11 +97,14 @@
...
         ),
+        vault_url="http://localhost:8200",
+        vault_client_token="this-token-gets-overridden-by-the-vault-agent",
+        vault_path_prefix="/secret-concourse",
     ),
     CONCOURSE_WORKER_NODE_TYPE: partial(
         ConcourseWorkerConfig,
-        ephemeral=True,
-        container_runtime="containerd",
+        container_runtime="guardian",
```
shaidar, line 106: Why guardian?

src/bilder/images/concourse/deploy.py
```diff
@@ -96,11 +97,14 @@
             "{{ .Data.data.github_client_secret }}"
             "{{ end }}"
         ),
+        vault_url="http://localhost:8200",
```
shaidar, line 100: Isn't this a magic number?

src/bilder/images/concourse/deploy.py
blarghmatey: Because it's the recommended volume filesystem and because the fallback default of overlay has performance issues.

src/bilder/images/concourse/deploy.py
blarghmatey: Because it is the current default and when trying to use containerd it led to the worker failing to start up. Containerd is the new preferred runtime, but we'll have to do some additional debugging to make it work.

src/bilder/images/concourse/deploy.py
blarghmatey: Sort of. I'll extract it.

src/ol_infrastructure/applications/concourse/__main__.py
blarghmatey: Most likely not, I'll test with putting it back to using port 2222, but when I was working through getting things working it seemed like there might have been some additional port issues. That being said, it still only allows traffic between the web and worker instances and not any other sources.

shaidar:
Didn't go through every line of the `models.py` but looks good overall. As far as the sign and verify pipeline, I'm not sure what I should be looking for as I have yet to spend time on that.

blarghmatey:
The pipeline works so no particular review needed there. You can view it at https://cicd.odl.mit.edu/teams/main/pipelines/sign-and-verify

PR #156, opened 2021-05-18
Add OCW build for Concourse

markbreedlove:
This adds a Concourse pipeline for building just the "main" pages of the OCW site and the page assets.
I'm mainly looking for feedback to make sure I'm on the right track. My main issue is the Rclone resource. It's probably OK, because I can get it to show the files in the correct directory `ocw-www/dist` and it seems to be trying to sync them; but it retries repeatedly because it doesn't have any credentials set up. It's intended to work with an IAM instance profile. The real way to test this will be to put it on our QA concourse server.

The build seems to work and populates the `ocw-www/dist` directory as I expected it to.

### DIFF
--- a/pipelines/ocw/pipeline.yml
+++ b/pipelines/ocw/pipeline.yml
@@ -74,5 +74,7 @@ jobs:
           source: ocw-www/dist/
           destination:
             - dir: s3-remote:((ocw-site-bucket))
+            - command: sync
             - args:
-              - "--dry-run"
+              - "--exclude"
+              - "courses"
--- a/pipelines/ocw/pipeline.yml
+++ b/pipelines/ocw/pipeline.yml
@@ -77,4 +77,4 @@ jobs:
               command: sync
               args:
                 - "--exclude"
-                - "courses"
+                - "/courses/**"
--- a/pipelines/ocw/pipeline.yml
+++ b/pipelines/ocw/pipeline.yml
@@ -37,7 +37,9 @@ jobs:
     plan:
       - get: ol-infrastructure
       - get: ocw-www
+        trigger: true
       - get: ocw-hugo-themes
+        trigger: true
       - task: copy-go-mod-file
         config:
           platform: linux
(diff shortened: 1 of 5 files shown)


pipelines/ocw/pipeline.yml
```diff
@@ -0,0 +1,76 @@
...
+        env_auth = true
+        region = us-east-1
+
+jobs:
+  - name: build
+    serial: true
+    plan:
+      - get: ol-infrastructure
+      - get: ocw-www
+      - get: ocw-hugo-themes
```
blarghmatey, line 40: It's probably worth having this and the ocw-www repositories trigger the pipeline build. As this is defined currently it will only build if it gets executed manually, unless we set up webhooks from GitHub to trigger these resources.

blarghmatey:
Visually this looks fine. Go ahead and set it up on the QA or production Concourse instances. In general you should feel free to set and delete pipelines as needed to verify them without having to wait for PRs.

pipelines/ocw/pipeline.yml
markbreedlove: :+1: I was planning to add `trigger: true` at some point, when I could be confident about it running automatically.

markbreedlove: I've added `trigger: true` in 71d76bdff3dbe3dbe4e28533bd67b7f88c33494f

markbreedlove:
This is looking pretty good now on `cicd-qa`. I can confirm the following:

* the `sync` operation removes deleted files, but ...
* rclone does not delete the `courses` directory if it already exists
* rclone updates files that need updating and doesn't copy anything else

The only next step is to have two S3 destination buckets, one for the "draft" site and one for the "live" one.

blarghmatey:
Looks good :+1:

markbreedlove:
There's just one more thing: clearing the Fastly cache. Should I do that in this PR?

blarghmatey:
Save that for another step since we don't actually have Fastly set up with those buckets yet.

PR #170, opened 2021-05-21
Add pipeline to build OCW build Docker image

markbreedlove:
This pipeline checks the `node:14` Docker image on Docker Hub every hour and builds another image of `mitodl/ocw-course-publisher:latest` and uploads it to Docker Hub.

I've tested it in a localhost Concourse installation using credentials passed on the commandline to `fly set-pipeline`. I then set the Docker Hub credential variable names to match the ones that I'm seeing in the `sign-and-verify.yml` pipeline. Should I assume that the `dockerhub.username` and `dockerhub.password` variables will be picked up and filled in automatically if I install this pipeline on our Concourse server?

If that works, then I can put this pipeline on our Concourse so we can try it.

I want to make a subsequent change to alter the name of the docker image. The name `ocw-course-publisher` is not really correct. It is really just Hugo plus Node. After I rename the image, I intend to issue a PR for `ocw-course-hugo-theme` to update [its Dockerfile](https://github.com/mitodl/ocw-course-hugo-theme/blob/888e39659f99e200d2b3079b10d0d33467d4de52/docker/Dockerfile) to inherit this docker image with `FROM mitodl/node-hugo:latest` so that the development image is in sync with what we're using to build the site.

### DIFF
--- a/pipelines/ocw/docker-image-pipeline.yml
+++ b/pipelines/ocw/docker-image-pipeline.yml
@@ -14,8 +14,8 @@ resources:
     - name: ocw-course-publisher-image
       type: docker-image
       source:
-          username: ((docker-hub-username))
-          password: ((docker-hub-password))
+          username: ((dockerhub.username))
+          password: ((dockerhub.password))
           repository: mitodl/ocw-course-publisher
           tag: latest
 
--- a/pipelines/ocw/docker-image-pipeline.yml
+++ b/pipelines/ocw/docker-image-pipeline.yml
@@ -5,6 +5,8 @@ resources:
       source:
           uri: https://github.com/mitodl/ol-infrastructure.git
           branch: ((ol-infrastructure-git-ref))
+          paths:
+              - dockerfiles/ocw/node-hugo/Dockerfile
     - name: node-image
       type: docker-image
       check_every: 1h
@@ -28,6 +30,7 @@ jobs:
           - get: node-image
             trigger: true
           - get: ol-infrastructure
+            trigger: true
           - put: ocw-course-publisher-image
             params:
                 build: ol-infrastructure/dockerfiles/ocw/node-hugo
--- a/pipelines/ocw/docker-image-pipeline.yml
+++ b/pipelines/ocw/docker-image-pipeline.yml
@@ -12,7 +12,7 @@ resources:
       check_every: 1h
       source:
           repository: node
-          tag: '14'
+          tag: 14-buster-slim
     - name: ocw-course-publisher-image
       type: docker-image
       source:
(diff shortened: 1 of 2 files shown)


pipelines/ocw/docker-image-pipeline.yml
```diff
@@ -0,0 +1,33 @@
...
+
+
+jobs:
+    - name: publish
+      public: true
+      serial: true
+      plan:
+          - get: node-image
+            trigger: true
+          - get: ol-infrastructure
```
blarghmatey, line 30: Should probably add a trigger for this as well with a path glob to match the Dockerfile that it's relying on.

markbreedlove:
Also, I don't know how far we want to go with this, but the Node version has to be maintained in two separate files: `dockerfiles/ocw/node-hugo/Dockerfile` and the pipeline in this PR. I'm not sure what the best way would be to keep them in sync, or to parameterize the node version.

pipelines/ocw/docker-image-pipeline.yml
markbreedlove: 👍

pipelines/ocw/docker-image-pipeline.yml
```diff
@@ -0,0 +1,33 @@
...
+      type: git
+      source:
+          uri: https://github.com/mitodl/ol-infrastructure.git
+          branch: ((ol-infrastructure-git-ref))
+    - name: node-image
+      type: docker-image
+      check_every: 1h
+      source:
+          repository: node
+          tag: '14'
```
blarghmatey, line 13: To cut down on the amount of data being pulled it might make sense to base our image build off of `14-buster-slim`

markbreedlove: Wow, it's a lot smaller:

```
(venv) ol-infrastructure$ docker image ls
REPOSITORY                    TAG                 IMAGE ID            CREATED             SIZE
[ ... ]
node                          14                  9153ee3e2ced        8 days ago          943MB
node                          14-buster-slim      d4c353c8e216        8 days ago          181MB
```

pipelines/ocw/docker-image-pipeline.yml
markbreedlove: adf00b6d should have changes to the Dockerfile trigger builds.

blarghmatey:
:+1: looks good. And yes, the dockerhub credentials will automatically get pulled from Vault.

blarghmatey:
As part of this it's probably worth updating the dockerfile to pull from the `14-buster-slim` image as the source.

pipelines/ocw/docker-image-pipeline.yml
markbreedlove: dd2b716b changes the base image.

PR #177, opened 2021-05-25
Add pipeline to build our own concourse-rclone-resource

markbreedlove:
Build our own Docker image for concourse-rclone-resource so that we can use our own fork of the project, which removes a hardcoded argument to 'rclone' that we do not want.

### DIFF
--- a/pipelines/concourse-rclone-docker-image/docker-image-pipeline.yml
+++ b/pipelines/concourse-rclone-docker-image/docker-image-pipeline.yml
@@ -1,0 +1,26 @@
+---
+resources:
+  - name: concourse-rclone-resource
+    type: git
+    source:
+      uri: https://github.com/mitodl/concourse-rclone-resource.git
+      branch: mitodl
+  - name: concourse-rclone-resource-image
+    type: docker-image
+    source:
+      username: ((dockerhub.username))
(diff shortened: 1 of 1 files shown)


pipelines/concourse-rclone-docker-image/docker-image-pipeline.yml
```diff
@@ -0,0 +1,26 @@
+---
+resources:
+  - name: concourse-rclone-resource
+    type: git
+    source:
+      uri: https://github.com/mitodl/concourse-rclone-resource.git
+      branch: mitodl
```
blarghmatey, line 7: Why not stick with main or master?

markbreedlove: Just because I thought it would be easier to make a PR to change the upstream this way, or to handle incoming pulls from upstream without our commits being on `master`. But maybe that's not such a big deal.

blarghmatey: Not a big deal either way, let's merge as is so as not to block on this. We can always change things later.

blarghmatey:
:+1:

PR #179, opened 2021-05-26
Added mitx-mfe resources

shaidar:
This is worked related to getting MITx app-learning MFE deployed.
- Added an app-learning MFE build pipeline
- Added S3 and cloudfront config for the MFE
- Fixed issue with policy in AWSBase
- Added mitx DNS Zone

### DIFF
--- a/pipelines/mitx/mfe-frontend-app-learning.yml
+++ b/pipelines/mitx/mfe-frontend-app-learning.yml
@@ -1,0 +1,65 @@
+---
+resource_types:
+    - name: rclone
+      type: docker-image
+      source:
+          repository: sothr/concourse-rclone-resource
+          tag: latest
+
+resources:
+  - name: mfe-app-learning
+    type: git
+    source:
+      uri: https://github.com/edx/frontend-app-learning.git
+      branch: master
+  - name: mfe-app-learning-bucket
+    type: rclone
+    source:
+      config: |
+        [s3-remote]
+        type = s3
+        provider = AWS
+        env_auth = true
+        region = us-east-1
+
+jobs:
+  - name: build
+    plan:
+      - get: mfe-app-learning
+        trigger: true
+      - task: npm-install-and-build
+        config:
+          platform: linux
+          image_resource:
+            type: registry-image
+            source:
+              repository: node
+          inputs:
+            - name: mfe-app-learning
+          outputs:
+            - name: mfe-app-learning/dist
+          run:
+            path: sh
+            args:
+            - -exc
+            - |
+              echo ACCESS_TOKEN_COOKIE_NAME='edx-jwt-cookie-header-payload' > ./mfe-app-learning/.env
+              echo BASE_URL='https://mitx-qa-next.mitx.mit.edu' >> ./mfe-app-learning/.env
+              echo CSRF_TOKEN_API_PATH='/csrf/api/v1/token' >> ./mfe-app-learning/.env
+              echo LMS_BASE_URL='https://mitx-qa-next.mitx.mit.edu' >> ./mfe-app-learning/.env
+              echo LOGIN_URL='https://mitx-qa-next.mitx.mit.edu/login' >> ./mfe-app-learning/.env
+              echo LOGOUT_URL='https://mitx-qa-next.mitx.mit.edu/logout' >> ./mfe-app-learning/.env
+              echo MARKETING_SITE_BASE_URL='https://mitx-qa-next.mitx.mit.edu' >> ./mfe-app-learning/.env
+              echo REFRESH_ACCESS_TOKEN_ENDPOINT='https://mitx-qa-next.mitx.mit.edu/login_refresh' >> ./mfe-app-learning/.env
+              echo SEARCH_CATALOG_URL='https://mitx-qa-next.mitx.mit.edu/courses' >> ./mfe-app-learning/.env
+              echo SITE_NAME='MITx QA' >> ./mfe-app-learning/.env
+              echo STUDIO_BASE_URL='https://studio-mitx-qa-next.mitx.mit.edu' >> ./mfe-app-learning/.env
+              echo USER_INFO_COOKIE_NAME='edx-user-info' >> ./mfe-app-learning/.env
+              cd ./mfe-app-learning
+              npm install
+              NODE_ENV=production npm run build
+      - put: mfe-app-learning-bucket
+        params:
+          source: mfe-app-learning/dist
+          destination:
+            - dir: "s3-remote:ol-eng-artifacts/mfe"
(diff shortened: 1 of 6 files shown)


src/ol_infrastructure/applications/mitx/__main__.py
```diff
@@ -0,0 +1,45 @@
...
+    name=mfe_app_learning_config.domains[0],
+    type="CNAME",
+    ttl=fifteen_minutes,
+    records=[mfe_app_learning.cloudfront_distribution.domain_name],
+    zone_id=mitx_zone_id,
+)
+
+export("mfe_app_learning_bucket", mfe_app_learning.site_bucket.bucket)
+export("xpro_certs_distribution_id", mfe_app_learning.cloudfront_distribution.id)
+export("xpro_certs_acm_cname", mfe_app_learning.site_tls.domain_name)
```
blarghmatey, line 45: Copy :spaghetti: on the export key naming

src/ol_infrastructure/applications/mitx/__main__.py
```diff
@@ -0,0 +1,45 @@
...
+        "Environment": mitx_environment,
+        "Application": Apps.mitx_edx.value,
+    }
+)
+
+mitx_zone_id = dns_stack.get_output("mitx_zone_id")
+mfe_app_learning_config = S3ServerlessSiteConfig(
+    site_name=f"mfe-app-learning-{mitx_environment}",
+    domains=[f"mfe-app-learning-{mitx_environment}.mitx.mit.edu"],
+    bucket_name=f"mitx-mfe-app-learning-{mitx_environment}",
```
blarghmatey, line 28: This shouldn't be specific to the learning MFE, we should have one bucket and distribution to accommodate all of the MFEs that we deploy

src/ol_infrastructure/applications/mitx/__main__.py
```diff
@@ -0,0 +1,45 @@
...
+        "OU": "mitx",
+        "Environment": mitx_environment,
+        "Application": Apps.mitx_edx.value,
+    }
+)
+
+mitx_zone_id = dns_stack.get_output("mitx_zone_id")
+mfe_app_learning_config = S3ServerlessSiteConfig(
+    site_name=f"mfe-app-learning-{mitx_environment}",
+    domains=[f"mfe-app-learning-{mitx_environment}.mitx.mit.edu"],
```
blarghmatey, line 27: For the domain we should probably set it to something like `frontends.mitx.mit.edu` or `static.` with appropriate qualifiers to match our other "purposes".

pipelines/mitx/mfe-frontend-app-learning.yml
```diff
@@ -0,0 +1,65 @@
...
+              echo CSRF_TOKEN_API_PATH='/csrf/api/v1/token' >> ./mfe-app-learning/.env
+              echo LMS_BASE_URL='https://mitx-qa-next.mitx.mit.edu' >> ./mfe-app-learning/.env
+              echo LOGIN_URL='https://mitx-qa-next.mitx.mit.edu/login' >> ./mfe-app-learning/.env
+              echo LOGOUT_URL='https://mitx-qa-next.mitx.mit.edu/logout' >> ./mfe-app-learning/.env
+              echo MARKETING_SITE_BASE_URL='https://mitx-qa-next.mitx.mit.edu' >> ./mfe-app-learning/.env
+              echo REFRESH_ACCESS_TOKEN_ENDPOINT='https://mitx-qa-next.mitx.mit.edu/login_refresh' >> ./mfe-app-learning/.env
+              echo SEARCH_CATALOG_URL='https://mitx-qa-next.mitx.mit.edu/courses' >> ./mfe-app-learning/.env
+              echo SITE_NAME='MITx QA' >> ./mfe-app-learning/.env
+              echo STUDIO_BASE_URL='https://studio-mitx-qa-next.mitx.mit.edu' >> ./mfe-app-learning/.env
+              echo USER_INFO_COOKIE_NAME='edx-user-info' >> ./mfe-app-learning/.env
```
blarghmatey, line 57: It's probably worth revisiting how we manage the environment variables during the pipeline execution now that we've figured out the mechanical aspects.

src/ol_infrastructure/applications/mitx/__main__.py
shaidar: But I was under the impression that each MFE has its own domain name, no?

blarghmatey: It doesn't have to and I'd prefer to not have to manage a dozen domains. It's entirely possible to have them all under the same domain and addressed via path. For example, the learning experience MFE could be located at `static.mitx.mit.edu/learn`

src/ol_infrastructure/applications/mitx/__main__.py
```diff
@@ -0,0 +1,45 @@
...
+    name=mfe_app_learning_config.domains[0],
+    type="CNAME",
+    ttl=fifteen_minutes,
+    records=[mfe_app_learning.cloudfront_distribution.domain_name],
+    zone_id=mitx_zone_id,
+)
+
+export("mitx_mfe_app_learning_bucket", mfe_app_learning.site_bucket.bucket)
+export("mfe_app_learning_distribution_id", mfe_app_learning.cloudfront_distribution.id)
+export("mfe_app_learning_acm_cname", mfe_app_learning.site_tls.domain_name)
```
blarghmatey, line 45: These should all be for generic MFEs, non-specific to `app_learning`

src/ol_infrastructure/applications/mitx/__main__.py
```diff
@@ -0,0 +1,45 @@
...
+aws_config = AWSBase(
+    tags={
+        "OU": "mitx",
+        "Environment": mitx_environment,
+        "Application": Apps.mitx_edx.value,
+    }
+)
+
+mitx_zone_id = dns_stack.get_output("mitx_zone_id")
+mfe_app_learning_config = S3ServerlessSiteConfig(
```
blarghmatey, line 25: I think we are going to have to parameterize this from the stack settings to provide for our different "purposes", since I don't think that we can share the compiled MFEs between draft and live for example.

shaidar: Do we even need MFE's to be running in draft?

blarghmatey: I'd defer to @pdpinch for that one, but we'll want to at least be able to support next vs current as we make that transition (maybe?).

shaidar: Ideally, I'd like to keep things as simple as possible at least until we can get this to work first and then figure out how to handle all the special cases we have, but yeah I'll defer to the two of you

blarghmatey: Let's just get it running as is and figure it out, like you said.

PR #185, opened 2021-05-28
Minor fixes related to MFE work

shaidar:


src/ol_infrastructure/infrastructure/aws/dns/__main__.py
```diff
@@ -4,7 +4,7 @@
 from ol_infrastructure.lib.aws.route53_helper import zone_opts
 from ol_infrastructure.lib.ol_types import AWSBase
 
-mitxpro_legacy_dns_name = "mitxpro.mit.edu"
+mitxpro_legacy_dns_name = "mitxpro.mit.edu."
```
blarghmatey, line 7: We actually want to keep the trailing `.` off now that we've fixed the stack state.

blarghmatey:
:+1:

PR #192, opened 2021-06-04
Adding GitHub webhook and shared secret for OCW Studio

blarghmatey:
OCW Studio needs to have a webhook registered to fire from ocw-hugo-projects when it is updated to allow the app to update the associated records in the DB


shaidar:
👍

mbertrand (approved):
👍

PR #197, opened 2021-06-08
Pipeline for publishing edx-sysadmin

annagav:
A pipeline that:
- gets triggered every time new commit is added to release branch on edx-sysadmin.
- builds the package
- uploads the new version

### DIFF
--- a/pipelines/edx-sysadmin-publish.yml
+++ b/pipelines/edx-sysadmin-publish.yml
@@ -28,8 +28,7 @@ jobs:
     plan:
     - get: edx-sysadmin
       trigger: true
-
-    - task: build-and-upload
+    - task: build
       config:
         platform: linux
         image_resource:
@@ -48,7 +47,7 @@ jobs:
             - python setup.py bdist_wheel sdist
     - put: edx-sysadmin-pypi
       params:
-        glob: 'edx-sysadmin/dist'
+        glob: 'edx-sysadmin/dist/*'
       get_params:
         count_retries: 10
         delay_between_retries: 30
--- a/pipelines/edx-sysadmin-publish.yml
+++ b/pipelines/edx-sysadmin-publish.yml
@@ -1,0 +1,56 @@
+---
+resource_types:
+- name: pypi
+  type: docker-image
+  source:
+    repository: cfplatformeng/concourse-pypi-resource
+    tag: latest
+
+resources:
+    - name: edx-sysadmin-repo
+      type: git
+      source:
+          uri: https://github.com/mitodl/edx-sysadmin.git
+          branch: ((release))
+
+    - name: edx-sysadmin-pypi
+      type: pypi
+      source:
+        name: edx-sysadmin
+        packaging: any
+        repository:
+          username: ((username))
+          password: ((password))
+
+
+jobs:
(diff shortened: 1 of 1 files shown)


pipelines/edx-sysadmin-publish.yml
```diff
@@ -0,0 +1,56 @@
...
+  source:
+    repository: cfplatformeng/concourse-pypi-resource
+    tag: latest
+
+resources:
+    - name: edx-sysadmin-repo
+      type: git
+      source:
+          uri: https://github.com/mitodl/edx-sysadmin.git
+          branch: ((release))
```
blarghmatey, line 14: This doesn't need to be a var if we're only going to publish from that branch. You can remove the `(( ))`

pipelines/edx-sysadmin-publish.yml
```diff
@@ -0,0 +1,56 @@
...
+  - get: edx-sysadmin
+    trigger: true
+
+  - task: build-and-upload
+      config:
+        platform: linux
+        image_resource:
+          type: registry-image
+          source:
+            repository: node
```
blarghmatey, line 39: Is the use of the node image due to a requirement for compiling JS/CSS during the package building?

pipelines/edx-sysadmin-publish.yml
```diff
@@ -0,0 +1,56 @@
...
+          uri: https://github.com/mitodl/edx-sysadmin.git
+          branch: ((release))
+
+    - name: edx-sysadmin-pypi
+      type: pypi
+      source:
+        name: edx-sysadmin
+        packaging: any
+        repository:
+          username: ((username))
```
blarghmatey, line 22: We should change these vars to be something like `pypi_creds.username` and `pypi_creds.password` so that we can put them under a single key in Vault that Concourse can pull dynamically.

blarghmatey:
:+1:

pipelines/edx-sysadmin-publish.yml
```diff
@@ -1,55 +1,56 @@
...
         run:
-          path: python
+          path: sh
           args:
-            - setup.py
-            - bdist_wheel
-            - sdist
+            - -exc
+            - |
+              cd edx-sysadmin
```
blarghmatey, line 47: There's a parameter for the run command that sets the directory that it executes in so that you don't have to run `cd` as part of the task. You can just add `dir: edx-sysadmin-repo` to the `run` block.

annagav:
@blarghmatey I think this should be all set now. Could you take another look at the changes I made?

blarghmatey:
:+1:

annagav:
@blarghmatey when the release was merged today for edx-sysadmin my pipeline that I was testing on cicd-qa.odl.mit.edu got triggered and uploaded that new release to PyPI. I am going to disable it now, so that you get set it up in production. Do you know if concourse waits for tests to pass after the merge of the commit before triggering the job?

blarghmatey:
As it is right now it just watches for a new version in the repo which equates to a different commit hash. We should be able to add https://github.com/dpb587/github-status-resource as an input to make it trigger based on status.

blarghmatey:
The other option is to put Concourse in charge of running the tests.

Issue #196, opened 2021-06-07
OCW Studio Publishing Workflow


blarghmatey:
There is an open PR to create a pipeline for a single course build [here](https://github.com/mitodl/ol-infrastructure/pull/180). There are still a few changes to be made to be fully functional as a pipeline template.

Updates needed for the pipeline in that PR to be functional as a template
- [ ] Template course content branch
- [ ] Remove duplicate publish to draft/live buckets
- [ ] Reduce the number of variables needed to get it set up
- [ ] Wrapper script for "installing" the pipeline on concourse using Fly CLI

In addition to having a pipeline template for an individual course we also need a meta pipeline for creating pipeline instances based on the template.
- [ ] Template based on course id
- [ ] Template based on course content branch
- [ ] Use git-repo resource to enumerate the courses in the GitHub organization

The current blocker to an individual course running successfully is the missing `course.json` file in the GitHub content that is published from OCW Studio
- [ ] course.json to be published with course content to Github

One shortcoming of the current overall setup with how the courses are built is that any time the hugo-theme or ocw-www repositories are modified it will trigger a build for _every_ course repository. 
- [ ] Experiment with removing content hashing for course themes to avoid thundering herd

blarghmatey:
This looks like a useful resource to consider for the meta-pipeline to cut down on polling and latency for course publishes https://github.com/homedepot/github-webhook-resource

blarghmatey:
Here's the resource that lists the repositories under a github organization which we can use for the meta-pipeline to enumerate the course repositories https://github.com/coralogix/eng-concourse-resource-github-list-repos

pdpinch:
@mbertrand do you think we can close this yet?

mbertrand:
The final checkbox is something I haven't looked at yet, but could separate that into it's own issue.  The unchecked ones before that are no longer relevant.

pdpinch:
> One shortcoming of the current overall setup with how the courses are built is that any time the hugo-theme or ocw-www repositories are modified it will trigger a build for every course repository.

> - [ ] Experiment with removing content hashing for course themes to avoid thundering herd

Have we seen this in practice yet?

mbertrand:
Not that I know of, but only about 21 sites have draft & live pipelines at this point on RC.  Could be a problem when we have 100s or 1000s of pipelines trying to run at once.

PR #204, opened 2021-06-10
Pipeline for ocw-www homepage

mbertrand:
Related to #196 

Pipeline for previewing/publishing the ocw-www homepage.
Can be tested on the concourse QA site at ../teams/ocw/pipelines/ocw-www-preview-pipeline/jobs/build-ocw-www/

Also added fly bash scripts for creating the pipelines.  To try them, you'll have to set the following env variables:

```
export FLY_TARGET=<your_concourse_target>
export FLY_TEAM=ocw
export ARTIFACT_BUCKET=ol-eng-artifacts
export GIT_CONFIG_ORG=ocw-content-rc
export GITHUB_DOMAIN=<MIT github domain>
export HUGO_THEME_BRANCH=release
export OCW_WWW_REPO=the-ocw-www-rc_2b7b49fc4f784ca293edee4368e35a88
export OCW_CONTENT_BUCKET_PREVIEW=ocw-content-draft-qa
export OCW_CONTENT_BUCKET_RELEASE=<TBD>
export OCW_STUDIO_BASE_URL=<RC or prod base url of ocw-studio>
export FASTLY_SERVICE_ID_PREVIEW=<secret>
export FASTLY_API_TOKEN_PREVIEW=<secret>
export FASTLY_SERVICE_ID_RELEASE=<secret>
export FASTLY_API_TOKEN_RELEASE=<secret>
export GIT_PRIVATE_KEY=<secret>
```

Then you can run:
```
sh pipeline-theme-assets.sh
```

and 
```
sh pipeline-ocw-www.sh preview
```
or
```
sh pipeline-ocw-www.sh release
```

### DIFF
--- a/pipelines/ocw/pipeline-ocw-www.yml
+++ b/pipelines/ocw/pipeline-ocw-www.yml
@@ -14,8 +14,8 @@ resources:
   - name: course-markdown
     type: git
     source:
-      uri: git@((git-domain)):((github-org))/((course-repo)).git
-      branch: ((course-repo-branch))
+      uri: git@((git-domain)):((github-org))/((ocw-www-repo)).git
+      branch: ((ocw-www-repo-branch))
       private_key: ((git-private-key))
   - name: ocw-hugo-projects
     type: git
@@ -64,7 +64,7 @@ jobs:
               tar -xvzf ocw-hugo-themes/ocw-hugo-themes-*.tgz -C ocw-hugo-themes/theme
               cd course-markdown
               export OCW_STUDIO_BASE_URL=((ocw-studio-url))
-              hugo --config ../ocw-hugo-projects/((site-config))/config.yaml --themesDir ../ocw-hugo-themes/theme
+              hugo --config ../ocw-hugo-projects/ocw-www/config.yaml --themesDir ../ocw-hugo-themes/theme
       - put: ocw-site
         params:
           source: course-markdown/public
--- a/pipelines/ocw/pipeline-ocw-www.yml
+++ b/pipelines/ocw/pipeline-ocw-www.yml
@@ -1,0 +1,102 @@
+---
+resource_types:
+  - name: rclone
+    type: docker-image
+    source:
+      repository: mitodl/concourse-rclone-resource
+      tag: latest
+resources:
+  - name: ocw-hugo-themes
+    type: s3
+    source:
+      bucket: ((artifact-bucket))
+      regexp: ocw-hugo-themes/((hugo-theme-branch))/ocw-hugo-themes-(.*).tgz
+  - name: course-markdown
+    type: git
+    source:
+      uri: git@((git-domain)):((github-org))/((course-repo)).git
+      branch: ((course-repo-branch))
+      private_key: ((git-private-key))
+  - name: ocw-hugo-projects
+    type: git
+    source:
+      uri: https://github.com/mitodl/ocw-hugo-projects.git
+      branch: main
+  - name: ocw-site
+    type: rclone
+    source:
+      config: |
+        [s3-remote]
+        type = s3
+        provider = AWS
+        env_auth = true
+        region = us-east-1
+
+jobs:
+  - name: build-ocw-www
+    serial: true
+    plan:
+      - get: ocw-hugo-themes
+        trigger: true
+      - get: ocw-hugo-projects
+        trigger: true
+      - get: course-markdown
+        trigger: true
+      - task: build-ocw-www-task
+        config:
+          platform: linux
+          image_resource:
+            type: docker-image
+            source: {repository: mitodl/ocw-course-publisher, tag: latest}
+          inputs:
+            - name: ocw-hugo-themes
+            - name: course-markdown
+            - name: ocw-hugo-projects
+          outputs:
+            - name: course-markdown
+            - name: ocw-hugo-themes
+          run:
+            path: sh
+            args:
+            - -exc
+            - |
+              mkdir ocw-hugo-themes/theme
+              tar -xvzf ocw-hugo-themes/ocw-hugo-themes-*.tgz -C ocw-hugo-themes/theme
+              cd course-markdown
+              export OCW_STUDIO_BASE_URL=((ocw-studio-url))
+              hugo --config ../ocw-hugo-projects/((site-config))/config.yaml --themesDir ../ocw-hugo-themes/theme
+      - put: ocw-site
+        params:
+          source: course-markdown/public
+          destination:
+            - dir: s3-remote:((ocw-bucket))
+              command: copyto
+              args:
+                - --ignore-size
+                - --checksum
+      - put: ocw-site
+        params:
+          source: ocw-hugo-themes/theme/base-theme/dist
+          destination:
+            - dir: s3-remote:((ocw-bucket))
+              command: copyto
+              args:
+                - --ignore-size
+                - --checksum
+      - task: clear-cdn-cache
+        config:
+          platform: linux
+          image_resource:
+            type: docker-image
+            source: {repository: curlimages/curl}
+          run:
+            path: curl
+            args:
+              - -f
+              - -X
+              - POST
+              - -H
+              - 'Fastly-Key: ((fastly-api-token))'
+              - -H
+              - 'Fastly-Soft-Purge: 1'
+              - https://api.fastly.com/service/((fastly-service-id))/purge_all
--- a/pipelines/ocw/pipeline-ocw-www.yml
+++ b/pipelines/ocw/pipeline-ocw-www.yml
@@ -21,7 +21,7 @@ resources:
     type: git
     source:
       uri: https://github.com/mitodl/ocw-hugo-projects.git
-      branch: main
+      branch: ((ocw-hugo-projects-branch))
   - name: ocw-site
     type: rclone
     source:
--- a/pipelines/ocw/pipeline-ocw-www.yml
+++ b/pipelines/ocw/pipeline-ocw-www.yml
@@ -10,7 +10,7 @@ resources:
     type: s3
     source:
       bucket: ol-eng-artifacts
-      regexp: ocw-hugo-themes/((hugo-theme-branch))/ocw-hugo-themes-(.*).tgz
+      regexp: ocw-hugo-themes/release/ocw-hugo-themes-(.*).tgz
   - name: course-markdown
     type: git
     source:
--- a/pipelines/ocw/pipeline-ocw-www.yml
+++ b/pipelines/ocw/pipeline-ocw-www.yml
@@ -5,12 +5,23 @@ resource_types:
     source:
       repository: mitodl/concourse-rclone-resource
       tag: latest
+  - name: simple-s3
+    type: docker-image
+    source:
+      repository: 18fgsa/s3-resource-simple
 resources:
   - name: ocw-hugo-themes
     type: s3
     source:
       bucket: ol-eng-artifacts
       regexp: ocw-hugo-themes/release/ocw-hugo-themes-(.*).tgz
+  - name: ocw-studio-s3
+    type: simple-s3
+    source:
+      bucket: ((ocw-studio-bucket))
+      options:
+        - "--exclude '*'"
+        - "--include '((ocw-www-site))/*'"
   - name: course-markdown
     type: git
     source:
@@ -42,6 +53,8 @@ jobs:
         trigger: true
       - get: course-markdown
         trigger: true
+      - get: ocw-studio-s3
+        trigger: false
       - task: build-ocw-www-task
         params:
           OCW_STUDIO_BASE_URL: ((ocw-studio-url))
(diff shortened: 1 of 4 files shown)


pipelines/ocw/scripts/pipeline-ocw-www.sh
```diff
@@ -0,0 +1,33 @@
...
+else
+  echo "Invalid branch $1"
+  exit 1
+fi
+
+yes | fly -t $FLY_TARGET set-pipeline \
+-p ocw-www-pipeline-$SITE_BRANCH \
+--team=$FLY_TEAM \
+--config=pipelines/ocw/pipeline-ocw-www.yml \
+-v artifact-bucket=$ARTIFACT_BUCKET \
```
blarghmatey, line 23: The artifact bucket should always be the same, so no need to template that, unless you're including the sub-path here?

pipelines/ocw/scripts/pipeline-ocw-www.sh
```diff
@@ -0,0 +1,33 @@
...
+-v git-domain=$GITHUB_DOMAIN \
+-v github-org=$GIT_CONFIG_ORG \
+-v hugo-theme-branch=$HUGO_THEME_BRANCH \
+-v ocw-www-repo=$OCW_WWW_REPO \
+-v ocw-www-repo-branch=$SITE_BRANCH \
+-v ocw-bucket=$OCW_CONTENT_BUCKET \
+-v ocw-studio-url=$OCW_STUDIO_BASE_URL \
+-v fastly-service-id=$FASTLY_SERVICE_ID \
+-v fastly-api-token=$FASTLY_API_TOKEN \
+-v git-private-key="$GIT_PRIVATE_KEY"
```
blarghmatey, line 33: This is only necessary for RC because of it being a private repository

pipelines/ocw/scripts/pipeline-ocw-www.sh
```diff
@@ -0,0 +1,33 @@
...
+-v artifact-bucket=$ARTIFACT_BUCKET \
+-v git-domain=$GITHUB_DOMAIN \
+-v github-org=$GIT_CONFIG_ORG \
+-v hugo-theme-branch=$HUGO_THEME_BRANCH \
+-v ocw-www-repo=$OCW_WWW_REPO \
+-v ocw-www-repo-branch=$SITE_BRANCH \
+-v ocw-bucket=$OCW_CONTENT_BUCKET \
+-v ocw-studio-url=$OCW_STUDIO_BASE_URL \
+-v fastly-service-id=$FASTLY_SERVICE_ID \
+-v fastly-api-token=$FASTLY_API_TOKEN \
```
blarghmatey, line 32: This should be getting pulled from Vault at runtime, so we shouldn't be setting it via the CLI

pipelines/ocw/scripts/pipeline-theme-assets.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+yes | fly -t $FLY_TARGET set-pipeline \
+-p ocw-theme-assets-$HUGO_THEME_BRANCH \
+--team=$FLY_TEAM \
```
blarghmatey, line 5: These should all be set under the same team of ocw

pipelines/ocw/scripts/pipeline-theme-assets.sh
```diff
@@ -0,0 +1,8 @@
...
+--config=pipelines/ocw/pipeline-theme-assets.yml \
+-v hugo-theme-branch=$HUGO_THEME_BRANCH \
+-v artifact-bucket=$ARTIFACT_BUCKET
```
blarghmatey, line 8: The artifact bucket should always be the same.

pipelines/ocw/pipeline-ocw-www.yml
```diff
@@ -0,0 +1,102 @@
...
+            - name: ocw-hugo-themes
+          run:
+            path: sh
+            args:
+            - -exc
+            - |
+              mkdir ocw-hugo-themes/theme
+              tar -xvzf ocw-hugo-themes/ocw-hugo-themes-*.tgz -C ocw-hugo-themes/theme
+              cd course-markdown
+              export OCW_STUDIO_BASE_URL=((ocw-studio-url))
```
blarghmatey, line 66: We should probably extract this as a param in the run args to handle setting environment variables.

pipelines/ocw/pipeline-ocw-www.yml
```diff
@@ -0,0 +1,102 @@
...
+            - name: course-markdown
+            - name: ocw-hugo-themes
+          run:
+            path: sh
+            args:
+            - -exc
+            - |
+              mkdir ocw-hugo-themes/theme
+              tar -xvzf ocw-hugo-themes/ocw-hugo-themes-*.tgz -C ocw-hugo-themes/theme
+              cd course-markdown
```
blarghmatey, line 65: There's a `dir` parameter to the `run` block that controls which directory the commands are executed in, so we can use that and remove the `cd` command.

pipelines/ocw/pipeline-ocw-www.yml
```diff
@@ -0,0 +1,102 @@
...
+            path: curl
+            args:
+              - -f
+              - -X
+              - POST
+              - -H
+              - 'Fastly-Key: ((fastly-api-token))'
+              - -H
+              - 'Fastly-Soft-Purge: 1'
+              - https://api.fastly.com/service/((fastly-service-id))/purge_all
```
blarghmatey, line 102: This is ok for now, but we should look at setting up cache keys to only purge the content that is being uploaded, rather than the whole cache.

blarghmatey (changes requested):
Given the number of variables, it would probably be best to create a few files that have all of the correct values for a given destination. For instance, one file that has everything needed to publish to draft/preview for RC, another for publishing to live/published for RC, etc.

pipelines/ocw/scripts/pipeline-ocw-www.sh
mbertrand: I was assuming all these would already be set as environment variables, via vault or some other process.  How would I specify that these values come from vault directly instead of an env. variable?

pipelines/ocw/pipeline-ocw-www.yml
mbertrand: Mark's course pipeline is set up that way, but since ocw-www is writing to the root folder I wasn't sure if that was possible here.  Maybe iterate through all the folders in the course repo, requesting a purge for each, plus the static folder?

pipelines/ocw/scripts/pipeline-ocw-www.sh
mbertrand: Can we have a private key for a public repo anyway?  Otherwise, it seems I will need separate pipelines for RC vs production, with the only difference (besides passed parameters) is the omission of `private_key: ((git-private-key))` from the production resource.  Passing a blank git private key doesn't work with a public repo.  

```
    source:
      uri: git@((git-domain)):((github-org))/((ocw-www-repo)).git
      branch: ((ocw-www-repo-branch))
      private_key: ((git-private-key))
```

pipelines/ocw/scripts/pipeline-ocw-www.sh
mbertrand: If all these vars are defined in vault (`ocw-bucket`, `fastly-service-id`, etc), will concourse automatically find them assuming `CONCOURSE_VAULT_URL` is set?

mbertrand:
@blarghmatey I just committed changes to address most of your comments, have some questions about other ones though.

pipelines/ocw/scripts/pipeline-ocw-www.sh
mbertrand: Related comment:
> it would probably be best to create a few files that have all of the correct values for a given destination

I could use some guidance on how to best handle this.  Some of these values should not be public (fastly api token, github key), so should I create a file for values that are okay to be public but the rest should go in vault?  And if they are in vault, does anything need to be modified in the pipeline generation script to get the values from there?

blarghmatey: Sorry, thought I had already responded to this. So, any unset var in the pipeline will automatically attempt to look up a key by that name in the configured credential provider (in this case Vault). More details of how that works are available at https://concourse-ci.org/vault-credential-manager.html

So, in order to handle those vars we just need to set the values in Vault. For values from Vault that have multiple values (e.g. Fastly API token and service Id) I like to keep those as a single key in Vault so that the var would look like `fastly.api_token` and `fastly.service_id`. The different values for preview and published Fastly distributions would be located based on the name of the pipeline group, so there's no need to special case the var names in the pipeline.

mbertrand: > on the name of the pipeline group

There aren't any pipeline groups specified yet, so I'll work on adding that too.

pipelines/ocw/scripts/pipeline-ocw-www.sh
```diff
@@ -0,0 +1,45 @@
...
+# OCW_WWW_REPO=<the github repo for the ocw-www site>
+# OCW_CONTENT_BUCKET_PREVIEW=<the S3 preview bucket>
+# OCW_CONTENT_BUCKET_RELEASE=<the S3 release bucket>
+# OCW_STUDIO_BASE_URL=<the ocw-studio base url>
+
+# The following need to be set via vault, per environment and pipeline group (preview, release):
+# fastly-service-id
+# fastly-api-token
+# git-private-key
+
```
mbertrand, line 17: @blarghmatey I incorporated groups into the scripts: there's now a `ocw-theme-assets` group with `release` and `release-candidate` branches, plus a `ocw-www-pipeline-via-vault` group with `preview` and `release` branches.

I removed the assignment of fastly & github key values from the ocw-www pipeline script so that it will try to pick them up from vault.  But I assume they are not there yet because those new pipelines are failing.

There are other values that will be different between RC and production, as documented above.  The values for  some of them  are yet to be determined for production.  So I think environment values would still be best for those?

Also, I'm still not sure how best to handle authentication, I'm unclear where these scripts are intended to be run from. For now I'm assuming authentication has already been taken care of, with fly target `odl-concourse`.

pipelines/ocw/pipeline-ocw-www.yml
```diff
@@ -79,6 +66,18 @@ jobs:
...
+          platform: linux
+          image_resource:
+            type: docker-image
+            source: {repository: amazon/aws-cli, tag: latest}
+          run:
+            path: sh
+            args:
+            - -exc
+            - |
+              aws s3 sync s3://((ocw-studio-bucket)) s3://((ocw-bucket)) --exclude '*' --include '((ocw-www-site))/*'
```
blarghmatey, line 80: I don't think we want to just copy over everything in the ocw studio bucket, as this could end up publishing an asset that isn't ready to be released yet. We should try for a more targeted sync command.

pipelines/ocw/pipeline-ocw-www.yml
```diff
@@ -5,23 +5,12 @@ resource_types:
     source:
       repository: mitodl/concourse-rclone-resource
       tag: latest
-  - name: simple-s3
```
blarghmatey: You still need this resource as it's being used for publishing the compiled site and assets to s3

pipelines/ocw/pipeline-ocw-www.yml
mbertrand: That's what the `--include '((ocw-www-site))/*'` is doing: only files specific to this site will be copied over.

blarghmatey (approved):

pipelines/ocw/pipeline-ocw-www.yml
mbertrand: The resource is longer being used for anything, it's been completely replaced by the aws cli docker image task

blarghmatey:
:+1:

pipelines/ocw/scripts/pipeline-ocw-www.sh
```diff
@@ -26,19 +26,20 @@ do
...
     echo "Invalid branch $1"
     exit 1
   fi
 
   yes | fly -t odl-concourse set-pipeline \
-  -p ocw-www-pipeline-via-vault \
+  -p ocw-www-pipeline-$VERSION \
   --team=ocw \
   --config=pipelines/ocw/pipeline-ocw-www.yml \
-  --instance-var branch=$branch \
```
blarghmatey: The branch still needs to be set as a var, it just won't be an instance var.

mbertrand: What will this var be used for?

mbertrand: `((fastly-service-id))-((branch))`?

blarghmatey:
:+1:

PR #251, opened 2021-07-12
Adding ASG creation for edxapp deployments

blarghmatey:
- Updated image building template to properly populate TLS certificates
- Added ASG creation for web and worker nodes
- Added automatic creation of ACM certificate for configured domains

### DIFF
--- a/src/ol_infrastructure/applications/edxapp/__main__.py
+++ b/src/ol_infrastructure/applications/edxapp/__main__.py
@@ -1,14 +1,20 @@
 # TODO: Manage database object creation
+import base64
 import json
 from pathlib import Path
 from string import Template
 
 import pulumi_vault as vault
+import yaml
 from pulumi import Config, StackReference, export
-from pulumi_aws import ec2, get_caller_identity, iam, s3
+from pulumi_aws import acm, autoscaling, ec2, get_caller_identity, iam, lb, route53, s3
 from pulumi_consul import Node, Service, ServiceCheckArgs
 
-from bridge.lib.magic_numbers import DEFAULT_MYSQL_PORT, DEFAULT_REDIS_PORT
+from bridge.lib.magic_numbers import (
+    DEFAULT_HTTPS_PORT,
+    DEFAULT_MYSQL_PORT,
+    DEFAULT_REDIS_PORT,
+)
 from ol_infrastructure.components.aws.cache import OLAmazonCache, OLAmazonRedisConfig
 from ol_infrastructure.components.aws.database import OLAmazonDB, OLMariaDBConfig
 from ol_infrastructure.components.services.vault import (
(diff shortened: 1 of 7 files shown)


src/ol_infrastructure/applications/edxapp/__main__.py
```diff
@@ -1,28 +1,48 @@
...
+)
 from ol_infrastructure.lib.aws.iam_helper import IAM_POLICY_VERSION, lint_iam_policy
-from ol_infrastructure.lib.ol_types import AWSBase
+from ol_infrastructure.lib.aws.route53_helper import acm_certificate_validation_records
+from ol_infrastructure.lib.ol_types import Apps, AWSBase
 from ol_infrastructure.lib.pulumi_helper import parse_stack
 from ol_infrastructure.lib.stack_defaults import defaults
 from ol_infrastructure.lib.vault import mysql_role_statements
 
+SSH_ACCESS_KEY_NAME = "salt-qa"
```
shaidar, line 39: Shouldn't this live in the stack definition?

src/ol_infrastructure/applications/edxapp/__main__.py
```diff
@@ -1,28 +1,48 @@
...
+MIN_WEB_NODES_DEFAULT = 3
+MAX_WEB_NODES_DEFAULT = 15
+MIN_WORKER_NODES_DEFAULT = 1
+MAX_WORKER_NODES_DEFAULT = 5
```
shaidar, line 43: Also wondering whether those should be in the stack definition as well and that way we can have different numbers for QA and prod or am I missing something?

blarghmatey: Those are default values that take place if the stack doesn't define an alternate value

src/ol_infrastructure/applications/edxapp/__main__.py
blarghmatey: That was a placeholder. I am thinking about creating a new SSH key that is distinct from the Salt infrastructure.

src/ol_infrastructure/applications/edxapp/__main__.py
```diff
@@ -426,6 +461,246 @@
...
+        arn=edxapp_instance_profile.arn,
+    ),
+    image_id=edxapp_web_ami.id,
+    vpc_security_group_ids=[
+        edxapp_security_group.id,
+        edxapp_vpc["security_groups"]["web"],
+        consul_security_groups["consul_agent"],
+    ],
+    instance_type=InstanceTypes[web_instance_type].value,
+    key_name="",
```
shaidar, line 581: Doesn't this need to be defined?

blarghmatey: Apparently not since I was able to launch the instances with the code in that state... Definitely should be though.

shaidar:
Apart from key_name, looks good

PR #253, opened 2021-07-13
Added setup of SES for use with edxapp

blarghmatey:
- Create domain identity
- Crate mailfrom domain
- Create SPF and DKIM records
- Update edxapp config to use django-ses for email sending


src/ol_infrastructure/applications/edxapp/__main__.py
```diff
@@ -404,6 +423,68 @@
...
+    type="TXT",
+    ttl=FIVE_MINUTES,
+    records=["v=spf1 include:amazonses.com -all"],
+)
+edxapp_ses_domain_dkim = ses.DomainDkim(
+    "edxapp-ses-domain-dkim", domain=edxapp_ses_domain_identity.domain
+)
+edxapp_ses_dkim_records = []
+for loop_counter in range(0, 3):
+    edxapp_ses_dkim_records.append(
```
shaidar, line 471: Not sure I follow what we're trying to do appending dkim records

blarghmatey: That's just adding it to the list object. Right now we're not doing anything with it so I can remove it, but if we want to pass it along to anything else or export it then that's where we would use that list.

blarghmatey: Mostly copy :spaghetti: from an example

shaidar: I would either remove or maybe add a comment explaining this just for record keeping

shaidar:
👍

PR #259, opened 2021-07-16
Add dynamic rendering of edx configs and install forum on web nodes

blarghmatey:
- Adds consul-template to the build to allow for faster iteration on configuration values
- Adds configuration values for supporting mitxonline frontend
- Adds Forum to edxapp build to address errors on courseware rendering

PR #266, opened 2021-07-20
Adding separate target for Studio to allow for sticky sessions

blarghmatey:
Studio has some workflows that are stateful, such as importing and exporting courses which requires files to be written and read from the same EC2 instance. This adds separate target groups and ALB listener rules to route requests for studio to a target group with session stickiness enabled so that these stateful workflows don't fail.


shaidar:
👍

blarghmatey:
Closing as this commit got pulled into the open PR #259

PR #267, opened 2021-07-20
Trigger on github status instead of new commit

### LINKED ISSUE #249: Trigger edx-sysadmin publish based on status
>As it is right now it just watches for a new version in the repo which equates to a different commit hash. We should be able to add https://github.com/dpb587/github-status-resource as an input to make it trigger based on status.

https://github.com/mitodl/ol-infrastructure/pull/197#issuecomment-859774452

annagav:
Fix #249
The publish pipeline should get triggered when the release branch status changes to success.
Testing on https://cicd-qa.odl.mit.edu/teams/main/pipelines/publish-edx-sysadmin-to-pypi

### DIFF
--- a/pipelines/edx-sysadmin-publish.yml
+++ b/pipelines/edx-sysadmin-publish.yml
@@ -22,7 +22,7 @@ resources:
     type: github-status
     source:
       repository: mitodl/edx-sysadmin
-      access_token: {{repo_github_token}}
+      access_token: ((github.access_token))
       branch: release
   - name: edx-sysadmin-pypi
     type: pypi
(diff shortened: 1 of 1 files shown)


pipelines/edx-sysadmin-publish.yml
```diff
@@ -13,6 +18,12 @@ resources:
       name: edx-sysadmin
       uri: https://github.com/mitodl/edx-sysadmin.git
       branch: release
+  - name: edx-sysadmin-status
+    type: github-status
+    source:
+      repository: mitodl/edx-sysadmin
+      access_token: {{repo_github_token}}
```
blarghmatey, line 25: The var syntax uses `(())` rather than `{{}}`. Also, if you can change the var name to something like `github.access_token` that would be helpful so that we can populate a Vault path with that data for the pipeline to pull it out.

blarghmatey:
:+1:

PR #274, opened 2021-07-22
Consul ASG

ebberg:
Migrate Consul instances to a Consul ASG and migrate infrastructure configuration from Salt to PyInfra.

Work in Progress.


src/ol_infrastructure/infrastructure/consul/__main__.py
```diff
@@ -175,16 +179,13 @@
     if subnet_object.availability_zone == "us-east-1e":
         continue
     instance_name = f"consul-{environment_name}-{count}"
-    salt_minion = OLSaltStackMinion(
-        f"saltstack-minion-{instance_name}",
-        OLSaltStackMinionInputs(minion_id=instance_name),
-    )
 
+    # Will cloud init be needed for pyinfra, rather than salt, setup?
```
blarghmatey, line 183: We will still be using cloud-init to populate the datacenter information and the appropriate auto-join string. You can see an example of that [here](https://github.com/mitodl/ol-infrastructure/blob/vector/src/ol_infrastructure/applications/edxapp/__main__.py#L722-L745)

src/ol_infrastructure/infrastructure/consul/__main__.py
```diff
@@ -200,17 +201,17 @@
         tags=instance_tags,
         volume_tags=instance_tags,
         subnet_id=subnet,
-        key_name=salt_config.require("key_name"),
+        # key_name=salt_config.require("key_name"),
```
blarghmatey, line 204: The key name should be updated to `"oldevops"`

src/ol_infrastructure/infrastructure/consul/__main__.py
```diff
@@ -219,18 +220,80 @@
         "private_ip": consul_instance.private_ip,
         "ipv6_address": consul_instance.ipv6_addresses,
     }
+######
 
 fifteen_minutes = 60 * 15
 consul_domain = route53.Record(
     f"consul-{environment_name}-dns-record",
-    name=f"consul-{salt_environment}.odl.mit.edu",
+    # name=f"consul-{salt_environment}.odl.mit.edu",
```
blarghmatey, line 228: We still want to set up a DNS record with the name scoped to the environment, but we no longer need to use the salt environment as the canonical name. This should template in the `environent_name` variable

src/ol_infrastructure/infrastructure/consul/__main__.py
```diff
@@ -2,27 +2,31 @@
...
-)
 
-stack_info = parse_stack()
+
+stack_info = parse_stack()  # needed without salt?
 env_config = Config("environment")
 consul_config = Config("consul")
-salt_config = Config("saltstack")
-environment_name = f"{stack_info.env_prefix}-{stack_info.env_suffix}"
+environment_name = f"{stack_info.env_prefix}-{stack_info.env_suffix}"  # without salt format?
```
blarghmatey, line 29: Yes, this line is still relevant once we remove salt.

src/bilder/images/consul/deploy.py
```diff
@@ -0,0 +1,66 @@
...
+from bilder.components.hashicorp.steps import (
+    configure_hashicorp_product,
+    install_hashicorp_products,
+    register_services,
+)
+from bilder.facts import has_systemd  # noqa: F401
+
+VERSIONS = {  # noqa: WPS407
+    "caddy_route53": "v1.1.1",
+    "consul": "1.9.5",
```
blarghmatey, line 25: We should update this to something like `os.environ.get("CONSUL_VERSION", "1.10.0")` so that we can override the version in a Concourse pipeline. That will let us trigger a new AMI build when new versions are released and pass that version number without having to change the code here every time.

src/ol_infrastructure/infrastructure/consul/__main__.py
```diff
@@ -219,18 +220,80 @@
...
+)
+
+consul_launch_config = ec2.LaunchTemplate(
+    "consul-launch-template",
+    name_prefix=f"consul-{environment_name}-",
+    description="Launch template for deploying Consul cluster",
+    iam_instance_profile=consul_instance_profile.id,
+    image_id=consul_ami.id,
+    instance_type=InstanceTypes[instance_type_name].value,
+    # key_name=salt_config.require("key_name"),
```
blarghmatey, line 254: The key name here will also just be `"oldevops"`

src/ol_infrastructure/infrastructure/consul/__main__.py
```diff
@@ -219,18 +220,80 @@
...
+    tags=instance_tags,
+    tag_specifications=[
+        ec2.LaunchTemplateTagSpecificationArgs(
+            resource_type="instance",
+            tags=aws_config.merged_tags({"Name": f"consul-{stack_info.env_suffix}"}),
```
blarghmatey, line 259: We should use the `environment_name` so that it's more descriptive when looking in the AWS console.

src/ol_infrastructure/infrastructure/consul/__main__.py
```diff
@@ -219,18 +220,80 @@
...
+        consul_server_security_group.id,
+    ],
+)
+
+consul_asg = autoscaling.Group(
+    f"consul-{environment_name}-autoscaling-group",
+    availability_zones=availability_zones,
+    desired_capacity=3,
+    max_size=5,
+    min_size=3,
```
blarghmatey, line 278: We'll probably want to set the capacity/max_size/min_size to the same value and pull it from the stack config, defaulting to 3 if it's not present.

src/ol_infrastructure/infrastructure/consul/__main__.py
blarghmatey: On further consideration, we won't be able to use the instance addresses as the destination for the DNS record since they will be managed by the ASG. We may need to set up a load balancer to handle routing to the instances for API calls.

src/ol_infrastructure/infrastructure/consul/__main__.py
```diff
@@ -25,7 +25,7 @@
 
 stack_info = parse_stack()  # needed without salt?
 env_config = Config("environment")
-consul_config = Config("consul")
+consul_config = Config("consul")  # where is this located?
```
blarghmatey, line 28: This is a Pulumi `Config` object that pulls data from the stack settings which is a YAML file that neighbors this one. The config object will read values of the form `consul:foovar: foovalue` so that you can then do `consul_config.get('foovar')` to retrieve `foovalue`. (https://www.pulumi.com/docs/intro/concepts/config/#code)

src/ol_infrastructure/lib/consul.py
```diff
@@ -0,0 +1,49 @@
...
+def build_consul_userdata(env_name):
+    """ Builds consul userdata for cloud init, to create auto scale group and launch configs
+        TODO setup for consul_esm """
+    return base64.b64encode(
+        "#cloud-config\n{}".format(
+            yaml.dump(
+                {
+                    "write_files": [
+                        {
+                            "path": "/etc/consul.d/99-autojoin.json",
```
blarghmatey, line 16: For Consul servers, we will also need a way to pass in a list of datacenters that they will peer with across VPCs. That will go into a `retry_join_wan` configuration parameter. Would probably make sense to keep that as a separate file for simplifity (e.g. 99-autojoin-wan.json) (https://www.consul.io/docs/agent/options#_retry_join_wan)

src/ol_infrastructure/lib/consul.py
```diff
@@ -1,12 +1,17 @@
 import base64
 import json
-import textwrap
+
 import yaml
 
 
 def build_consul_userdata(env_name):
```
blarghmatey, line 7: This is going to need to be updated to accept a list of environment names for the WAN peers, and the autojoin_wan file should only be written out if it is a non-`None` value.

src/ol_infrastructure/lib/consul.py
```diff
@@ -25,6 +30,19 @@ def build_consul_userdata(env_name):
...
                             "owner": "consul:consul",
                         },
+                        {
+                            "path": "/etc/consul.d/99-autojoin-wan.json.json",
+                            "content": json.dumps(
+                                {
+                                    "retry_join_wan": [
+                                        "provider=aws tag_key=consul_env "
+                                        f"tag_value={env_name}"
+                                    ],
```
blarghmatey, line 40: We'll want to build up a list of strings for each entry in the wan_peer_envs (or whatever name we use) list so that the server will join up with all of them.

src/ol_infrastructure/lib/consul.py
```diff
@@ -25,6 +30,19 @@ def build_consul_userdata(env_name):
...
+                                    "datacenter": env_name,
```
blarghmatey, line 41: We don't need to reiterate the `datacenter` argument in this file.

src/ol_infrastructure/lib/consul.py
```diff
@@ -1,12 +1,17 @@
...
 
 def build_consul_userdata(env_name):
-    """ Builds consul userdata for cloud init, to create auto scale group and launch configs
-        TODO setup for consul_esm """
+    """Build consul userdata for cloud init, to create auto scale group and launch configs.
+
+    :param env_name: consul environment being built
+    :returns: base64-encoded yaml cloud init data
+
+    TODO setup for consul_esm
```
blarghmatey, line 13: There shouldn't actually need to be anything done in cloud-init for consul-esm. It should be able to be entirely naive to the environment that it's running in.

src/ol_infrastructure/infrastructure/consul/__main__.py
```diff
@@ -205,6 +231,47 @@
...
+                "write_files": [
+                    {
+                        "path": "/etc/consul.d/99-autojoin.json",
+                        "content": json.dumps(
+                            {
+                                "retry_join": [
+                                    "provider=aws tag_key=consul_env "
+                                    f"tag_value={env_name}"
+                                ],
+                                "datacenter": env_name,
```
blarghmatey, line 254: If we're using the VPC ID as the datacenter then we need to change it here too.

src/bilder/images/consul/deploy.py
```diff
@@ -0,0 +1,68 @@
...
+
+VERSIONS = {  # noqa: WPS407
+    "caddy_route53": "v1.1.1",
+    "consul": os.environ.get("CONSUL_VERSION", "1.10.0"),
+}
+
+install_baseline_packages()
+consul_configuration = {
+    Path("00-default.json"): ConsulConfig(
+        acl=ConsulACL([ConsulACLToken(master="<value>")]),
```
ebberg, line 32: Is this needed here? Seems like it needs to bootstrap first, then have ACLTokens added?

blarghmatey: The master ACL token is used to bootstrap the overall ACL system (which we aren't currently using really, but the master token is still relevant for the server cluster). I've got a branch started for integrating Sops with a Python wrapper to populate secrets like that one at build time.

blarghmatey: Alternatively, once I get the Vault ASG deployment running we can pull in values like that via the Vault agent and templates.

src/bilder/components/hashicorp/consul/models.py
```diff
@@ -101,9 +101,9 @@ class Consul(HashicorpProduct):
     def systemd_template_context(self):
         return self
 
-    def render_configuration_files(self) -> Iterable[Tuple[Path, str]]:
+    def render_configuration_files(self) -> Iterable[Tuple[str, str]]:
         for fpath, config in self.configuration.items():  # noqa: WPS526
-            yield self.configuration_directory.joinpath(fpath), config.json(
+            yield str(self.configuration_directory.joinpath(fpath)), config.json(
```
blarghmatey, line 106: Rather than doing these string coercions I've got PyInfra pinned to 1.4.4 which still works with Path objects as inputs. We should stick with that until the project adds support for Paths in a more holistic manner.

src/bilder/components/caddy/steps.py
```diff
@@ -87,7 +87,7 @@ def install_caddy(caddy_config: CaddyConfig, state=None, host=None):
     files.put(
         name="Configure systemd to load environment variables from file",
         dest="/etc/systemd/system/caddy.service",
-        src=Path(__file__).parent.joinpath("templates", "caddy.service.override"),
+        src=Path(__file__).parent.joinpath("templates", "caddy.service.j2"),
```
blarghmatey, line 90: The override file is intended to be loaded here, the problem is that it needs to go under `/etc/systemd/system/caddy.service.d/env_override.conf`

Issue #149, opened 2021-05-16
Automate deployment of Vault clusters with auto-unseal


blarghmatey:
We are currently deploying Vault using SaltStack and the service is co-located with the Consul servers. We would like to automate management of the clusters to be deployed using pre-built AMIs that get launched into an Amazon auto-scale group. As part of this, we should migrate to use the auto-unseal functionality with the AWS KMS integration. This will make it easier to deploy more clusters of Vault to more easily separate and scale our usage of Vault across different logical boundaries.

To make that happen we will need to write the PyInfra and Packer configurations to build the images. We will then need Pulumi code to create the auto-scale groups and provision the necessary KMS keys.

PR #287, opened 2021-07-29
Deploy Vault servers in an autoscale group with KMS auto-unseal

blarghmatey:
In order to support a scalable method of managing secrets across our infrastructure we need an easy way to deploy Vault clusters. This adds code to build Vault server images that can be deployed in an autoscale group without requiring any additional systems being in place beyond AWS services.


blarghmatey:
Closes #149

src/bilder/components/hashicorp/vault/models.py
```diff
@@ -67,12 +70,17 @@ class VaultAgentCache(FlexibleBaseModel):
     use_auto_auth_token: Union[str, bool] = True
 
 
-class VaultSealConfig(FlexibleBaseModel, abc.ABC):
-    pass  # noqa: WPS420, WPS604
+class VaultAwsKmsSealConfig(FlexibleBaseModel):
```
shaidar, line 73: What does the `FlexibleBaseModel` provide us? I see the class definition, but not sure I understand what it's helping us with

blarghmatey: It lets us add additional, arbitrary attributes to the class object without it throwing an exception. The intent being that we can add configuration parameters for classes that aren't completely fleshed out to match what the project is expecting for configuration inputs.

blarghmatey:
Going to merge this as-is and do the review post-hoc to find and follow-up with any style/organizational changes.

PR #303, opened 2021-08-10
Hide the 'country' field for registration

rhysyngsun:
This hides the country field so that users can login without being prompted to fill it out.

See https://github.com/mitodl/salt-ops/blob/ab5b7491de4fd5b63a432123b2a0d9f6b4919bb5/pillar/edx/ansible_vars/xpro.sls#L71


blarghmatey:
:+1:

PR #310, opened 2021-08-11
Enable the mobile user api for micromasters

rhysyngsun:
MicroMasters requires this API in order to retrieve user data.


pdpinch (approved):
📱

PR #333, opened 2021-08-19
Redirect the user on openedx if they go to /login

### LINKED ISSUE #329: Configure the openedx extensions plugin for MITx Online
We're observing that the user ends up at /login on OpenedX and has to click the button in order to continue, they should instead be redirected to MITx Online.

This probably needs to be done by installing and configuring https://github.com/mitodl/mitxpro-openedx-extensions

- [ ] Verify the extension is being installed
- [ ] Configure additional settings if necessary

rhysyngsun:
Fixes #329 

Removes `/login` as a url that is allowed without redirect. `django-oauth-toolkit` appears to redirect to `/login` so when attempting to start an oauth login the user ends up seeing this page when they shouldn't.


blarghmatey:
My only concern here is whether this will cause problems with login from studio, as that redirects to LMS for auth. I think we just merge it and see what happens. :+1:

PR #336, opened 2021-08-20
set SKIP_EMAIL_VALIDATION to true

pdpinch:
email validation will be handled by the marketing app

Questions: 
- [ ] should I be adding this `# Modified` comment, like you do?
- [ ] will this change apply to QA and production?


blarghmatey:
In the case where we're adding a new line that wasn't present I've been annotating it as `# ADDED` so that we can identify things that aren't part of the vanilla Ansible build. That will help us going forward as we figure out how to feed back upstream with a sane set of defaults for edx-platform.

And yes, this will apply to both production and QA

PR #338, opened 2021-08-20
Renaming learning-mfe to mfe

shaidar:
Changed learning-mfe to mfe. Already made the change on fastly and the pipeline


blarghmatey:
:+1:

PR #339, opened 2021-08-20
Adding more templating of domains for edxapp config

blarghmatey:
- Template and manage MFE domain for edxapp
- Template value for marketing domain and populate the value from the mitxonline project


src/ol_infrastructure/applications/mitxonline/__main__.py
```diff
@@ -162,4 +163,14 @@
 )
 mitxonline_vault_backend = OLVaultDatabaseBackend(mitxonline_vault_backend_config)
 
+# Set Consul key for use in edxapp configuration template
+consul.Keys(
+    "mitxonline-app-domain-for-edxapp",
+    keys=[
+        consul.KeysKeyArgs(
+            path="edxapp/marketing-domain", value=mitxonline_config.require("domain")
```
shaidar, line 171: Doesn't this also need the edxapp/mfe-domain?

blarghmatey: No, because that's getting set in the edxapp project.

PR #344, opened 2021-08-23
Changed service role path

shaidar:
Based on [AWS Case ID](https://console.aws.amazon.com/support/home#/case/?displayId=8719817001&language=en) MediaConvert role is expected to have the following format:
`arn:aws:iam::0000000000:role/ROLENAME format`

Any nesting in the role name was leading to a failure in parsing and thus making an anonymous call to S3 which was naturally failing. This changes the role name and removes the path nesting.


blarghmatey:
:+1:

PR #348, opened 2021-08-24
Fixing compilation of theme during AMI build

blarghmatey:
The theme wasn't being compiled fully during the AMI build because the configuration necessary to point the update-assets script at the right location isn't getting loaded until the instance is launched. This adds the necessary Ansible values to the playbook file so that the minimal necessary config is in place at build/compilation time.

### DIFF
--- a/src/bilder/images/edxapp/prebuild.py
+++ b/src/bilder/images/edxapp/prebuild.py
@@ -1,25 +1,36 @@
-from pyinfra.operations import files, git
+from pyinfra.operations import files, git, server
 
 from bilder.components.baseline.steps import install_baseline_packages
 from bilder.images.edxapp.lib import WEB_NODE_TYPE, node_type
 
+EDX_USER = "edxapp"
+
 install_baseline_packages(
     packages=[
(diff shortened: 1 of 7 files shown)


src/bilder/images/edxapp/prebuild.py
```diff
@@ -0,0 +1,31 @@
...
+
+if node_type == WEB_NODE_TYPE:
+    files.directory(
+        name="Ensure themes directory is present",
+        path="/edx/app/edxapp/themes/",
+        present=True,
+    )
+    git.repo(
+        name="Load theme repository",
+        src="https://github.com/mitodl/mitxonline-theme",
```
shaidar, line 27: Should we also make this generic or be part of the config so that it can be used with other builds?

blarghmatey: Once we start expanding this to work in other environments we will make that value configurable. For the time being I'm fine with it being hard-coded for the sake of simplicity.

shaidar:
👍

PR #349, opened 2021-08-24
Adding validation of sender email address for edxapp

blarghmatey:
In order for us to send email via SES from an address that doesn't use a validated domain then we need to also validate the address.


shaidar:
👍

PR #363, opened 2021-09-01
enable embargo feature

pdpinch:
This should allow us to block access, on a per course basis, from certain countries. 

from the [documentation](https://edx.readthedocs.io/projects/edx-platform-technical/en/latest/featuretoggles.html#featuretoggle-FEATURES['EMBARGO']):

>reverse proxy should be configured appropriately for example Client IP address headers (e.g HTTP_X_FORWARDED_FOR) should be configured.

This setting may have some unintended side effects, as well as some undocumented dependencies. We should enable it in a CI or QA environment first, for testing.


blarghmatey:
:+1:

PR #375, opened 2021-09-10
Updating edxapp and Vector to use our logging plugin

blarghmatey:
- Adds a log directory for writing logs to
- Installs our logging plugin for edx-platform
- Updates Vector configuration to read JSON logs from disk for edx-platform

### DIFF
--- a/src/bilder/images/edxapp/templates/vector/edxapp.yaml
+++ b/src/bilder/images/edxapp/templates/vector/edxapp.yaml
@@ -8,9 +8,10 @@ log_schema:
 
 sources:
   edxapp:
-    type: journald
-    include_units:
-    - supervisor
+    type: file
+    file_key: log_file
+    include:
+    - /var/log/edxapp/*.log
   auth_log:
     type: file
     file_key: log_file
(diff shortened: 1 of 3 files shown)


src/bilder/images/edxapp/templates/vector/edxapp.yaml
```diff
@@ -19,19 +20,21 @@ sources:
...
-      .environment = "${ENVIRONMENT}"
+      event, err = parse_json(.message)
+      if event != null {
+        . = merge!(., event)
+        .@timestamp = .timestamp
+        .labels = ["edxapp", "${ENVIRONMENT}"]
+        .application = "edxapp"
+        .environment = "${ENVIRONMENT}"
+      } else {
+        .malformed = true
```
orchidpatch, line 36: should this be documented in an informal schema, or otherwise made obvious, in case of a change and sudden ingress of malformed logs into ES?

orchidpatch (approved):
LGTM -- one minor comment that shouldn't block this.

src/bilder/images/edxapp/templates/vector/edxapp.yaml
orchidpatch: I think the general concept is good tho, and every log can and should have an 'event' tag in the json

blarghmatey: Definitely agree that we need to establish a core schema for log messages and labeling. The `event, err` piece is just arbitrary temporary variables that are used to store the results of the `parse_json` function call. The contents of `event` are just whatever attributes are contained in the JSON message.

PR #381, opened 2021-09-17
Adding Vector and Concourse configurations to ship metrics to Cloudwatch

blarghmatey:
In order for us to autoscale Concourse worker pools intelligently based on queued jobs we need to be able to populate Cloudwatch with the useful metrics.

- Add Vector install to Concourse
- Add IAM permissions to write metrics to cloudwatch
- Add Vector configuration to read prometheus metrics from Concourse and write to Cloudwatch

### DIFF
--- a/src/bilder/images/concourse/deploy.py
+++ b/src/bilder/images/concourse/deploy.py
@@ -1,5 +1,6 @@
 import os
 from functools import partial
+from ipaddress import IPv4Address
 from pathlib import Path
 from typing import Union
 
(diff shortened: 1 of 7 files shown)


src/bilder/images/concourse/deploy.py
```diff
@@ -103,6 +110,8 @@
         vault_url=f"http://localhost:{VAULT_HTTP_PORT}",
         vault_client_token="this-token-gets-overridden-by-the-vault-agent",
         vault_path_prefix="/secret-concourse",
+        prometheus_bind_ip=IPv4Address("127.0.0.1"),
+        prometheus_bind_port=9391,
```
shaidar, line 114: Should this be a `magic` number?

src/bilder/images/concourse/deploy.py
```diff
@@ -50,16 +51,23 @@
...
 from bilder.components.hashicorp.vault.steps import vault_template_permissions
+from bilder.components.vector.models import VectorConfig
+from bilder.components.vector.steps import (
+    configure_vector,
+    install_vector,
+    vector_service,
+)
 from bilder.facts import has_systemd  # noqa: F401
 from bridge.lib.magic_numbers import (
+    CONCOURSE_DEFAULT_PROMETHEUS_PORT,
```
shaidar, line 62: Think the name is a little confusing here. Is it the default prometheus port or concourse or ...?

blarghmatey: So, it is the default port on which Concourse exposes Prometheus metrics according to the [table of port number allocations](https://github.com/prometheus/prometheus/wiki/Default-port-allocations) in the Prometheus wiki. Happy to change the name if there's a way to make that clearer.

shaidar: How about `CONCOURSE_GRAPHITE_EXPORTER_PORT`?

shaidar:
👍

PR #387, opened 2021-09-24
Adding configuration to send edX logs and metrics to Grafana Cloud

blarghmatey:
- Add Vector configuration for sending edxapp logs and metrics to Grafana Cloud
- Adding configuration in cloud-init to populate credentials for Vector to use when sending data
- Updating edx-username-changer version


shaidar:
👍

Issue #328, opened 2021-08-18
Script for initial setup of new Vault cluster


blarghmatey:
We have added support for launching and upgrading Vault clusters running in an auto-scale group. When a new cluster is launched we need to initialize it and do initial setup so that it is usable with Pulumi for populating mounts and secret values. This can be managed via a script using the [HVAC](https://hvac.readthedocs.io/en/stable/) library.

The steps to execute are:
- [ ] Initialize vault cluster with number of key shares and unseal threshold (preferably using GPG keys)
- [ ] Set up audit device (file writing to stdout that will be captured by the SystemD journal)
- [ ] Mount authentication backend(s) to be used by programmatic and human users. Initial setup should likely include username/password auth and create a user account for Pulumi
- [ ] Revoke root token returned by initialization step

blarghmatey:
This snippet has some useful example code for the initialization step https://hvac.readthedocs.io/en/stable/usage/index.html#initialize-and-seal-unseal

PR #409, opened 2021-10-15
Automate initialization of newly deployed Vault clusters

blarghmatey:
This is based on https://github.com/mitodl/ol-infrastructure/pull/408 which needs to merge first.

Once we deploy a new Vault cluster into an autoscaling group we need to be able to initialize it and set up the initial set of authentication and audit backends. This adds a script that will initialize the cluster, log the recovery keys, and provision an account to be used by Pulumi for managing the cluster going forward.


blarghmatey:
Addresses https://github.com/mitodl/ol-infrastructure/issues/328

shaidar:
👍

PR #415, opened 2021-10-25
Vector config for consul and vault observability

shaidar:
Added configs to collect metrics and logs from vault and consul clusters and ship them to Grafana.


src/ol_infrastructure/infrastructure/consul/__main__.py
```diff
@@ -400,6 +416,9 @@ def cloud_init_userdata(
             init_dict["retry_join_wan"],
             consul_dns_name,
             init_dict["pulumi_password"],
+            init_dict["grafana_api_key"],
+            init_dict["grafana_prometheus_user"],
+            init_dict["grafana_loki_user"],
```
blarghmatey, line 421: These will need to be added to the `Output.all` parameters and pulled from the stack configs.

src/ol_infrastructure/infrastructure/vault/__main__.py
```diff
@@ -426,6 +449,9 @@ def cloud_init_user_data(
         init_inputs["tls_key"],
         init_inputs["tls_cert"],
         init_inputs["ca_cert"],
+        init_inputs["grafana_api_key"],
+        init_inputs["grafana_prometheus_user"],
+        init_inputs["grafana_loki_user"],
```
blarghmatey, line 454: These will need to be added to the `Output.all` parameters and pulled from the stack configs.

src/ol_infrastructure/infrastructure/consul/__main__.py
blarghmatey: You should be able to avoid having to pass them through the lambda by using the `read_yaml_secrets` method inside the cloud-init function. For example: https://github.com/mitodl/ol-infrastructure/blob/maple_ami/src/ol_infrastructure/applications/edxapp/__main__.py#L934-L936

shaidar: How is that grafana yaml file generated?

blarghmatey: I created it using SOPS and checked it into the repository for the CI, QA, and Production environments.

shaidar: > These will need to be added to the `Output.all` parameters and pulled from the stack configs.

By stack configs are u referring to the grafana sops file where the creds are written to?  In looking at the code, unlike the vault one below, I can't use `grafana_credentials` as the Output function is outside the cloud-init one, so do we make another call to `read_yaml_secrets`? I'm also not very clear on what purpose this `Output.all` serves here.

blarghmatey:
:+1:

PR #417, opened 2021-10-27
Added SOPS management for Consul password and updated stacks

blarghmatey:
- Removed Salt references in stack configs now that we aren't relying on salt for configuring the instances
- Added code to pull the Consul basic auth password from SOPS files


shaidar:
👍

PR #429, opened 2021-11-03
Adding Opensearch cluster deployment via AWS managed service

blarghmatey:
In order to simplify the work of deploying and managing our Elasticsearch/OpenSearch clusters we are migrating to use the AWS managed service. This adds a project to launch a cluster.


shaidar:
👍

PR #440, opened 2021-11-09
Adding pipeline definition for running ocw-to-hugo

blarghmatey:
In order to manage the cutover from Plone to ocw-studio we need a way to keep the content in sync that is independent of the EC2 build server that is deprecated. This adds a pipeline definition that executes the ocw-to-hugo utility for converting the Plone generated JSON files to the Markdown format expected by Hugo and then uploads that output to the S3 bucket used by OCW Studio to import the legacy content.

### DIFF
--- a/pipelines/ocw/ocw-to-hugo.yml
+++ b/pipelines/ocw/ocw-to-hugo.yml
@@ -1,0 +1,56 @@
+---
+jobs:
+- name: convert-plone-to-hugo
+  plan:
+  - get: ocw-to-hugo
+    trigger: true
+  - get: ocw-json
+    trigger: true
+  - config:
+      image_resource:
+        name: ''
+        source:
+          repository: node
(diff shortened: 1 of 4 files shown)


pipelines/ocw/ocw-to-hugo.yml
```diff
@@ -0,0 +1,53 @@
...
+  - put: ocw-markdown
+resource_types:
+- name: s3-sync
+  source:
+    repository: mitodl/concourse-s3-sync-resource
+  type: docker-image
+resources:
+- name: ocw-json
+  source:
+    bucket: open-learning-course-data-rc
```
gumaerc, line 40: Shouldn't this be templated based on the environment, as well as the output bucket below?

gumaerc (changes requested):
Just one thing I noticed:

gumaerc (approved):
:+1:

PR #441, opened 2021-11-10
updated environment variable to allow staff to download course grades and ORA report downloasd.

Wassaf-Shahzad:
# Description
Updated `ALLOW_COURSE_STAFF_GRADE_DOWNLOADS `  to allow course grades and ORA report to download for course staff with access role data_researcher.

Closes https://github.com/mitodl/mitxonline/issues/275


src/bilder/images/edxapp/templates/common_values.yml
```diff
@@ -284,6 +284,7 @@ FEATURES:
     SHOW_HEADER_LANGUAGE_SELECTOR: false
     SKIP_EMAIL_VALIDATION: true # ADDED KEY
     ALLOW_PUBLIC_ACCOUNT_CREATION: true
+    ALLOW_COURSE_STAFF_GRADE_DOWNLOADS: true
```
blarghmatey, line 287: If you can just append a comment of `# ADDED KEY` that would be great. I've been doing that so we can track what values we are changing and adding to support our installation as a way of making it easier to understand how we can build a set of common defaults and manage our different installations.

blarghmatey:
:+1:

PR #442, opened 2021-11-10
[WIP] Initial pass at a vault snapshot pipeline

blarghmatey:
In order to be able to keep our Vault deployments healthy we need a way to create periodic backups to allow for point in time recovery. This is a first pass at creating a Concourse pipeline to generate a Raft snapshot and upload it to S3 on an hourly basis.


blarghmatey:
This has been addressed with a script on the Vault instances themselves.

Issue #444, opened 2021-11-11
copy legacy OCW static assets during ocw-to-hugo pipeline run


gumaerc:
In https://github.com/mitodl/ol-infrastructure/pull/440, we created a definition for a standalone pipeline that will run `ocw-to-hugo` and upload the output to the proper S3 bucket for consumption by `ocw-to-hugo`.  This pipeline does not currently handle the static assets (pdfs, images, etc.) associated with any courses and therefore when the site is imported into `ocw-studio` and deployed they are not included.  The `ocw-to-hugo` pipeline should handle this and make sure the static assets end up in the proper folder inside the `ocw-studio` buckets as part of the run.

PR #445, opened 2021-11-12
sync static resources during ocw-to-hugo run

gumaerc:
#### Pre-Flight checklist

- [ ] Testing
  - [ ] Changes have been manually tested

#### What are the relevant tickets?
Closes https://github.com/mitodl/ol-infrastructure/issues/444

#### What's this PR do?
This PR adds a step to the `ocw-to-hugo.yml` pipeline definition that will sync static resources (pdfs, images, etc.) from a defined `open-learning-course-data-bucket` to an `ol-ocw-studio-app-bucket` as part of the pipeline run, ensuring that these will be in place when the converted courses are eventually published by `ocw-studio`.  Vars for RC and production have been included.

#### How should this be manually tested?
 - Make sure you have a Concourse target locally set up to RC on the `ocw` team.  I named mine `rc`.
 - Upsert the pipeline to Concourse RC with ` fly -t rc set-pipeline --pipeline ocw-to-hugo-rc --config pipelines/ocw/ocw-to-hugo.yml --load-vars-from pipelines/ocw/vars/ocw-to-hugo/rc.yml`
 - Run the pipeline and ensure it completes successfully without error

### DIFF
--- a/pipelines/ocw/ocw-to-hugo.yml
+++ b/pipelines/ocw/ocw-to-hugo.yml
@@ -46,13 +46,12 @@ jobs:
       run:
         args:
         - -exc
-        - >
-          yarn install --pure-lockfile
-
-          node . \
-          -i ../ocw-json \
-          -o ../ocw-markdown \
-          --staticPrefix /courses \
+        - >-
+          yarn install --pure-lockfile &&
+          node .
+          -i ../ocw-json
+          -o ../ocw-markdown
+          --staticPrefix /courses
           --strips3
         dir: ocw-to-hugo
         path: sh
--- a/pipelines/ocw/ocw-to-hugo.yml
+++ b/pipelines/ocw/ocw-to-hugo.yml
@@ -20,15 +20,15 @@ jobs:
         path: sh
         args:
         - -exc
-        - |-
-          aws s3 sync \
-          s3://((open-learning-course-data-bucket))/ \
-          s3://((ol-ocw-studio-app-bucket))/courses/ \
-          --include '*' \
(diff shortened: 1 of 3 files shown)


pipelines/ocw/ocw-to-hugo.yml
```diff
@@ -22,9 +46,14 @@ jobs:
       run:
         args:
         - -exc
-        - |
+        - >
```
blarghmatey, line 49: This will break unless we add an && between the `yarn` and `node` commands because the `>` operator will render all following contents onto one line, whereas the `|` operator will retain newlines.

pipelines/ocw/ocw-to-hugo.yml
```diff
@@ -1,11 +1,35 @@
...
+        name: ''
+        source:
+          repository: amazon/aws-cli
+          tag: latest
+        type: docker-image
+      run:
+        path: sh
+        args:
+        - -exc
+        - |-
```
blarghmatey, line 23: For this line we can use the `>` operator and remove the `\` separators.

pipelines/ocw/ocw-to-hugo.yml
gumaerc: I ended up putting the && after the `yarn` step.  The yarn lint pre-commit hook thinks the line is too long and keeps moving the `--strips3` argument to its own line which would run `ocw-to-hugo` without it.

gumaerc:
@blarghmatey ready for another look

blarghmatey:
:+1:

PR #446, opened 2021-11-12
Required ProctorTrack setting

shaidar:


src/bilder/images/edxapp/templates/common_values.yml
```diff
@@ -378,15 +378,16 @@ PLATFORM_NAME: MITx Online  # MODIFIED
...
 PROCTORING_BACKENDS:
     DEFAULT: 'proctortrack'
     'proctortrack':
-      {{ with secret "secret-mitxonline/edxapp" }}
       'client_id': '{{ .Data.proctortrack_client_id }}'
       'client_secret': '{{ .Data.proctortrack_client_secret }}'
-      {{ end }}
       'base_url': '{{ key "edxapp/proctortrack-base-url" }}'
     'null': {}
+PROCTORING_USER_OBFUSCATION_KEY: {{ .Data.proctortrack_user_obfuscation_key }}
```
blarghmatey, line 389: From the documentation, isn't this just a copy of the Django secret key? Should we just use the existing reference rather than duplicating it as another value?

shaidar: Looks at the docs, it seems that they use both keys. There's nothing stopping from using the same key, but it's probably better to use a separate one if we can.

blarghmatey:
:+1:

PR #454, opened 2021-11-17
updated tag and added option to avoid conflict

Wassaf-Shahzad:
Updated tag and added command to avoid peer dependency collision


blarghmatey:
:+1:

blarghmatey (approved):

Issue #458, opened 2021-11-19
s3 sync on ocw-to-hugo Concourse job fails


gumaerc:
### Steps to Reproduce

 - Run the `ocw-to-hugo-rc` pipeline in Concourse

### Expected Behavior

 - The pipeline completes successfully

### Actual Behavior

 - The pipeline fails on the S3 sync step with errors

### Stacktrace

```
copy failed: s3://open-learning-course-data-rc/res-15-004-system-dynamics-systems-thinking-and-modeling-for-a-complex-world-january-iap-2020/811446c0ce86ebd1d483baa5a5fb3eda_MITRES15-004IAP20-th.jpg to s3://ol-ocw-studio-app-qa/courses/res-15-004-system-dynamics-systems-thinking-and-modeling-for-a-complex-world-january-iap-2020/811446c0ce86ebd1d483baa5a5fb3eda_MITRES15-004IAP20-th.jpg An error occurred (AccessDenied) when calling the CopyObject operation: Access Denied
etc...
```

PR #459, opened 2021-11-19
add s3:CopyObject permission to OCW IAM profile

gumaerc:
#### What are the relevant tickets?
Closes https://github.com/mitodl/ol-infrastructure/issues/458

#### What's this PR do?
In https://github.com/mitodl/ol-infrastructure/pull/445, we added a step to the `ocw-to-hugo` Concourse pipeline that syncs resources from a configured `open-learning-course-data` bucket to the appropriate `ol-ocw-studio-app` bucket, so when these courses are eventually imported all of the static resources are in the right place. This PR adds the `s3:CopyObject` permission to the Concourse IAM profile for OCW S3 buckets which was missed in the original PR and is necessary to avoid permissions errors during the sync step.

#### How should this be manually tested?
Needs to be tested in QA once the new permissions are applied.

### DIFF
--- a/src/ol_infrastructure/applications/concourse/__main__.py
+++ b/src/ol_infrastructure/applications/concourse/__main__.py
@@ -114,7 +114,7 @@ concourse_iam_permissions = {
         },
         {
             "Effect": "Allow",
-            "Action": ["s3:GetObject*", "s3:ListBucket"],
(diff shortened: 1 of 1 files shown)


blarghmatey:
:+1:

PR #475, opened 2021-11-29
Added ES policy to edxapp

shaidar:
This is currently failing when I tried it out with the following error:
```TypeError: Object of type Output is not JSON serializable```

I tried different combinations for rendering `edxapp_vpc["cidr"]` however it was either the error above or it ended up returning an object instead of the IP strings


blarghmatey:
So, the reason that it won't render is that the `edxapp_vpc` object is a `pulumi.Output` which means we would need to pass the whole policy through a `.apply` call. That being said, I'm pretty sure this policy needs to be set on the ES domain resource, not on the edxapp IAM role.

blarghmatey:
I ended up fixing this with a permissive policy and some tweaks to the Forum code.

PR #476, opened 2021-11-29
Adding domain access policy to OpenSearch domain deployment

blarghmatey:


blarghmatey:
This is an invalid approach. OpenSearch domains in a VPC can't use IP access policies, they need to use security groups and sign the requests.

PR #480, opened 2021-12-01
Fix concourse permissions

shaidar:
OCW publishing was failing with error below:
```An error occurred (AccessDenied) when calling the GetObjectTagging operation: Access Denied```

PR adds the necessary permissions


blarghmatey:
:+1:

PR #484, opened 2021-12-03
Expanded concourse worker disk size

shaidar:


blarghmatey:
Looks like you added in some unrelated changes. Guessing that wasn't intentional?

shaidar:
I had this from before but hadn't committed them and figured might as well updated that stuff

blarghmatey:
The problem there is that I'm seeing locks for packages that we aren't actually using (e.g. sops, that's a Go binary, we're not using the Python package as it's deprecated). Also, I'm unclear what the difference is in the committed SOPS binary. Is that a newer version than what is currently vendored?

blarghmatey:
:+1:

PR #486, opened 2021-12-03
Added logic to check out our custom branch of edx-platform

blarghmatey:
We often have to add custom logic for our installations of edx-platform to handle use cases that are unique to us. This adds a step to check out our branch of the platform during the AMI build that contains any patches that we need to maintain for a given release. Rather than fall back on conditional logic, this sets the origin and branch for any unmodified deployments to be the same as what is already present on disk which ends up being a no-op.


shaidar:
👍

PR #493, opened 2021-12-06
For ocw themes, use s3 copy, narrow down copy path

### LINKED ISSUE #485: OCW theme assets pipeline S3 syncs are overwriting each other
If you run the release version of ocw-theme-assets, the existing release-candidate tarball in S3 will be erased, and vice-versa.  The `sync` command should be changed to a `copy`

mbertrand:
Fixes #485

Changes have been uploaded and tested on RC


blarghmatey:
:+1:

PR #494, opened 2021-12-06
Remove stub storage config from Vault build to fix auto clustering

blarghmatey:
The specific configuration for the raft cloud auto-join is written in by cloud-init. Having the stub storage configuration in the pre-baked config file conflicts with loading the full storage configuration at run time which prevents the node from automatically joining the cluster. This sets the storage block to optional in the Vault config model so that we can exclude it from the baked image in order for the proper config to load and join the running cluster.


shaidar:
👍

PR #495, opened 2021-12-07
Fixing metrics and log configurations for Consul and Vault

blarghmatey:
- Fix missing parameters for telemetry configuration blocks in Vault and Consul builds
- Fix prometheus endpoints used for scraping metrics from Consul and Vault


shaidar:
👍

PR #512, opened 2021-12-20
asadiqbal08/suggested_price should not be null

asadiqbal08:
fixes: https://github.com/mitodl/mitx-theme/issues/136

`COURSE_MODE_DEFAULTS` has honor slug and with some suggested price is not set in it. If that is the case then `OpenEdx` is keeping the course behind the paywall by following this [condition](https://github.com/edx/edx-platform/blob/open-release/maple.master/common/djangoapps/course_modes/models.py#L729-L754)


blarghmatey:
:+1:

PR #513, opened 2021-12-20
Adding fixes for xPro production edxapp deployment

blarghmatey:
- Added bucket policy to allow video thumbnails to be loaded from S3
- Added fix for worker AMI lookup filters
- Toggled dns management for xPro production on


shaidar:
👍

PR #515, opened 2021-12-21
Adding values for edx_api_key to address enrollment bug in xPro

blarghmatey:
Adding back the edx_api_key setting so that it can be used in requests from the marketing application to handle enrollments via the REST API.


src/bilder/images/edxapp/templates/edxapp/mitxonline/lms_only.yml
```diff
@@ -1,6 +1,7 @@
 {{ with secret "secret-mitxonline/edxapp" }}
 SOCIAL_AUTH_OAUTH_SECRETS:
   mitxpro-oauth2: {{ .Data.mitxonline_oauth_secret }}
+EDX_API_KEY: {{ .Data.edxapp_api_key }}  # DEPRECATED SETTING
```
shaidar, line 4: Why does it still say deprecated setting when we're using it?

shaidar:
👍

PR #535, opened 2022-01-07
Resize Concourse Prod DB

shaidar:
Production DB is currently being resized through the RDS console and this change is just to make sure it's in code.


blarghmatey:
:+1:

PR #546, opened 2022-01-10
Enabling P2P Volume streaming on concourse workers.

Ardiea:
This will enable P2P volume streaming between concourse workers within an environment. This should lead to a performance boost on OCW jobs where related tasks cannot be scheduled on the same node. Rather than having to stream the volume to the next worker via a web node, it can now stream directly.

Still need to research what SG configuration changes are needed (if any).


blarghmatey:
:+1:

PR #549, opened 2022-01-12
set DISABLE_START_DATES to false

pdpinch:
fix https://github.mit.edu/mitx/spring-2022/issues/24

To test: 

1. Create a course with a start date in the past
2. Create a section in the course with a release date in the future. 
3. Create a subsection and a Unit within the section, to make it easier to see the content
4. Publish the course.
5. View the section with the release date in the future as an admin and use staff debug to confirm that the section's release is "Not Yet" 
5. Masquerade as a student (or login as a student) and attempt to view the section =

Before this change:

You will be able to view the section, despite the release date. 

After the change:

You won't be able to view the section, because it isn't released yet. 

Documentation for this setting doesn't mention its affect on section and subsection release dates: https://edx.readthedocs.io/projects/edx-platform-technical/en/latest/featuretoggles.html#featuretoggle-FEATURES['DISABLE_START_DATES']

However, examining the relevant code makes it clear that a setting of true will cause all release dates to be ignored: 
https://github.com/edx/edx-platform/blob/3168673ecfd462d2f8afcd72fc6e25d01897f154/lms/djangoapps/courseware/access_utils.py#L67-L83


pdpinch:
@blarghmatey I wasn't sure whether it was better to remove the key, or set it explicitly to the default. This key is not set in mitxonline, or xPRO, neither of which exhibits the bug.

blarghmatey (approved):
:+1:

Issue #552, opened 2022-01-12
ocw theme assets pipeline missing search URL


gumaerc:
The `ocw-theme-assets` build does not set the `SEARCH_API_URL` env variable before running the `ocw-hugo-themes` webpack build, resulting in the built site having a non-functioning search page.  This value needs to be set by the pipeline definition per environment.

gumaerc:
Closed by committing https://github.com/mitodl/ol-infrastructure/commit/bf4721afd93bdf5ecce8b60d81a814e572cdb204 directly into main due to issues with pre-commit

PR #553, opened 2022-01-12
add search url var and set up theme assets pipeline to use vars files

gumaerc:
Closes https://github.com/mitodl/ol-infrastructure/issues/552

This PR adds `SEARCH_API_URL` to the `params` in the `build-ocw-www-hugo-theme` task in `pipeline-theme-assets.yml`, which gives the search app the proper URL to hit the API on.  Some vars `.yml` files were also set up for the different environments and the script for updating the pipeline was updated to use them.


gumaerc:
Closing this and merging changes directly into `main` per @blarghmatey

PR #559, opened 2022-01-14
update pipeline for new s3 sync bucket sync feature

mschuchard:


pipelines/ocw/ocw-to-hugo.yml
```diff
@@ -7,29 +7,8 @@ jobs:
...
-          s3://((open-learning-course-data-bucket))/
-          s3://((ol-ocw-studio-app-bucket))/courses/
-          --include '*'
-          --exclude "extracts/*"
-          --exclude "*.html"
-          --exclude "*.html.json"
-          --exclude "*_master.json"
-          --exclude "*parsed.json"
+  - get: sync-static-resources
+    trigger: true
```
blarghmatey, line 11: This should be a `put` step so that it can take advantage of the bucket-to-bucket sync.

mschuchard: Good call; and I will assume we do not need to `get` the `source_bucket` prior to the `put`.

blarghmatey:
Looks good :+1:

Issue #565, opened 2022-01-18
ocw-www and other root sites should publish a git hash for doof to watch


gumaerc:
In the [build script](https://github.com/mitodl/salt-ops/blob/main/salt/apps/ocw/templates/webhook-publish.sh.jinja) that we've been using to publish the OCW Next website up until now, there is a step to run `npm run build:githash` to deposit the current commit hash into the output of the site at an expected path.  The Concourse based builds don't do this, so a step should be added to the concourse build to do so so that Doof can detect when an RC / release has been deployed.

PR #566, opened 2022-01-19
run npm run build:githash after building ocw webpack assets

gumaerc:
Closes https://github.com/mitodl/ol-infrastructure/issues/565

This PR adds the `npm run build:githash` command to the theme assets pipeline definition so that the `hash.txt` file is generated alongside the theme assets and Doof will be able to read the current deployed commit from the Concourse generated OCW Next sites.


blarghmatey:
:+1:

PR #567, opened 2022-01-19
Feat/concourse perf updates

Ardiea:
Updated sops gpg fingerprints.
Updated vector / cloudwatch config on concourse servers.
Created configuration for new concourse ci environment.
Encrypted configuration for new concourse ci environment. 
Updated concourse pulumi code to dynamically load in iam definitions based on configuration settings. 
Change how the consul datacenter is determined. 
Allowed the target vpc to be defined in config rather than code.


blarghmatey:
:+1:

PR #569, opened 2022-01-20
Fixing sender email address and cookie name conflicts in edX deployments

blarghmatey:
This fixes two distinct and unrelated issues that we are seeing in our edX installations.

- When working across installations that share a subdomain (e.g. QA and production or residential staging and live) users are unable to stay logged into the corresponding edX installations. This is due to a conflict in the cookie names related to JWT sessions for e.g. the learner MFE. This adds environment scoping to all of the cookie names that might conflict across installations.
- When sending bulk emails through the instructor dashboard we are seeing various failure modes. The root of the issue is that the `BULK_EMAIL_DEFAULT_FROM_EMAIL` value needs to be overridden or else those bulk emails will be sent from `no-reply@example.com`. This adds an appropriate value for that setting to each of the configuration templates relating to our separate deployments.

### DIFF
--- a/pipelines/edx/mfe-frontend-app-learning-mitx-ci.yaml
+++ b/pipelines/edx/mfe-frontend-app-learning-mitx-ci.yaml
@@ -40,7 +40,7 @@ jobs:
       outputs:
       - name: mfe-app-learning/dist
       params:
-        ACCESS_TOKEN_COOKIE_NAME: edx-jwt-cookie-header-payload
+        ACCESS_TOKEN_COOKIE_NAME: mitx-ci-edx-jwt-cookie-header-payload
         BASE_URL: https://lms-ci.mitx.mit.edu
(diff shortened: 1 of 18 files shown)


pdpinch:
lgtm, although it's an ambitious change set for one PR. 

I think we should set `EMAIL_USE_COURSE_ID_FROM_FOR_BULK` to False as well, for all our open edx distributions. It defaults to False, but I think it will make more sense the next we look 

For reference, I think https://github.com/openedx/edx-platform/pull/29001 changed the bulk mail settings, and it was backported to maple shortly before release.

briangrossman:
Desired from email addresses for MITx Online and xPRO:

- MITx Online: mitxonline-support@mit.edu
- xPRO: support@xpro.mit.edu

pdpinch:
one more email address:
- Residential MITx: mitx-support@mit.edu

PR #574, opened 2022-01-24
Feature/p2p streaming support

Ardiea:
Fixed vector to send metrics to cloudwatch + grafana on concourse nodes.
Fixed dynamic IAM policy loading.
Found and flagged an issue with the infra iam policy which isn't needed yet.
Fixed p2p streaming for concourse workers.
Enhancmenet to allow Mike to use an included sops binary as well because he is special.
Fixed the github app id + secret for concourse ci login configs. Finished setting up concoruse ci. Fixed the security groups for p2p streaming.


src/bilder/images/concourse/templates/vector.yaml
```diff
@@ -19,28 +25,61 @@ transforms:
...
 
+  add_labels_to_metrics:
+    type: remap
+    inputs:
+    - '*metrics'
+    source: |
+      .tags.environment = "${ENVIRONMENT}"
+      .tags.application = "concourse"
+      .tags.service = "concourse"
+      .tags.hostname = "${HOSTNAME}"
```
blarghmatey, line 42: We don't need to add the hostname tag. That's one of the labels that contributes to our cardinality issues.

src/bilder/images/concourse/templates/vector.yaml
```diff
@@ -1,5 +1,11 @@
 ---
 sources:
+  vector_metrics:
```
blarghmatey, line 3: I've been debating whether it's worth capturing the Vector metrics. Happy to hear any thoughts that you have on the matter.

src/bilder/images/concourse/templates/vector.yaml
```diff
@@ -19,28 +25,61 @@ transforms:
...
+      password: ${GRAFANA_CLOUD_API_KEY}
+      user: "${GRAFANA_CLOUD_LOKI_API_USER}"
+    endpoint: https://logs-prod-us-central1.grafana.net
+    encoding:
+      codec: json
+    labels:
+      environment: ${ENVIRONMENT}
+      application: concourse
+      service: concourse
+      hostname: ${HOSTNAME}
```
blarghmatey, line 66: Let's skip on the hostname label for logs so that it doesn't contribute too much to explosion of log streams.

src/bilder/images/concourse/templates/vector.yaml
Ardiea: 👍

src/bilder/images/concourse/templates/vector.yaml
blarghmatey: Yeah, I think we just drop collection of those metrics from the agents. I think it would be helpful when running Vector as an aggregator, but having it on each host isn't terribly useful.

src/bilder/images/concourse/templates/vector.yaml
```diff
@@ -1,11 +1,5 @@
 ---
 sources:
-  vector_metrics:
-    type: internal_metrics
-    scrape_interval_secs: 30
-  host_metrics:
```
blarghmatey: We should keep the host metrics, that's useful to have for CPU, disk, memory, etc. Particularly for worker nodes so we can understand what their performance looks like.

blarghmatey:
:+1:

PR #577, opened 2022-01-27
Adjusting how DNS servers and DHCP are handled on servers with consul

Ardiea:
Adjusting how DNS servers and DHCP are handled to ensure that localhost is the first DNS server on any servers that require it (servers that use consul).


blarghmatey:
:+1:

blarghmatey (approved):

PR #585, opened 2022-02-01
Excluding cgroups from host metrics collection

blarghmatey:
The cgroups metrics are contributing significantly to our overall metrics cardinality in Grafana. This also sets the scrape interval to 60 seconds.


Ardiea (approved):
👍

PR #586, opened 2022-02-01
config: update value for COURSE_ABOUT_VISIBILITY_PERMISSION to see_exists

arslanashraf7:
## Related Ticket:
None, We were seeing 404 error responses from `/api/course_home/course_metadata/<course_id>)` for every student who was enrolled and tried to access the course. This was happening based on access permissions where `COURSE_ABOUT_VISIBILITY_PERMISSION=staff` and any student when tried to load the course would see a 404.

## What this PR does?
- Sets the value for `COURSE_ABOUT_VISIBILITY_PERMISSION=see_exists` so that enrolled students could load the course.


arslanashraf7:
@blarghmatey We are looking to test this change on `mitx-qa-draft.mitx.mit.edu`. I didn't really see any specific file for defining the configurations for `mitx-qa-draft`. I guess it inherits the config from some other place. But I didn't know where and how is it happening. 

So, can you suggest if this change is correct or this needs to happen somewhere else if we want to only use this config for `mitx-qa-draft.mitx.mit.edu`.

arslanashraf7:
FYI, What we are looking to do is:

```
1- COURSES_INVITE_ONLY=True
2- COURSE_CATALOG_VISIBILITY_PERMISSION='staff`
3- COURSE_ABOUT_VISIBILITY_PERMISSION="see_exists" 
```
So as for the file, I changed in this PR only one value needed to be changed. It's possible if we have to make changes in some other file there might be more values to change. But in the end we want the above three configurations to have values as mentioned above.

blarghmatey:
This is the right file to change. As far as only applying to QA it is possible, but shouldn't be necessary as the deployment will go to CI -> QA before being promoted to production.

blarghmatey (approved):
:+1:

arslanashraf7:
@blarghmatey Thanks, And do I need to worry about the pre-commit failure? It mentions about detecting secrets in my commit 😨

arslanashraf7:
I saw some previous commits in master mentioning the same failure, And knowing that this PR doesn't really mess with any secrets. I'm going to merge this.

blarghmatey:
Yeah, you can ignore that failure. We need to do some cleanup on the pre-commit configs to make that less noisy.

PR #587, opened 2022-02-01
Feature/concourse worker classes

Ardiea:
Created the ability to define classes of concourse workers, each with their own tags and teams. These workers can also have over-ridden AWS tags for cost reporting. Each class of worker can be customized with different instance types + storage.

### DIFF
--- a/src/ol_infrastructure/applications/concourse/__main__.py
+++ b/src/ol_infrastructure/applications/concourse/__main__.py
@@ -10,6 +10,7 @@ import importlib
 import json
 import textwrap
 from pathlib import Path
+from typing import List
 
 import pulumi_vault as vault
 import yaml
@@ -39,6 +40,10 @@ from ol_infrastructure.lib.ol_types import AWSBase
 from ol_infrastructure.lib.pulumi_helper import parse_stack
 from ol_infrastructure.lib.stack_defaults import defaults
 from ol_infrastructure.lib.vault import setup_vault_provider
+
+##################################
+##    Setup + Config Retrival   ##
+####################################
 
 if Config("vault_server").get("env_namespace"):
     setup_vault_provider()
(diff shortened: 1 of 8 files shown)


src/ol_infrastructure/applications/concourse/__main__.py
```diff
@@ -40,6 +42,62 @@
 from ol_infrastructure.lib.stack_defaults import defaults
 from ol_infrastructure.lib.vault import setup_vault_provider
 
+
+def build_worker_user_data(concourse_team, concourse_tags, consul_dc):
```
blarghmatey, line 46: It would be good to add some type information to the function signature. Something like:
`def build_worker_user_data(concourse_team: str, concourse_tags: List[str], consul_dc: str) -> str:`

src/ol_infrastructure/applications/concourse/__main__.py
```diff
@@ -533,102 +606,151 @@
...
+        path="/ol-applications/concourse/profile/",
+    )
+
+    concourse_worker_instance_profiles.append(concourse_worker_instance_profile)
+
+    # Create EC2 resources
+    build_worker_user_data_partial = partial(
+        build_worker_user_data,
+        worker_def.get("concourse_team"),
+        worker_def.get("concourse_tags") or [],
```
blarghmatey, line 655: Because `worker_def` is a dictionary you can pass the `[]` as the default value if there is no key at `"concourse_tags"`. So, `worker_def.get("concourse_tags", [])`

blarghmatey:
Very minor comments on some stylistic things, otherwise looks great! :+1:

src/ol_infrastructure/applications/concourse/__main__.py
Ardiea: 👍

src/ol_infrastructure/applications/concourse/__main__.py
Ardiea: 👍

PR #603, opened 2022-02-08
Adding read access to the ol-ocw-studio-app-<env> buckets for open learning 

Ardiea:


blarghmatey:
Looks good to me :+1:

PR #606, opened 2022-02-10
Vector config centralization and tidying 

Ardiea:
Refactoring the code for vector configurations to have just one place where we have defined how to collect host metrics.
Tidy up other vector configurations.


blarghmatey:
Looks good :+1:

PR #609, opened 2022-02-11
Issue 599 install vector globally

Ardiea:
- Bringing edxapp pyinfra code for installing vector into line with other applications. Closes 599 
- Installing a base vector configuration on all concourse servers rather than just web servers. Closes 599 
- Adding vector configuration validation. Closes 608

### DIFF
--- a/src/bilder/components/vector/steps.py
+++ b/src/bilder/components/vector/steps.py
@@ -78,6 +78,18 @@ def configure_vector(vector_config: VectorConfig, state=None, host=None):
         )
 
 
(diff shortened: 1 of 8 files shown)


blarghmatey:
This all looks great. Minor design/semantic question: Should we just add the validate step to the configure function rather than making it a separate function call?

Ardiea:
> This all looks great. Minor design/semantic question: Should we just add the validate step to the configure function rather than making it a separate function call?

Sure, I will come back to it today. I don't have strong opinions one way or the other.

PR #618, opened 2022-02-14
Stalled Concourse worker mitigation

Ardiea:
Added ELB, target group, and LB Listeners to concourse workers that will allow us to utilize automatic instance health management via AWS.

Tested in CI as follows:

Deployed.
Connected to a CI worker.
`systemctl stop concourse`
Wait.
A short while later the instance is reported as 'Unhealthy' in the TG on the AWS console. 
Shortly there after, the instance is terminated and replaced.


src/ol_infrastructure/applications/concourse/__main__.py
```diff
@@ -708,13 +709,62 @@ def build_worker_user_data(
...
+            timeout=20,
+            unhealthy_threshold=5,
+        ),
+        name=f"concourse-worker-tg-{worker_class_name[:3]}-{stack_info.env_suffix[:2]}"[
+            :TARGET_GROUP_NAME_MAX_LENGTH
+        ],
+        tags=aws_config.merged_tags(worker_def["aws_tags"]),
+    )
+
+    worker_lb_alb_listener = lb.Listener(
```
blarghmatey, line 748: Do we need a listener configured since we won't actually be routing any traffic to the workers?

blarghmatey: Maybe if we just add the workers as a target group to the existing Concourse ALB and just don't use a listener since we don't actually care about (or necessarily want?) traffic making it to the workers.

blarghmatey: Roger that. Maybe we can scope the rules for the listener to only accept traffic on the 10. address range so that we don't get arbitrary traffic hitting those endpoints?

Ardiea: So, you raise interesting questions and I did science. 

1. A normal elb address from the web instances:
Screen Shot 2022-02-14 at 12 47 31 PM

2. One of the worker elb addresses:
Screen Shot 2022-02-14 at 12 45 53 PM

So the web ELB instance returns _public_ addresses of concourse-web nodes. The 'internal' (line 718 in this file), returns the private IP addresses from the CI ops vpc. It is interesting that it can be resolved from anywhere (assuming you could guess the whole name), but it resolves to an address that is not route-able outside of the VPC.

blarghmatey (approved):

blarghmatey:
:+1:

PR #619, opened 2022-02-15
Bundle concourse resource_types 

Ardiea:
Set up processes for bundling some of our frequently used resource_types onto our concourse worker AMIs when we build them.


pipelines/ocw/docker-image-pipeline.yml
```diff
@@ -1,37 +0,0 @@
----
-resources:
-- name: ol-infrastructure
-  type: git
-  source:
-    uri: https://github.com/mitodl/ol-infrastructure.git
-    branch: ((ol-infrastructure-git-ref))
-    paths:
-    - dockerfiles/ocw/node-hugo/Dockerfile
```
blarghmatey, line 9: So, this Docker image is actually not a resource, just an image that has all of the pieces needed to be able to run the course builds and asset compilation.

pipelines/resources/rclone.yaml
```diff
@@ -24,3 +24,52 @@ jobs:
...
+          echo \ \ \"type\": \"rclone\", >> resource_metadata.json;
+          echo \ \ \"version\": \"${version}\", >> resource_metadata.json;
+          echo \ \ \"privileged\": false, >> resource_metadata.json;
+          echo \ \ \"version_history\": false >> resource_metadata.json;
+          echo } >> resource_metadata.json;
+          echo "resource_metadata.json contents:"
+          cat resource_metadata.json;
+          tar czf ./rclone.tgz rootfs.tgz resource_metadata.json;
+        path: sh
+  - task: packaged-resource-upload
```
blarghmatey, line 62: Might make sense to run this as a `put` step with the s3-sync resource so that the output can be tracked as a resource.

blarghmatey:
Given the similarity of the pipelines, maybe we should write it as an `across` step to iterate over whichever resources we want packaged?

src/bilder/components/concourse/models.py
```diff
@@ -1464,6 +1464,14 @@ def concourse_env(self) -> Dict[str, str]:
 class ConcourseWorkerConfig(ConcourseBaseConfig):
     _node_type: str = "worker"
     user: str = "root"
+    ## TODO: MAD 02162022 Correct this deploy_directory reference.
+    ## This 'works' but it isn't correct. It will always reference the default value for
+    ## deploy_directory, rather than any that is overridden in the actual instantiation, whereever
+    ## that may be. Be aware.
+    resource_types_directory: Path = (
+        ConcourseBaseConfig().deploy_directory / "resource-types"
```
blarghmatey, line 1472: The problem with this approach is that we want to make this relative to the configured deploy directory. With this it will always use the default value from the parent class, rather than whatever customized location is passed into the instance creation of this class.

blarghmatey:
Another option would be just creating a pipeline template that we populate from a vars file when creating it

pipelines/resources/build-concourse-resources.yaml
```diff
@@ -0,0 +1,86 @@
...
+- name: build-and-publish-((resource-name))
+  public: false
+  serial: true
+  plan:
+  - get: ((resource-name))-git-resource
+    trigger: true
+  - put: ((resource-name))-docker-resource
+    params:
+      build: ((resource-name))-git-resource
+      additional_tags: ((resource-name))-git-resource/tag
```
blarghmatey, line 51: Should we also tag it with the git hash from the resource repo?

pipelines/resources/README.md
```diff
@@ -0,0 +1,17 @@
+This is a unified process for building MIT-ODL custom resource types, publishing them to docker-hub, and also staging them to a s3 bucket so they can be bundled into AMI bakes of concourse workers. In order to include a resource type in this process you will need to ensure a few things.
+
+1. The resource type must have its own publicly accessible git repository.
+2. Within that git repository, it the resource type must utilize a `tag` file containing a semantic version number for the git resource. This tag file must be within the root of the repository.
+3. There must be a Dockerfile in the root of the repository and it must build cleanly without any extra steps beyond `docker build .`.
+
+To add a new resource type to this process, create a new `vars_{resource-name}.yaml` file and populate the following three variables.
+
+1. `git-repo` : Aforementioned publicly accessible git repository matching the above prerequisites.
+2. `docker-repo` : the full docker image repository name, something like `mitodl\{resource-name}`
```
blarghmatey, line 10: Minor typo. Should be `/` rather than `\`

pipelines/resources/README.md
```diff
@@ -0,0 +1,17 @@
+This is a unified process for building MIT-ODL custom resource types, publishing them to docker-hub, and also staging them to a s3 bucket so they can be bundled into AMI bakes of concourse workers. In order to include a resource type in this process you will need to ensure a few things.
+
+1. The resource type must have its own publicly accessible git repository.
+2. Within that git repository, it the resource type must utilize a `tag` file containing a semantic version number for the git resource. This tag file must be within the root of the repository.
```
blarghmatey, line 4: Typo: "Within that git repository, ~it~ the resource"

pipelines/resources/build-concourse-resources.yaml
```diff
@@ -0,0 +1,86 @@
...
+  - task: 'tar-resource-files'
+    config:
+      image_resource:
+        name: ''
+        source:
+          repository: node
+          tag: 14-buster-slim
```
blarghmatey, line 58: What's the purpose for using a Node container?

src/bilder/components/concourse/steps.py
```diff
@@ -192,6 +192,47 @@ def _manage_worker_node_keys(
         )
 
 
+@deploy("Pull down pre-bundled resource types")
+def _install_resource_types(
+    concourse_config: ConcourseWorkerConfig, sudo=True, host=None, state=None
+):
+    if concourse_config.additional_resource_types:
```
blarghmatey, line 199: You can remove the `if` conditional and change the `for` loop to be:
`for resource in concourse_config.additional_resource_types or []:`

blarghmatey:
:+1:

PR #625, opened 2022-02-17
Modify the ocw-studio event rule for sns notifications

mbertrand:
Adds an additional filter to the event rule for ocw-studio MediaConvert SNS notifications.

Should not be merged/applied until after https://github.com/mitodl/ocw-studio/pull/1018 and https://github.com/mitodl/salt-ops/pull/1501


blarghmatey:
:+1:

Ardiea (approved):

PR #638, opened 2022-03-01
Added vault ami pipeline

mitangelo:


pipelines/infrastructure/concourse_ami_new_release.yaml
```diff
@@ -0,0 +1,102 @@
...
+  source:
+    owner: concourse
+    repository: concourse
+    release: true
+- icon: github
+  name: packer_templates
+  source:
+    branch: main
+    paths:
+    - src/bilder/images/
```
blarghmatey, line 39: Now that we're building individual images this should be updated to read:
```
path:
  - src/bilder/images/concourse/
  - src/bilder/images/packer.pkr.hcl
  - src/bilder/images/variables.pkr.hcl
  - src/bilder/images/config.pkr.hcl
```

Similarly for the other pipelines. Otherwise it will trigger a rebuild of the AMI for changes unrelated to the actual image.

pipelines/infrastructure/consul_ami_new_release.yaml
```diff
@@ -0,0 +1,77 @@
...
+resources:
+- check_every: 24h
+  name: consul
+  source:
+    project: consul
+  type: hashicorp-release
+- icon: github
+  name: packer_templates
+  source:
+    branch: add_app_version_to_ami
```
blarghmatey, line 28: The branch will need to be updated to point to `main`

blarghmatey (commented):
Just a couple of minor tweaks, otherwise looks great!

pipelines/infrastructure/concourse_ami_new_release.yaml
mitangelo: got it...thanks...Put specific paths for the specific app.

blarghmatey (approved):
:+1:

Issue #639, opened 2022-03-02
Create OpenSearch clusters for Micromasters et al


Ardiea:
Start looking at the pulumi configuration needed to build out OpenSearch clusters for use by the applications. 

Related : issue https://github.com/mitodl/micromasters/pull/5166

Need to migrate the existing read-only-rest configurations to OpenSearch RBAC or w/e AWS does (their version of xpack).

PR #647, opened 2022-03-07
Initial commit of opensearch enhancmenets

Ardiea:
- Adds a conditional feature to create a 'public' opensearch cluster, rather than VPC cluster. This feature flag will also restrict access to the cluster to only two predefined roles that are then exported for consumption by application stacks. 
- Adds a conditional feature to enabled TLS intercommunication between nodes.
- Parameterizes the consul service name for opensearch clusters.
- Created stack configurations for opensearch in CI, QA, and Productions
- Created stack configurations for consul.apps.CI


src/ol_infrastructure/infrastructure/aws/opensearch/__main__.py
```diff
@@ -111,27 +189,25 @@
...
     "aws-opensearch-consul-node",
     address=search_domain.endpoint,
     opts=pulumi.ResourceOptions(provider=consul_provider),
 )
-
 opensearch_service = consul.Service(
     "aws-opensearch-consul-service",
     node=opensearch_node.name,
-    name="elasticsearch",
+    name="opensearch",
```
Ardiea, line 200: I should make this a configuration setting `opensearch:consul_service_name` and then update the old stacks to set it to `elasticsearch`

src/ol_infrastructure/infrastructure/aws/opensearch/__main__.py
```diff
@@ -1,14 +1,14 @@
-import json
 from pathlib import Path
 
 import pulumi
 import pulumi_aws as aws
+import pulumi_aws.iam as iam
```
blarghmatey, line 5: Here I would either just do `aws.iam.<function/class>` or do `from pulumi_aws import iam`

src/ol_infrastructure/infrastructure/aws/opensearch/__main__.py
```diff
@@ -39,27 +39,65 @@
...
-    tags=aws_config.merged_tags({"Name": f"{environment_name}-opensearch"}),
-    description="Grant access to the OpenSearch service domain",
-    egress=default_egress_args,
-    ingress=[
+##########
+# CREATE #
+##########
+
+# Networking
+if search_config.get_bool("public_web"):
```
blarghmatey, line 47: I would initialize a variable of `is_public_web` or similar with the value of the config fetch so that it's not necessary to repeat that everywhere.

src/ol_infrastructure/infrastructure/aws/opensearch/__main__.py
```diff
@@ -39,27 +39,65 @@
...
+        "node_to_node_encryption"
+    ] = aws.elasticsearch.DomainNodeToNodeEncryptionArgs(
+        enabled=True,
+    )
+
 search_domain = aws.elasticsearch.Domain(
     "opensearch-domain-cluster",
     domain_name=f"opensearch-{environment_name}"[:SEARCH_DOMAIN_NAME_MAX_LENGTH],
-    elasticsearch_version=search_config.get("engine_version") or "7.10",
+    elasticsearch_version=search_config.get("engine_version") or 7.10,
```
blarghmatey, line 100: Will this work as a decimal value, or does it need to remain as a string?

src/ol_infrastructure/infrastructure/aws/opensearch/__main__.py
blarghmatey: I think that's a good practice in general, that way it allows for multiple clusters to co-exist in the same Consul datacenter without conflicting.

src/ol_infrastructure/infrastructure/aws/opensearch/__main__.py
Ardiea: oof good catch.

src/ol_infrastructure/infrastructure/consul/Pulumi.infrastructure.consul.apps.CI.yaml
```diff
@@ -0,0 +1,8 @@
+---
+secretsprovider: awskms://alias/PulumiSecrets
+encryptedkey: AQICAHionUR8LBW1ALuVC0rCH3AE2oQIfGMCx3XmpDH9HjM2LQEN4tntCwkkcLiIv0Bk3NmXAAAAfjB8BgkqhkiG9w0BBwagbzBtAgEAMGgGCSqGSIb3DQEHATAeBglghkgBZQMEAS4wEQQMNMvx1SE6r73gkyn9AgEQgDsqXU6PuIbyKPtePL++niXfKCQuThehdYM2SirIsMnTneJtFVgq1L3kSs1MyBOgSooYKbA2g8LViLurSg==
+config:
+  aws:region: us-east-1
+  consul:instance_count: "5"
```
blarghmatey, line 6: Should only need 3 instances for CI

blarghmatey:
The only issue I see currently is that on the existing stacks (e.g. mitxonline.QA) the IAM linter is not happy. Once that's resolved I think this is good to merge.

Ardiea:
> The only issue I see currently is that on the existing stacks (e.g. mitxonline.QA) the IAM linter is not happy. Once that's resolved I think this is good to merge.

Done. Based on this:

https://docs.aws.amazon.com/opensearch-service/latest/developerguide/ac.html

Specifically,

> The trailing /* in the Resource element is significant and indicates that resource-based policies only apply to the domain's subresources, not the domain itself. In resource-based policies, the es:* action is equivalent to es:ESHttp*.

We were giving a bunch of irrelevant actions to use against domain _subresources_. Most recent commit addresses that.

blarghmatey (approved):
:+1:

Ardiea:
Closes #639

PR #650, opened 2022-03-10
Enable disk metric collection with vector

### LINKED ISSUE #649: Enable Disk Metrics via Vector
Enable Disk Metrics via Vector

This is the only way to get IOPS data which is a frequent trouble point. Need to figure out a filter to use to keep from collecting a bunch of useless series from pseudo devices.

Ardiea:
Updated the global vector host metric configuration file to collect disk metrics but only from root devices and not partitions.

Closes #649


blarghmatey:
:+1:

PR #666, opened 2022-03-14
Initial creation of cloud custodian pipelines

### LINKED ISSUE #496: Add a periodic task to clean old AMIs
With our shift to immutable deployments we are generating a huge number of AMIs. At any given time the majority of those are old and not useful, so we want to have a periodic task to delete any that aren't actively attached to an autoscale group or a deployed instance. A likely candidate for this is to create a Concourse pipeline that uses [cloud custodian](https://cloudcustodian.io/) to query for unused images and delete them. One thing to be cautious of as we start working on this project is to ensure that we only delete images that we are creating.

Ardiea:
Initial creation of cloud custodian pipelines that will clean up AMI related resources. 

Only in dry-run mode at the moment. Closes #496


src/ol_infrastructure/applications/concourse/iam_policies/infra.py
```diff
@@ -12,6 +12,8 @@
         {
             "Effect": "Allow",
             "Action": [
+                "autoscaling:DescribeAutoScalingGroups",
+                "autoscaling:DescribeLaunchConfigurations",
```
blarghmatey, line 16: We should probably split the permissions into different statements so we can track which permissions are needed for what purposes. It might get unwieldy, but it's probably worth a good college try.

pipelines/infrastructure/cloud_custodian_policies/sync_ec2_tags.yaml
```diff
@@ -0,0 +1,52 @@
+---
+policies:
```
blarghmatey, line 2: Probably worth creating a separate directory structure for cloud-custodian so that we can organize the policies in a way that is internally consistent to them, rather than tying them to the pipelines. If you have a different opinion on that I'm happy to be wrong.

pipelines/infrastructure/cloud_custodian_policies/sync_ec2_tags.yaml
```diff
@@ -0,0 +1,52 @@
...
+  resource: aws.ebs
+  description: |
+    Finds and propagates tags from ec2 instances found in the previous policy and adds those tags to the block devices on each instance.
+  filters:
+  - type: value
+    key: "VolumeId"
+    op: in
+    value_from:
+      # TODO, will need to figure out what the concourse path will be here?
+      # Include a comment on how to run it locally?
```
blarghmatey, line 24: Should we make these separate steps in the pipeline so that we can use the input/output? Or should this be a relative path in the assumption that the previous step will make the file in the local directory?

blarghmatey:
:+1:

PR #668, opened 2022-03-14
Long timeout for ocw-to-hugo pipeline

mbertrand:
Increases the timeout for a long s3 sync and fixes the RC install script for it.


blarghmatey:
:+1:

PR #680, opened 2022-03-21
Vector Proxy

Ardiea:
This will build and provision a vector_log_proxy service stack. Include building the AMI

Still to do:

- [x] documentation: how to setup a log drain
- [x] documentation: other?
- [x] Update concourse AMI building pipeline to build this AMI as well (The vector component doesn't support being passed a version number at this time so it is kind of moot. Need to open an enhancement to do that)
- [X] Author stacks for QA and production
- [x] Build stacks for QA and production\
- [X] setup log drains for heroku apps in CI, QA, and Production

### DIFF
--- a/src/ol_infrastructure/infrastructure/vector_log_proxy/__main__.py
+++ b/src/ol_infrastructure/infrastructure/vector_log_proxy/__main__.py
@@ -36,6 +36,8 @@ dns_stack = StackReference("infrastructure.aws.dns")
 consul_stack = StackReference(f"infrastructure.consul.operations.{stack_info.name}")
 mitodl_zone_id = dns_stack.require_output("odl_zone_id")
 
+env_name = f"{stack_info.env_prefix}-{stack_info.env_suffix}"
+
 target_vpc_name = (
     vector_log_proxy_config.get("target_vpc") or f"{stack_info.env_prefix}_vpc"
 )
@@ -45,10 +47,13 @@ VECTOR_LOG_PROXY_PORT = vector_log_proxy_config.get("listener_port") or 9000
 
 consul_security_groups = consul_stack.require_output("security_groups")
 aws_config = AWSBase(
-    tags={"OU": "operations", "Environment": f"operations-{stack_info.env_suffix}"}
+    tags={
+        "OU": vector_log_proxy_config.get("business_unit") or "operations",
+        "Environment": f"{env_name}",
+    }
 )
 aws_account = get_caller_identity()
-ops_vpc_id = target_vpc["id"]
+vpc_id = target_vpc["id"]
 vector_log_proxy_ami = ec2.get_ami(
     filters=[
         ec2.GetAmiFilterArgs(name="name", values=["vector_log_proxy-*"]),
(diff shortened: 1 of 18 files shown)


src/ol_infrastructure/infrastructure/vector_log_proxy/__main__.py
```diff
@@ -0,0 +1,345 @@
...
+target_vpc_name = (
+    vector_log_proxy_config.get("target_vpc") or f"{stack_info.env_prefix}_vpc"
+)
+target_vpc = network_stack.require_output(target_vpc_name)
+
+VECTOR_LOG_PROXY_PORT = vector_log_proxy_config.get("listener_port") or 9000
+
+consul_security_groups = consul_stack.require_output("security_groups")
+aws_config = AWSBase(
+    tags={"OU": "operations", "Environment": f"operations-{stack_info.env_suffix}"}
```
blarghmatey, line 48: If we're making this flexible to run in different VPCs then we should allow for stack config to override the OU and fall back to "operations"

blarghmatey: For the environment, we should probably just create that value from `f"{stack_info.env_prefix}-{stack_info.env_suffix}"`

src/ol_infrastructure/infrastructure/vector_log_proxy/__main__.py
```diff
@@ -0,0 +1,345 @@
...
+)
+aws_account = get_caller_identity()
+ops_vpc_id = target_vpc["id"]
```
blarghmatey, line 51: Since we're being agnostic to which VPC we should change this naming to just `vpc_id` or `target_vpc_id` or similar

src/ol_infrastructure/infrastructure/vector_log_proxy/__main__.py
```diff
@@ -0,0 +1,345 @@
...
+    filters=[
+        ec2.GetAmiFilterArgs(name="name", values=["vector_log_proxy-*"]),
+        ec2.GetAmiFilterArgs(name="virtualization-type", values=["hvm"]),
+        ec2.GetAmiFilterArgs(name="root-device-type", values=["ebs"]),
+    ],
+    most_recent=True,
+    owners=[aws_account.account_id],
+)
+
+vector_log_proxy_tag = f"vector-log-proxy-{stack_info.env_suffix}"
```
blarghmatey, line 62: Maybe `vector-server`? Since we can use it for more than just proxying logs. Also, should probably replace any usage of `env_suffix` with `env_name` constructed above from the prefix and suffix.

src/ol_infrastructure/infrastructure/vector_log_proxy/__main__.py
```diff
@@ -0,0 +1,345 @@
...
+    description="Launch template for deploying vector-log-proxy nodes",
+    iam_instance_profile=ec2.LaunchTemplateIamInstanceProfileArgs(
+        arn=vector_log_proxy_instance_profile.arn,
+    ),
+    image_id=vector_log_proxy_ami.id,
+    block_device_mappings=[
+        ec2.LaunchTemplateBlockDeviceMappingArgs(
+            device_name="/dev/xvda",
+            ebs=ec2.LaunchTemplateBlockDeviceMappingEbsArgs(
+                volume_size=vector_log_proxy_config.get_int("disk_size") or 8,
```
blarghmatey, line 238: Probably worth defaulting to ~25GB since 8 can pretty quickly get overwhelmed, particularly if we start spooling any logs to disk while batching up groups.

blarghmatey:
:+1: looks good to me

PR #689, opened 2022-03-22
Created backup buckets for ocw-content, both live and draft.

Ardiea:
Created backup buckets for ocw-content, both live and draft.


blarghmatey:
Looks good, we'll just want to confirm that the Concourse IAM policy allows access to the new buckets.

PR #698, opened 2022-03-25
Issue 617

Ardiea:
Moved the github_shared_secret out of pulumi config and into sops config.

Manually populated github_app_private_key to QA. secret-ocw-studio does not exist in CI.


src/ol_infrastructure/applications/ocw_studio/__main__.py
```diff
@@ -231,11 +233,15 @@
...
 
-vault_secrets = ocw_studio_config.require_secret_object("vault_secrets")
+vault_secrets = read_yaml_secrets(
+    Path(f"ocw_studio/ocw_studio.{stack_info.env_suffix}.yaml")
+)
 vault.generic.Secret(
     "ocw-studio-vault-secrets",
     path=ocw_studio_secrets.path.apply("{}/app-config".format),
-    data_json=vault_secrets.apply(json.dumps),
+    data_json=json.dumps(
```
blarghmatey, line 242: Ideally we'll have the structure in the YAML set to match exactly what we are expecting downstream in Concourse, the app, etc. That will let us just naively do something like `json.dumps(ocw_secrets)`

PR #699, opened 2022-03-26
ocw-content backup adjustments

Ardiea:
Updated the ocw-content-backup-* buckets to use versioning with a 90 day version expiration policy. Updated the backup jobs to run every 6 hours.


blarghmatey (approved):
:+1:

PR #705, opened 2022-03-28
Moving github_shared_secret to sops 

Ardiea:
Moved the github_shared_secret out of pulumi config and into sops config/

Closes 617


src/bridge/secrets/ocw_studio/ocw_studio.ci.yaml
```diff
@@ -0,0 +1,77 @@
+---
+github_shared_secret: ENC[AES256_GCM,data:Pdrn+lFOh9Rb53XpxIxzZvLpzNKIe6lbGXvT1Gbu3ZsnaKseUjqKHcqCA5l3RdHJahSBOWzvgGn5BWG/5El4r8GQ,iv:ug0p3Kgqc/p2+04rO8hZTjlE3MnbXS+/Awl/n1GFy14=,tag:NYOrJH9meTTCE9MamXHcLQ==,type:str]
```
blarghmatey, line 2: The data in Vault also contains a `github_app_private_key` which we either need to include in these SOPS files or determine with the devs whether it is still needed.

PR #708, opened 2022-03-28
Added instance refresh and metadata disable false

mitangelo:


pipelines/infrastructure/ol_infrastructure/instance_refresh_vault_ami_qa.yaml
```diff
@@ -0,0 +1,57 @@
...
+- name: vault-ami
+  type: ami
+  check_every: 1m
+  source:
+    region: us-east-1
+    filters:
+      owner-id: 610119931565
+      is-public: false
+      state: available
+      name: vault-server
```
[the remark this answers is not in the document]
mitangelo: I see there are 2 params. One is the name tag an then there is the ami name. Not so explicit in the docs to use ami-name and not name...

pipelines/infrastructure/ol_infrastructure/instance_refresh_concourse_ami_ci.yaml
```diff
@@ -0,0 +1,98 @@
...
+  - task: set-aws-creds
+    config:
+      platform: linux
+      image_resource:
+        type: registry-image
+        source:
+          repository: amazon/aws-cli
+          tag: 'latest'
+      params:
+        AWS_ROLE: arn:aws:iam::610119931565:role/ol-applications/concourse/role/concourse-instance-role-worker-infra-qa-4c8b313
```
blarghmatey, line 60: Do we need to hard-code this? Do we need to pass it in at all? It would be great if we can pull it dynamically from the metadata API via curl or something similar.

pipelines/infrastructure/ol_infrastructure/instance_refresh_consul_ami_ci.yaml
```diff
@@ -0,0 +1,85 @@
...
+    params:
+      env_pulumi:
+        AWS_SHARED_CREDENTIALS_FILE: aws_creds/credentials
+      action: update
+      env_os:
+        AWS_DEFAULT_REGION: us-east-1
+        PYTHONPATH: /usr/lib/:/tmp/build/put/pulumi_files/src/
+      project_name: ol-infrastructure-consul-server
+      source_dir: pulumi_files/src/ol_infrastructure/infrastructure/consul
+      stack_name: infrastructure.consul.operations.CI
```
blarghmatey, line 86: So, for Consul we'll actually need to have multiple `put` steps, with one for each environment where we are deploying Consul clusters to.

pipelines/infrastructure/ol_infrastructure/instance_refresh_concourse_ami_ci.yaml
```diff
@@ -0,0 +1,96 @@
...
+          $(aws sts assume-role \
+          --role-arn $AWS_ROLE \
+          --role-session-name MySessionName \
+          --query "Credentials.[AccessKeyId,SecretAccessKey,SessionToken]" \
+          --output text))
+          cd aws_creds
+          echo "[default]" > credentials
+          echo "aws_access_key_id=$AWS_ACCESS_KEY_ID" >> credentials
+          echo "aws_secret_access_key=$AWS_SECRET_ACCESS_KEY" >> credentials
+          echo "aws_session_token=$AWS_SESSION_TOKEN" >> credentials
```
blarghmatey, line 77: It's probably worth pulling this script into its own file in the repository so we can use it across all of our pipelines without having to copy/paste it each time. Then we can update these three pipelines to use the same script.

mitangelo: hmm is there a way how to include a script in concourse? include statement?

blarghmatey: You would just pass the path of the script in the repository as the parameter to the `run` argument of the task step. So, rather than in-lining the script as you did above with the `run` parameter being `sh` it would just be the path to the script as it exists in the repository.

pipelines/infrastructure/ol_infrastructure/instance_refresh_concourse_ami_ci.yaml
```diff
@@ -0,0 +1,83 @@
...
+    filters:
+      owner-id: 610119931565
+      is-public: false
+      state: available
+      name: concourse-web-*
+
+- icon: github
+  name: pulumi_files
+  source:
+    branch: issue_679
```
shaidar, line 38: ?

pipelines/infrastructure/ol_infrastructure/instance_refresh_concourse_ami_ci.yaml
```diff
@@ -0,0 +1,83 @@
...
+    - pulumi_files
+    get_params:
+      skip_implicit_get: true
+    params:
+      env_pulumi:
+        AWS_SHARED_CREDENTIALS_FILE: aws_creds/credentials
+      action: update
+      env_os:
+        AWS_DEFAULT_REGION: us-east-1
+        PYTHONPATH: /usr/lib/:/tmp/build/put/pulumi_files/src/
```
shaidar, line 80: That's a pretty awkward path

pipelines/infrastructure/ol_infrastructure/instance_refresh_vault_ami_ci.yaml
```diff
@@ -0,0 +1,72 @@
...
+        type: registry-image
+        source:
+          repository: amazon/aws-cli
+          tag: 'latest'
+      inputs:
+      - name: pulumi_files
+      outputs:
+      - name: aws_creds
+      run:
+        path: pulumi_files/pipelines/infrastructure/scripts/generate_aws_config_from_instance_profile.sh
```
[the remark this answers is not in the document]
blarghmatey: It only runs when the pipeline executes

pipelines/infrastructure/scripts/generate_aws_config_from_instance_profile.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/sh
```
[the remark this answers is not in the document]
blarghmatey: The file won't exist because this script runs as part of the pipeline, so every time it executes it will be a fresh environment.

shaidar:
Would be good to have a description to accompany the PR even if it's a copy of the title. Either way, 👍

PR #719, opened 2022-03-29
Created healthcheck.io alerting for when the job succeeeds. 

Ardiea:
If the job should not succeed for  more than 2(3?) runs it will generate an alert in healthchecks.io


blarghmatey:
:+1:

PR #722, opened 2022-03-30
Initial monitoring inventory

Ardiea:
## Description
Inventory of monitoring tools and what applications utilize them. 

## Motivation and Context
Documentation

## How Has This Been Tested?
Not applicable.

## Types of changes
- [X] Documentation

## Checklist:


docs/monitoring_inventory/inventory.md
```diff
@@ -0,0 +1,36 @@
...
+  - Purpose: Absence of alerting detection. For determining when other alerting tools may have failed or become unresponsive.
+- GrafanaCloud
+  - Purpose: Collecting and storing metric and log data from applications and infrastructure.
+  - Subcomponents:
+    - Grafana: Visualization and alerting on metric and log data.
+    - Cortex: Backend for storing metric data.
+    - Loki: Backend for storing log data
+- Pingdom
+  - Purpose: External synthetic and HTTP monitoring.
+
```
blarghmatey, line 15: Should we include Cloudwatch here?

Ardiea: Yes? Probably? It seems to have limited utility compared to basically any other source of the same types of information but in some circumstances it is our _only_ source of information, like RDS. 

And then in the weird world of EBS and IOPS there doesn't seem to be an accurate relationship between what an on-system utility like vector reports vs. what is recorded in CloudWatch. The value in CloudWatch seems to be more important/relevant if we're looking at something related to storage performance and running out of credits or whatever.

shaidar: Do we include newrelic on the Heroku apps?

Ardiea: Do we still use new relic? If we're sunsetting it I don't want to spend too much time finding where it is and is not used.

PR #740, opened 2022-04-07
Public Opensearch clusters -> Finegrained access control

Ardiea:
Public Opensearch clusters -> Finegrained access control

## Description
* Cleaned up the code for opensearch pulumi builds and made things a little more consistent
* Added a bootstrapping script to configure users and roles in public+secured opensearch clusters.
* Enabled encryption at rest on secured opensearch clusters.
* Removed most of the IAM config from opensearch cluster buildouts.
* Changed the public opensearch clusters to use finegrained security rather than resource or user based policies.
* Created documentation on the bootstrap script. 

## Motivation and Context
Role-based or user-based access control on opensearch clusters would not be a drop-in replacement for the elasticsearch clusters we are currently using. Finegrained security is a drop-in replacement that we currently need.

## How Has This Been Tested?
* Destroyed and built new apps clusters for CI/QA/Prod.
* Verified expected changes on existing not-apps-clusters (should just be a minor tweak to the security policy of existant, non-apps-clusters) 

## Types of changes

- [X] Breaking change (fix or feature that would cause existing functionality to change)
  - Maybe, but I don't think so. I destroyed all the user/role based clusters. Those would definitely be broken if not. 

## Checklist:

------------------------------
One follow up is to revisit the role definitions and define indices and make them more restrictive to just the permissions the two users need. They are pretty broad at the moment.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -89,3 +89,4 @@ repos:
     additional_dependencies:
     - pydantic
     - types-PyYAML
+    - types-requests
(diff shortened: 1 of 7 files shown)


blarghmatey:
:+1:

PR #744, opened 2022-04-11
build: override frontend-lib-special-exam to use 1.16.1

arslanashraf7:
## Description
After our required change was released in https://github.com/edx/frontend-lib-special-exams/releases/tag/v1.16.1 we need to use this package to reflect this change in our system. So, we need to override it in our system. 

## Motivation and Context
- We will only have to override this package if we don't want to upgrade our system to the latest open edX because with the latest upgrade this change would already be part of the edX installation itself. So, if we only want to pick up this change while remaining on the old edX installation we need to override it.
https://github.com/mitodl/mitxonline/issues/374
- In addition, I created this PR in advance but we need to check at which point our production instance is. I know our QA instance is on the latest but keeping in mind that we might not update our installation we need to check if this override would be feasible or not.
- If the difference b/w 1.16.1 of `frontend-lib-special-exam` and the installed version in production is big enough we might have to do the change in our fork by creating a branch from that specific tag, upload the package to npm registry, and then install that in this override.

## How Has This Been Tested?
Testing instructions https://github.com/mitodl/mitxonline/issues/374

## Types of changes

## Checklist:

### DIFF
--- a/pipelines/edx/mfe-frontend-app-learning-mitxonline-production.yml
+++ b/pipelines/edx/mfe-frontend-app-learning-mitxonline-production.yml
@@ -79,6 +79,7 @@ jobs:
           npm install
           npm install @edx/frontend-component-footer@npm:@mitodl/frontend-component-footer-mitol@latest --legacy-peer-deps
           npm install @edx/frontend-component-header@npm:@mitodl/frontend-component-header-mitol@latest --legacy-peer-deps
+          npm install @edx/frontend-lib-special-exams@1.16.1 --legacy-peer-deps
           NODE_ENV=production npm run build
   - put: mfe-app-learning-bucket
     params:


arslanashraf7:
Added a `Do not merge` label because this change is only valid in two cases that still need to be clarified:
1) We are not upgrading our prod edX instance to the latest as of now.
2) The difference between 1.16.1 and the currently installed version of `frontend-lib-special-exam` is not broad enough to break anything.

So, we need to check at which point are we in the production instance of edX.

shaidar:
@blarghmatey Is this PR even needed since the pipeline pulls from the `master` branch of the app-learning repo which in turn [installs the latest](https://github.com/openedx/frontend-app-learning/blob/master/package.json#L35) `frontend-lib-special-exam` lib?

arslanashraf7:
@shaidar The upstream PR was merged and since we pick from master this PR isn't necessary. There were only two specific cases where it was needed(mentioned above) but they don't seem related to our usecase now.

arslanashraf7:
Closing this PR in favor of above comment.

PR #751, opened 2022-04-11
Support removal of edX Python packages

blarghmatey:
## Description
There are some cases where we want to remove a Python package that is installed as part
of the edX build. This adds a data structure that provides a list of package names to be
uninstalled in the image customization stage of the build process.

## Motivation and Context

Addresses the conversation in https://github.com/mitodl/mitxonline/issues/343

## How Has This Been Tested?
This will be tested by building a new image and deploying it to a pre-production environment.

## Types of changes

## Checklist:

### DIFF
--- a/src/bilder/images/edxapp/deploy.py
+++ b/src/bilder/images/edxapp/deploy.py
@@ -98,8 +98,16 @@ git_auto_export()
 # Install additional Python dependencies for use with edxapp
 pip.packages(
     name="Install additional edX dependencies",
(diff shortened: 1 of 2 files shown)


shaidar:
Looks good to me for now 👍 . Ideally would be nicer to not have something be installed in the first place, but based on our slack discussion, that would require modifying the [`requirements.txt`](https://github.com/openedx/edx-platform/blob/master/requirements/edx/base.txt#L462) file

PR #755, opened 2022-04-14
Pin named releases to specific envs by stack

blarghmatey:
## Description
As we test out new named releases we need to maintain a way for deploying the previous
release to QA and production environments in the event of a security patch or
bugfix. This adds an AMI tag and associated stack configuration to filter the AMI that
is deployed per environment/stack. This lets us build e.g. Nutmeg AMIs to deploy to
residential CI while preventing accidental deployment to QA and production.

## Motivation and Context

This is necessary to allow us to flexibly propagate a given named release to different
environment levels without blocking us from deploying prior releases to production while
we are still in testing.

## How Has This Been Tested?

This will be tested in the CI environment once new AMIs have been built with the
requisite tags.

## Types of changes

## Checklist:

### DIFF
--- a/src/bilder/images/edxapp/custom_install.pkr.hcl
+++ b/src/bilder/images/edxapp/custom_install.pkr.hcl
@@ -53,6 +53,7 @@ source "amazon-ebs" "edxapp" {
     OU      = "${var.business_unit}"
     app     = "${local.app_name}"
     purpose = "${local.app_name}-${var.node_type}"
+    edxapp_release = "${var.edx_platform_version}"
   }
   # Base all builds off of the most recent Ubuntu 20.04 image built by the Canonical organization.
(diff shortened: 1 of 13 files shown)


shaidar:
I see the new `edxapp_release` tag and the AMI filtering on that tag, but two things i’m not clear on:
- Where is the AMI build using that tag to clone the correct branch?
- `edxapp_release` is assigned the value of `edx_platform_version` but i don’t see where that’s being set from where it’s defined in the stack configs?

shaidar:
`edx_platform_version` is defined in the packer vars file, example [here](https://github.com/mitodl/ol-infrastructure/blob/main/src/bilder/images/edxapp/packer_vars/nutmeg.pkrvars.hcl)

Issue #824, opened 2022-05-18
Create a standard EC2 base image for standardizing deployments


blarghmatey:
In order to reduce the overhead involved in installing and configuring existing and net-new services we want to establish a standard approach for how to get software into a running environment. While Kubernetes offers the promise of scale and flexibility, it also brings with it a requirement that we adapt to it for the best outcome. It also has shortcomings with how secrets are managed that makes it non-trivial to properly secure that information.

The middle ground that we are taking to reduce the variance and custom logic required to get an application or service onto an EC2 instance and integrating with our standard tooling for dynamic configuration, log and metric collection, and deployment is to use docker-compose as the deployment artifact that we build into the images. We already have PyInfra logic for each of the agents that we need to support such an installation, but we do not yet have the container management piece developed.

To Do:
- [x] Create a PyInfra deployment file that installs and provides the core configuration for each of:
  - [x] Vault Agent
  - [x] Consul Agent
  - [x] consul-template
  - [x] Vector agent
- [x] Create a PyInfra component definition for installing and configuring docker and managing `docker compose` as a service
  - [x] This will involve accepting one or multiple YAML files with `compose` specifications that will need to be executed
- [x] The base EC2 box should make it easy to add N configuration templates that will be rendered by consul-template and loaded into a standard directory for mounting into the compose runtime

blarghmatey:
Completed with #858

PR #826, opened 2022-05-19
Adding docker install

mitangelo:


mitangelo:
#fixes https://github.com/mitodl/ol-infrastructure/issues/824

blarghmatey:
pre-commit.ci autofix

mitangelo:
@blarghmatey tested and working....

src/bilder/components/docker/steps.py
```diff
@@ -0,0 +1,78 @@
...
+from pyinfra.facts.server import Command, LinuxDistribution
+from pyinfra.operations import apt, files
+
+
+def _apt_install():
+
+    apt.packages(
+        name="Ensure Docker CE prerequisites are present",
+        present=True,
+        packages=["curl", "gnupg-agent", "software-properties-common", "lsb-release"],
```
blarghmatey, line 18: You can probably remove `lsb-release` since we're not relying on it any more.

src/bilder/components/docker/steps.py
```diff
@@ -0,0 +1,78 @@
...
+
+    apt.packages(
+        name="Install Docker via apt",
+        packages="docker-ce",
+        update=add_apt_repo.changed,
+    )
+
+
+@deploy("Deploy Docker")
+def deploy_docker(config=None):
```
blarghmatey, line 56: We generally try to take advantage of type hints so it's probably worth adding `config: Optional[Dict[str, Any]] = None` rather than using the `isinstance` check further down.

src/bilder/components/docker/steps.py
```diff
@@ -0,0 +1,78 @@
...
+        raise DeployError(
+            ("Apt not found, " "pyinfra-docker cannot provision this machine!"),
+        )
+
+    config_file = config
+
+    if isinstance(config, dict):
+        config_hash = make_hash(config)
+
+        config_file = StringIO(json.dumps(config, indent=4))
```
blarghmatey, line 70: This is a useful approach. Another way that we've done it in other places is with `tempfile`s. An example is visible [here](https://github.com/mitodl/ol-infrastructure/blob/main/src/bilder/components/hashicorp/consul/steps.py#L33-L40). Not a requirement, just for reference.

src/bilder/components/docker/steps.py
mitangelo: :+1:

PR #837, opened 2022-05-24
autoscaling component resource

### LINKED ISSUE #825: Simplify the interface for launching EC2 images in autoscale groups
In order to reduce the overhead involved in getting new workloads deployed we are building a standard EC2 base image that will run applications via docker compose specifications. This will simplify the work of building new server images, so we will need a way to make it faster and easier to run those images in AWS.

The standard pattern that we need for running an application or service is:
- [ ] Create an IAM policy for access to any AWS resources (typically S3 buckets and EC2 discovery)
- [x] Create an RDS host that can be accessed by the deployed service
- [ ] Create appropriate security groups and attach them to the launch template definitions
- [x] Create one or more auto-scale groups for running web and/or worker nodes
- [x] Create a load-balancer for attaching to the web-server autoscale group

This leads to a lot of boilerplate code that could be extracted into a set of functions with standard default arguments that reduces the time needed to deploy a new service. We would like either [`ComponentResource'](https://www.pulumi.com/docs/intro/concepts/resources/components/) definitions or functions that generate the relevant resources to address the above requirements.

Ardiea:
Created a component resource to represent the standard pattern we use for autoscaling resources is AWS. Refactored tika code to use the new component resource. 

Addresses part of #825 

## Motivation and Context
This will provide a simplified interface for creating autoscaling groups (and associated resources) that follow our standard pattern. 

## How Has This Been Tested?

This is really v1 and it has only been tested against Tika. We should open tasks to refactor existing environments to use these new resources and reduce some of the boiler plate that is scattered around. 

With further expansion, we may find shortcoming to these resources as written and may need to revisit them. 

## Types of changes

## Checklist:


src/ol_infrastructure/components/aws/auto_scale_group.py
```diff
@@ -0,0 +1,345 @@
...
+from ol_infrastructure.lib.ol_types import AWSBase
+
+
+class BlockDeviceMapping(BaseModel):
+    """Container for describing a block device mapping for an EC2 instance/launch template"""
+
+    delete_on_termination: bool = True
+    device_name: str = "/dev/xvda"
+    volume_size: PositiveInt = PositiveInt(25)
+    volume_type: DiskTypes = DiskTypes.legacy_ssd
```
blarghmatey, line 45: Why are we defaulting to the gp2 disk type?

src/ol_infrastructure/components/aws/auto_scale_group.py
```diff
@@ -0,0 +1,345 @@
...
+    """Component to create an autoscaling group with defaults as well as managed associated resources"""
+
+    lb_dns_name: str
+
+    def __init__(
+        self,
+        asg_config: OLAutoScaleGroupConfig,
+        lt_config: OLLaunchTemplateConfig,
+        tg_config: OLTargetGroupConfig,
+        lb_config: OLLoadBalancerConfig,
```
blarghmatey, line 187: Should we make this `Optional[OLLoadBalancerConfig]` and then key off of that to determine whether to create the load balancer? That way we can easily use the same component for ASGs that do and don't want an LB

src/ol_infrastructure/components/aws/auto_scale_group.py
```diff
@@ -0,0 +1,345 @@
...
+            opts,
+        )
+
+        # Shared attributes
+        resource_options = pulumi.ResourceOptions(parent=self).merge(opts)  # type: ignore
+        resource_name_prefix = asg_config.asg_name + "-"
+
+        pulumi.error(msg=f"{lb_config.subnets.__type__}", resource=lb_config.subnets)
+        # Create Load Balancer
+        load_balancer = LoadBalancer(
```
blarghmatey, line 204: For each attribute that is getting created it's useful to assign it as e.g. `self.load_balancer`, etc. That way when you create an instance of the component resource it's easy to access the child resources and their relevant attributes.

src/ol_infrastructure/components/aws/auto_scale_group.py
Ardiea: That is a good question to which I do not have an answer. Is gp3 more appropriate for a default?

src/ol_infrastructure/components/aws/auto_scale_group.py
Ardiea: That makes sense. I didn't know there was a use case that omits the LB.

src/ol_infrastructure/components/aws/auto_scale_group.py
blarghmatey: Yeah, we should default to gp3 because it's generally cheaper with better performance than gp2. It's the newest generation of generic SSD so we should bias toward that unless there's a reason that we can't.

blarghmatey:
:+1:

blarghmatey:
closes #825

PR #842, opened 2022-05-27
build: pin packages(rapid-response-reports, canvas-integration) upgraded for Nutmeg

arslanashraf7:
This pins the packages for `ol-openedx-rapid-response-reports`, `ol-openedx-canvas-integration` for production and QA. 

## Description
We did some migration with our plugins to support serving assets through the plugins. That change will only work for Nutmeg branch. So, for the instances Prod/QA which are not using Nutmeg release yet so we need to pin the old packages to be used there to keep them from breaking. 
I couldn't see the QA configurations and the staging seems to be the closes to the QA so I changed in that.

## Motivation and Context
Inspired by https://github.com/mitodl/edx-platform/pull/297#issuecomment-1138713468

## How Has This Been Tested?
No testing is needed as such,  since the packages pinned were the ones published when we deployed Prod/QA and were being used already.

## Types of changes

## Checklist:

### DIFF
--- a/src/bilder/images/edxapp/group_data/all.py
+++ b/src/bilder/images/edxapp/group_data/all.py
@@ -55,8 +55,8 @@ edx_plugins_added = {
         "ol-openedx-sentry",
         "ol-openedx-course-export",
         "rapid-response-xblock==0.6.0",
-        "ol-openedx-canvas-integration",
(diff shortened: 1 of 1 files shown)


blarghmatey:
:+1:

PR #845, opened 2022-05-27
edx database migrations on boot 

### LINKED ISSUE #420: Run database migrations as edX instances are released
In order for us to more easily automate deployment of edX images we need a way to reliably run any database migrations automatically. For this to work well we need to meet a few requirements:
- [ ] Only run migrations on one instance at a time (we can use [Consul locking](https://www.consul.io/commands/lock) to help here)
- [ ] Notify on migration failures (post to slack, ping opsgenie, etc.)
- [ ] Terminate an in-process instance refresh?

Ardiea:
## Description
Setup a one shot systemd service to run database migrations on boot. Will use a consul lock to ensure that only one database migration runs at a time. Closes #420 

## Motivation and Context
https://github.com/mitodl/ol-infrastructure/issues/420 

## How Has This Been Tested?
Built a new AMI, updated the XPro CI environment, verified that the migration service ran on both workers and web nodes. 

Trusting the `consul lock` functionality to work as advertised but can verify no overlapping timestamps in grafana (aside from the messages saying it couldn't acquire the lock and blocking).

## Types of changes

## Checklist:

### DIFF
--- a/src/bilder/images/edxapp/files/migrations.service
+++ b/src/bilder/images/edxapp/files/migrations.service
@@ -1,0 +1,17 @@
+[Unit]
+Description="Obtain an exclusive lock and run database migrations at startup."
+Requires=network-online.target
+After=consul-template.service
+
+[Service]
+Type=oneshot
+ExecStart=/usr/local/bin/consul lock -timeout=1h -verbose lsm_migrations /edx/bin/edxapp-migrate-lms
+ExecStart=/usr/local/bin/consul lock -timeout=1h -verbose cms_migrations /edx/bin/edxapp-migrate-cms
+RemainAfterExit=True
+StandardOutput=journal
+StandardError=journal
+SyslogIdentifier=migrations
+Restart=no
+
+[Install]
+WantedBy=multi-user.target
--- a/src/bilder/images/edxapp/files/migrations.service
+++ b/src/bilder/images/edxapp/files/migrations.service
@@ -5,8 +5,8 @@ After=consul-template.service
 
 [Service]
 Type=oneshot
-ExecStart=/usr/local/bin/consul lock -timeout=1h -verbose lsm_migrations /edx/bin/edxapp-migrate-lms
-ExecStart=/usr/local/bin/consul lock -timeout=1h -verbose cms_migrations /edx/bin/edxapp-migrate-cms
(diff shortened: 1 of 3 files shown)


src/bilder/images/edxapp/files/migrations.service
```diff
@@ -0,0 +1,17 @@
+[Unit]
+Description="Obtain an exclusive lock and run database migrations at startup."
+Requires=network-online.target
+After=consul-template.service
+
+[Service]
+Type=oneshot
+ExecStart=/usr/local/bin/consul lock -timeout=1h -verbose lsm_migrations /edx/bin/edxapp-migrate-lms
```
blarghmatey, line 8: nit: typo. `lsm_migrations` -> `lms_migrations`

src/bilder/images/edxapp/files/migrations.service
```diff
@@ -0,0 +1,17 @@
...
+ExecStart=/usr/local/bin/consul lock -timeout=1h -verbose cms_migrations /edx/bin/edxapp-migrate-cms
```
blarghmatey, line 9: It would be best to actually serialize the LMS and CMS migrations because they have a number of shared models, so if you run all of the LMS migrations first there may not actually be any CMS migrations that need to run.

src/bilder/images/edxapp/files/migrations.service
blarghmatey: It might be better to set a lock timeout longer than 1 hour to account for the potential of migrations to take a long time on large tables. It _should_ be plenty in most cases but we should err on the side of a longer expiry.

Ardiea: 👍 LSM is an acronym from a previous life...

src/bilder/images/edxapp/files/migrations.service
Ardiea: This SO page seems to suggest these will run serially: https://stackoverflow.com/questions/48195340/systemd-with-multiple-execstart as does this redhat blog post: https://www.redhat.com/sysadmin/systemd-oneshot-service under "Chaining commands in a oneshot service" but I can't actually find a statement in the systemd documentation specifying it.

Ardiea: Another SO post supporting that statement: https://stackoverflow.com/questions/39032100/what-is-the-difference-between-systemds-oneshot-and-simple-service-types

PR #866, opened 2022-06-07
Airbyte Initial Draft 

### LINKED ISSUE #828: Support operation of our own installation of Airbyte
As a data platform owner I would like to be able to run Airbyte on our own infrastructure so that raw data doesn't leave our premises or get accessed by third-party systems.

We are incorporating Airbyte as a component of our data platform infrastructure to support Extract and Load functionality for data integration. We currently have an account with their cloud offering but we would like to ultimately run it ourselves. There is an existing option for managing it on EC2 via docker-compose, but we will likely need to modify it slightly to avoid using the built-in database container.

- [x] Create an AMI that incorporates a modified docker-compose file for running Airbyte as a service
  - Additional instructions are available at https://docs.airbyte.com/deploying-airbyte/on-aws-ec2/
- [x] Proxy the Airbyte web service with a process that will manage user authentication (basic auth is acceptable for the v1 implementation)
- [x] Deploy the Airbyte AMI into an AWS autoscale group to ensure uptime and simplify deployments of the image
- [x] Create an RDS database for Airbyte to use as persistent storage of its state and configurations using Pulumi

Ardiea:
## Description
This creates an airbyte AMI as well as a QA environment in pulumi. 

## Motivation and Context
Closes #828 

## How Has This Been Tested?
Not ready yet. 

## Types of changes

- [X] Bug fix (non-breaking change which fixes an issue)

## Checklist:

- [X] My change requires a change to the documentation.
- [x] I have updated the documentation accordingly.

## TODO Round 1
- [x] Include the .env file after it is scrubbed for 'secrets'.
- [x] Setup the RDS instance with pulumi 
- [x] Setup credentials from vault for the rds instance
- [x] Take those credentials and append them to the end of the RDS instance. 
- [x] Figure out the default route for pomerium, currently goes to /.pomerium, very frustrating. 
- [X] Documentation on all the silly little secrets as well as how to setup the github side of the auth. 
  - [x] Move application config out of my user into shared user / bot user. 
- [x] Replace the SENTRY_DSN in .env.tmpl with our own (but store it in vault!) 
- [x] Move the database passwords out of the sops config and into the stack config.
- [x] Make `     storage=airbyte_config.get("db_capacity"), ` an optional attribute by setting a default. 
- [x] NGINX or caddy in front of pomerium 
  - [x] to allow differently-authenticated access from dagster.
- [x] Move secrets to `secret-airbyte` (convo below). 
- [x] Fix stack name due to orthographic processing deficit.


src/ol_infrastructure/applications/airbyte/Pulumi.airbyte.applications.QA.yaml
```diff
@@ -0,0 +1,15 @@
+---
```
blarghmatey, line 1: The naming convention for the stacks is `applications.airbyte.QA` so that everything is namespaced under `applications`, `infrastructure`, etc.

src/ol_infrastructure/applications/airbyte/Pulumi.yaml
```diff
@@ -0,0 +1,6 @@
+---
+name: ol-infrastructure-tika-server
```
blarghmatey, line 2: Copy :spaghetti:

src/ol_infrastructure/applications/airbyte/__main__.py
```diff
@@ -0,0 +1,268 @@
...
+    bound_account_ids=[aws_account.account_id],
+    bound_vpc_ids=[vpc_id],
+    token_policies=[airbyte_server_vault_policy.name],
+)
+
+##################################
+#     Network Access Control     #
+##################################
+# Create security group
+airbyte_server_security_group = ec2.SecurityGroup(
```
blarghmatey, line 133: Unless we're using the sec group as an identifier for other ingress rules there's not much need for it, since every VPC has a web server group built into it https://github.com/mitodl/ol-infrastructure/blob/main/src/ol_infrastructure/infrastructure/aws/network/security_groups.py#L18-L55

src/ol_infrastructure/applications/airbyte/__main__.py
```diff
@@ -0,0 +1,268 @@
...
+    subnets=target_vpc["subnet_ids"],
+    security_groups=[airbyte_server_security_group],
+    tags=aws_config.merged_tags({"Name": airbyte_server_tag}),
+)
+
+tg_config = OLTargetGroupConfig(
+    vpc_id=vpc_id,
+    health_check_interval=60,
+    health_check_matcher="200-399",
+    health_check_path="/.pomerium",
```
blarghmatey, line 163: Is there an Airbyte specific path that we can use for the healthcheck to know if the underlying service is happy?

src/ol_infrastructure/applications/airbyte/__main__.py
```diff
@@ -0,0 +1,268 @@
...
+
+auto_scale_config = airbyte_config.get_object("auto_scale") or {
+    "desired": 1,
+    "min": 1,
+    "max": 2,
+}
+asg_config = OLAutoScaleGroupConfig(
+    asg_name=f"airbyte-server-{env_name}",
+    aws_config=aws_config,
+    desired_size=auto_scale_config["desired"] or 2,
```
blarghmatey, line 245: You're already providing a default object, so this is redundant

src/ol_infrastructure/applications/airbyte/airbyte_server_policy.hcl
```diff
@@ -0,0 +1,35 @@
...
+
+path "secret-data/data-ci/airbyte/*" {
+  capabilities = ["read"]
+}
+
+path "secret-data/data-qa/airbyte/*" {
+  capabilities = ["read"]
+}
+
+path "secret-data/data-production/airbyte" {
```
blarghmatey, line 13: We have separate Vault clusters for each environment level, so there's no need to include the env-specific paths in the policy definition. We should aim to keep these as flat as possible.

blarghmatey: Maybe even mount a backend specific to Airbyte so that we can scope things easily (e.g. `secret-airbyte`)

src/ol_infrastructure/applications/airbyte/Pulumi.airbyte.applications.QA.yaml
```diff
@@ -0,0 +1,15 @@
...
+secretsprovider: awskms://alias/infrastructure-secrets-qa
+encryptedkey: AQICAHi7xhTkB8tf1ObyPMxDhODJja4Mn4jyIo32zZVlOiZPFgFqyheTEwQhQwt19Cm5Z2wyAAAAfjB8BgkqhkiG9w0BBwagbzBtAgEAMGgGCSqGSIb3DQEHATAeBglghkgBZQMEAS4wEQQMUHRADb5VO/Osd+QMAgEQgDthoUItjofWZZROiCvTVCFgPdVM1f3tWDvxoezDTWseixH0X/RXXDnIDnOXDiC9dHp/ZRoXORovMifXGg==
+config:
+  airbyte:business_unit: operations
+  airbyte:target_vpc: data_vpc
+  airbyte:web_host_domain: airbyte-qa.odl.mit.edu
+  aws:region: us-east-1
+  consul:address: https://consul-data-qa.odl.mit.edu
+  tika:auto_scale:
```
blarghmatey, line 10: Copy :spaghetti:

src/bilder/images/airbyte/airbyte.pkr.hcl
```diff
@@ -0,0 +1,92 @@
...
+  snapshot_tags = {
+    Name           = "${local.app_name}-${var.node_type}-ami"
+    OU             = "${var.business_unit}"
+    app            = "${local.app_name}"
+    purpose        = "${local.app_name}-${var.node_type}"
+  }
+  # Base all builds off of the most recent docker_baseline_ami built by us, based of Debian 11
+  source_ami_filter {
+    filters = {
+      name                = "docker_baseline_ami-${var.node_type}-*"
```
blarghmatey, line 53: Is this the motivation for creating a new Packer template?

src/bilder/images/airbyte/deploy.py
```diff
@@ -0,0 +1,134 @@
...
+)
+vault = Vault(
+    version=VERSIONS["vault"],
+    configuration={Path("vault.json"): vault_config},
+)
+consul = Consul(version=VERSIONS["consul"], configuration=consul_configuration)
+consul_template_config = ConsulTemplate()
+
+hashicorp_products = [vault, consul]
+install_hashicorp_products(hashicorp_products)
```
blarghmatey, line 121: If you're building on top of the docker-compose base image then this step shouldn't be necessary

src/bilder/components/pomerium/files/config.yaml.j2
```diff
@@ -0,0 +1,41 @@
...
+idp_service_account: "{{ .Data.idp_service_account }}"
+
+# https://github.com/pomerium/pomerium/blob/main/scripts/generate_self_signed_signing_key.sh
+signing_key: "{{ .Data.signing_key }}"
+
+# Generate 256 bit random keys  e.g. `head -c32 /dev/urandom | base64`
+cookie_secret: "{{ .Data.cookie_secret }}"
+
+routes:
+- from: "{{ .Data.authenticate_server_url }}/"
```
blarghmatey, line 31: Is there a way to set up a non-authenticated route that we can use for communication from Dagster to Airbyte?

Ardiea: I think we would probably need to put another layer in front of things to accomplish that. Which, tbh, is probably the appropriate thing to do anyways. 
```
Internet -> Nginx/Caddy -> (User) -> Pomerium -> Github -> Pomerium -> Airbyte
                  \->(Special URL or Path) -> (Dagster + AuthToken)---/^
```

I stand and accuse you of scope creep. Unless that is written down in the requirements that I never actually read. :-)

src/bilder/images/airbyte/deploy.py
Ardiea: I believe you're correct. I was getting pretty confused at one point with this and what happens in the base AMI and what happens here. I think lines 124:125 do the needful I was actually looking for.

src/bilder/images/airbyte/airbyte.pkr.hcl
Ardiea: Yes.

src/ol_infrastructure/applications/airbyte/__main__.py
Ardiea: Not that pomerium is going to let us talk to without authenticating. We will need to do nginx/caddy in front to sneak around for specific paths. Probably a round 2 worthy task.

src/ol_infrastructure/applications/airbyte/airbyte_server_policy.hcl
Ardiea: The first thing is addressed, the second comment not yet.

src/bilder/components/pomerium/files/config.yaml.j2
Ardiea: ok

src/bilder/images/docker_baseline_ami/deploy.py
```diff
@@ -85,16 +80,10 @@
 configure_vector(vector_config)
 vector_service(vector_config)
 
-for product in hashicorp_products:
-    configure_hashicorp_product(product)
```
blarghmatey, line 89: Curious if there's utility in keeping this here and just removing the step that sets up the services?

src/ol_infrastructure/applications/airbyte/__main__.py
```diff
@@ -147,6 +157,118 @@
...
+    keys=[
+        consul.KeysKeyArgs(path=f"airbyte/database-host", value=db_address),
+        consul.KeysKeyArgs(path=f"airbyte/database-port", value=db_port),
+        consul.KeysKeyArgs(path=f"airbyte/database-name", value=db_name),
+        consul.KeysKeyArgs(
+            path=f"airbyte/database-connection-string", value=connection_string
+        ),
+    ],
+    opts=consul_provider,
+)
```
blarghmatey, line 270: Nothing to do for this PR, but this whole section seems to be a good candidate for wrapping in a function or component resource

src/bilder/images/airbyte/templates/.env.tmpl
```diff
@@ -0,0 +1,100 @@
...
+JOB_MAIN_CONTAINER_MEMORY_REQUEST=
+JOB_MAIN_CONTAINER_MEMORY_LIMIT=
+
+
+### LOGGING/MONITORING/TRACKING ###
+TRACKING_STRATEGY=segment
+# Although not present as an env var, expected by Log4J configuration.
+LOG_LEVEL=INFO
+# Although not present as an env var, helps Airbyte track job healthiness.
+SENTRY_DSN="https://d4b03de0c4574c78999b8d58e55243dc@o1009025.ingest.sentry.io/6102835"
```
blarghmatey, line 68: Is this a valid Sentry DSN for us, or is this something from the upstream config? Might be worth using our own Sentry and moving the DSN into Vault config so it doesn't get spammed.

src/bridge/secrets/data/airbyte/rds.production.yaml
```diff
@@ -0,0 +1,77 @@
+---
+password: ENC[AES256_GCM,data:OQD2ZcC0TraHwt+1s7G9Gh89fCDYSOPZZbkkBcrp9pLqr/5QoFEoPFvpeLlJhq8I,iv:og23WTNucTdU7GM//N60pZfPPX3JhL+NcZTGAjwb1pE=,tag:qeXZSAkmpRy2/1m5JjkxqQ==,type:str]
```
blarghmatey, line 2: Since this is a simple secret that is only valid for the context of Pulumi it's actually better to just use the Stack for encrypting and managing this value. You can use the `--secret` flag to config set to manage the encryption https://www.pulumi.com/docs/reference/cli/pulumi_config_set/

src/ol_infrastructure/applications/airbyte/Pulumi.airbyte.applications.QA.yaml
```diff
@@ -2,14 +2,16 @@
 secretsprovider: awskms://alias/infrastructure-secrets-qa
 encryptedkey: AQICAHi7xhTkB8tf1ObyPMxDhODJja4Mn4jyIo32zZVlOiZPFgFqyheTEwQhQwt19Cm5Z2wyAAAAfjB8BgkqhkiG9w0BBwagbzBtAgEAMGgGCSqGSIb3DQEHATAeBglghkgBZQMEAS4wEQQMUHRADb5VO/Osd+QMAgEQgDthoUItjofWZZROiCvTVCFgPdVM1f3tWDvxoezDTWseixH0X/RXXDnIDnOXDiC9dHp/ZRoXORovMifXGg==
 config:
+  airbyte:auth_domain: airbyte-qa-auth.odl.mit.edu
```
blarghmatey, line 5: Maybe `auth.airbyte-qa.odl.mit.edu`? Doesn't actually matter, just a stylistic nit

src/ol_infrastructure/applications/airbyte/__main__.py
blarghmatey: It's also being used as access control to the DB which is reason enough to keep it.

src/ol_infrastructure/applications/airbyte/__main__.py
```diff
@@ -147,6 +157,118 @@
...
+
+#########################
+#     Database Setup    #
+#########################
+rds_defaults = defaults(stack_info)["rds"]
+rds_defaults["instance_size"] = (
+    airbyte_config.get("db_instance_size") or rds_defaults["instance_size"]
+)
+rds_password = read_yaml_secrets(
```
blarghmatey, line 189: If you move the secret from Sops to the stack config then this would change to `airbyte_config.require_secret("db_password")` or similar

src/ol_infrastructure/applications/airbyte/__main__.py
```diff
@@ -147,6 +157,118 @@
...
+    Path(f"data/airbyte/rds.{stack_info.env_suffix}.yaml")
+)["password"]
+
+airbyte_db_config = OLPostgresDBConfig(
+    instance_name=f"airbyte-db-{stack_info.env_suffix}",
+    password=rds_password,
+    storage=airbyte_config.get("db_capacity"),
```
blarghmatey, line 196: Is this an optional or required attribute? If you want it to fail if it's not set you can use `airbyte_config.require`

src/bilder/components/pomerium/files/config.yaml.j2
blarghmatey: :rofl:

src/bilder/images/docker_baseline_ami/deploy.py
Ardiea: I don't know. I would need to do a more methodical experiment with it. I'm a little too scattered at the moment. Intuitively, it feels like invoking once with one config object and then again later with a totally different config object is just asking for trouble but I don't have much to support that statement yet. I will make a note to look into it further.

src/bilder/images/airbyte/templates/.env.tmpl
Ardiea: I honestly didn't even notice it. We should replace it with our own, I agree.

src/ol_infrastructure/applications/airbyte/Pulumi.airbyte.applications.QA.yaml
Ardiea: Will that work with our wildcard?

blarghmatey: Good point, I had read it as `airbyte-qa.auth` (note the dot, not dash). Go ahead and keep as is

src/bilder/images/airbyte/templates/.env.tmpl
Ardiea: I set this up but it doesn't seem to feed any data into the project. Prob not worth spending much time on at this moment.

src/ol_infrastructure/applications/airbyte/__main__.py
```diff
@@ -186,14 +186,13 @@
...
-    Path(f"data/airbyte/rds.{stack_info.env_suffix}.yaml")
-)["password"]
+
+rds_password = airbyte_config.require("rds_password")
 
 airbyte_db_config = OLPostgresDBConfig(
     instance_name=f"airbyte-db-{stack_info.env_suffix}",
     password=rds_password,
-    storage=airbyte_config.get("db_capacity"),
+    storage=airbyte_config.get("db_capacity") or "334",
```
blarghmatey, line 195: Probably worth moving that `334` into the magic numbers module with either a descriptive name or a comment to describe where that number comes from. Probably also worth just adding that into the defaults dictionary.

src/ol_infrastructure/applications/airbyte/__main__.py
odlbot: Moved this item into the phase two issue in the data milestone https://github.com/mitodl/ol-infrastructure/issues/899

Ardiea: Moved this item to the phase two issue in the data milestone https://github.com/mitodl/ol-infrastructure/issues/899

src/bilder/images/airbyte/deploy.py
```diff
@@ -0,0 +1,243 @@
...
+    Path("00-default.json"): ConsulConfig(
+        addresses=ConsulAddresses(dns="127.0.0.1", http="127.0.0.1"),
+        advertise_addr='{{ GetInterfaceIP "ens5" }}',
+        services=[
+            ConsulService(
+                name="airbyte",
+                port=DEFAULT_HTTPS_PORT,
+                check=ConsulServiceTCPCheck(
+                    name="airbyte-https",
+                    tcp="localhost:443",
```
blarghmatey, line 158: Are the other services that discover this node going to be communicating via HTTPS on the local network? If so, will they need a different TLS certificate to support access at a `.consul` TLD?

src/bilder/components/pomerium/templates/compose-pomerium.yaml.j2
```diff
@@ -0,0 +1,24 @@
+version: "3"
```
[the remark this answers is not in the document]
Ardiea: Yeah I think we should probably pair nginx + pomerium. I'm sure other situations will arise where we need similar bypasses.

src/bilder/images/airbyte/deploy.py
```diff
@@ -0,0 +1,243 @@
...
+    VaultTemplate(
+        contents='{{ with secret "secret-operations/global/odl_wildcard_cert" }}'
+        "{{ printf .Data.value }}{{ end }}",
+        destination=Path(f"{pomerium_config.certificate_file}"),
+    ),
+]
+consul_configuration = {
+    Path("00-default.json"): ConsulConfig(
+        addresses=ConsulAddresses(dns="127.0.0.1", http="127.0.0.1"),
+        advertise_addr='{{ GetInterfaceIP "ens5" }}',
```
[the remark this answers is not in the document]
Ardiea: Possibly? I'm not sure what the wider implications would be so I'm reluctant to do that without some larger testing with non-docker use cases. This is needed because we're running docker and consul gets confused with all the extra network interfaces that docker introduces (I guess?). 

It seems like the ubuntu images we use also use 'ens5' as the default interfaces, too, so it is something that can be looked into.

src/ol_infrastructure/applications/airbyte/airbyte_server_policy.hcl
```diff
@@ -0,0 +1,23 @@
+path "secret-operations/global/odl_wildcard_cert" {
+  capabilities = ["read"]
+}
+
+path "postgres-airbyte/creds/app" {
+  capabilities = ["read"]
+}
+
+path "postgres-airbyte/creds/admin" {
```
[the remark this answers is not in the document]
Ardiea: Yeah, I'm honestly not entirely clear on it myself but it has something to do with the 'temporal' service that runs along side airbyte. It wants to create a database(s). I didn't dig too deep into it but I suspect we can either bootstrap what it wants prior to starting airbyte the first time or we can adjust the approle to give it the specific powers it wants and nothing more. 

The same 'temporal' required turning off the 'forcessl' setting. It doesn't use the DATABASE_CONNECTION_STRING (or whatever the env var is called), so it doesn't get the two required parameters and refuses to speak with the DB because of it.

src/bilder/images/airbyte/deploy.py
```diff
@@ -75,47 +75,47 @@
 
 env_template_file = Path("/etc/consul-template/.env.tmpl")
 nginx_pomerium_conf_template_file = Path(
```
blarghmatey, line 77: This is already a `Path` object so it's not necessary to wrap it in the `Path` constructor

src/bilder/images/airbyte/deploy.py
Ardiea: Okay, so expose a :80 listener?

blarghmatey: Yeah, it's worth allowing Nginx to accept traffic over port 80 for that host

src/bilder/components/pomerium/models.py
```diff
@@ -0,0 +1,34 @@
+from pathlib import Path
+
+from bilder.lib.linux_helpers import DOCKER_COMPOSE_DIRECTORY
+from bilder.lib.model_helpers import OLBaseSettings
+
+
+class PomeriumConfig(OLBaseSettings):
+
+    configuration_directory: Path = Path("/etc/pomerium")
+    configuration_file: Path = Path(str(configuration_directory) + "/config.yaml")
```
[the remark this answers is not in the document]
Ardiea: This breaks the interpolation of the docker-compose file for pomerium/nginx.

```
==> amazon-ebs.airbyte: pyinfra.api.exceptions.OperationError: Error in template: /Users/michaeldavidson/code/ol-infrastructure/src/bilder/components/pomerium/templates/compose-pomerium.yaml.j2 (L10): 'bilder.components.pomerium.models.PomeriumConfig object' has no attribute 'certificate_file'
==> amazon-ebs.airbyte: ...
```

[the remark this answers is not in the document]
Ardiea: This whole situation could be cleaned up though. 👍

pipelines/infrastructure/ami_builds/airbyte_server_ami_new_releaase.yaml
```diff
@@ -0,0 +1,66 @@
+---
+##
+# This one does not check for new upstream releases. Will need to manually update
+# the python code in src/bilder when a tika upgrade is needed.
+#
+# Application docker compose file is configured to use the :latest, therefore to upgrade,
```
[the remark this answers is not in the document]
Ardiea: Added it to the phase two issue.

blarghmatey (approved):
:+1: let's get this merged :slightly_smiling_face:

Issue #163, opened 2021-05-20
Create Concourse pipeline for building and publishing edX Forum container


blarghmatey:
In order to simplify deployment and management of the edX Forum application we would like to run it in a container. To make this possible we need to develop a Dockerfile and associated Concourse pipeline to build the container and publish it to a container registry.

blarghmatey:
We can crib from the Dockerfile in Tutor for forum [here](https://github.com/overhangio/tutor/blob/master/tutor/templates/build/forum/Dockerfile) to figure out a working solution for our deployment.

mitangelo:
WIP....

PR #870, opened 2022-06-08
Added Concourse Pipeline for docker baseline ami and docker container

mitangelo:
Added 2 concourse pipelines. One to build the docker baseline ami and the other to build the docker image for forum.

## Types of changes

## Checklist:


dockerfiles/openedx-forum/Dockerfile
```diff
@@ -0,0 +1,50 @@
+FROM docker.io/ruby:2.5.7-slim-stretch
+MAINTAINER Overhang.IO <contact@overhang.io>
```
blarghmatey, line 2: We should change this line since we're not part of the Overhang organization

pipelines/open-edx/docker_baseline_ami.yaml
```diff
@@ -0,0 +1,182 @@
...
+    owner: moby
+    repository: moby
+    branch: master
+    release: true
+
+- check_every: 24h
+  name: vector-release
+  type: github-release
+  source:
+    owner: moby
```
blarghmatey, line 66: This looks like copy :spaghetti: :slightly_smiling_face:

pipelines/open-edx/forum_docker_build_pipeline.yaml
```diff
@@ -0,0 +1,61 @@
...
+resource_types:
+- name: github-release
+  type: registry-image
+  source:
+    repository: concourse/github-release-resource
+
+resources:
+- check_every: 24h
+  name: forum-release
+  type: github-release
```
blarghmatey, line 11: This won't be a github-release resource type because that repo doesn't generally use the releases, instead we'll need to use the `git` resource type and trigger when there is a new commit on a given branch.

blarghmatey (commented):
Looking good! Just a few comments.

blarghmatey:
This adds build automation for the docker-compose image built for #824 as well as building the Forum image for #163

pipelines/open-edx/forum_docker_build_pipeline.yaml
mitangelo: actually i tested this with https://github.com/openedx/cs_comments_service/releases

src/bilder/components/docker/docker_compose_steps.py
```diff
@@ -1,15 +1,21 @@
+import os
 from pathlib import Path
 from typing import Any, Dict, Optional
 
 from pyinfra.api.deploy import deploy
 from pyinfra.operations import files, server
 
+try:
+    DOCKER_COMPOSE_VERSION = os.environ["DOCKER_COMPOSE_VERSION"]
```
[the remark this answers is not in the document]
mitangelo: I did this so i dont need to hardcode the compose version. is it better to give  a default? i mean eventually that version will be outdated?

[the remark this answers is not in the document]
blarghmatey: I think it's ok to have a concrete version, just to make it easier to run it while testing/validating without having to set it every time. By putting it in `bridge.versions` it keeps all versions in a single place so they're easy to periodically update.

blarghmatey:
:+1:

PR #881, opened 2022-06-10
refactor: Use scoped security groups instead of CIDR for DB access

blarghmatey:
Updated the database security groups to grant access based on security group membership
for Vault and data integration services rather than blanket CIDR groups to help reduce
the scope of access.

## Description

This updates some of our access control definitions to be scoped based on service roles rather than just network addresses.

- adds a pair of security groups to the data VPC for the integration and orchestration services so that they can be used for scoping access to various services
- uses the Vault security group for granting access to that service for credential creation

## Motivation and Context

This is an enhancement that has been in the making for a while, but the necessary pieces are now in place to make it a reality.

## How Has This Been Tested?

This has not been tested yet, but is copying a pattern that we have used previously with success.

## Types of changes

## Checklist:


src/ol_infrastructure/applications/edxapp/__main__.py
```diff
@@ -378,13 +379,16 @@
     description="Access from Edxapp instances to the associated MariaDB database",
```
shaidar, line 379: Is this description accurate?

PR #895, opened 2022-06-13
version: Update OCW redirects dict to use production values

blarghmatey:
The OCW production live dictionary of redirects was manually edited after we took the
initial dump of its contents. This reconciles the current values with what we have in
source control so that we can use code as the source of truth going forward.

## Description

This updates the redirect dict for OCW sites to use the values as they are in production currently.

## Motivation and Context

Move to using this code as the source of truth for the OCW Fastly service

## How Has This Been Tested?

The values were pulled directly from production via the Fastly API

## Types of changes

## Checklist:


shaidar:
👍

PR #915, opened 2022-06-22
Updated 404 page

Wassaf-Shahzad:
## Description

Updated OCW site 404 page.

## How Has This Been Tested?

- Opened 404 

## Types of changes

### DIFF
--- a/src/ol_infrastructure/applications/ocw_site/404.html
+++ b/src/ol_infrastructure/applications/ocw_site/404.html
@@ -1,352 +1,148 @@
-<!doctype html>
-<html lang="en">
+<!DOCTYPE html>
+<html lang="{{ $.Site.Language.Lang }}">
   <head>
-
-
-<link href="/static/css/www.93a2f.css" rel="stylesheet">
-<link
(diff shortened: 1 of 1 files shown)


src/ol_infrastructure/applications/ocw_site/404.html
```diff
@@ -1,352 +1,148 @@
-<!doctype html>
-<html lang="en">
+<!DOCTYPE html>
+<html lang="{{ $.Site.Language.Lang }}">
```
blarghmatey, line 2: Is this a Hugo template tag? It doesn't seem like this will be rendered properly when loaded via Fastly

blarghmatey (commented):
It seems like this might be a template fragment that is expected to be embedded into another template? I'm just wondering if there are other elements (like the footer) that are expected to be in the document?

src/ol_infrastructure/applications/ocw_site/404.html
Wassaf-Shahzad: Sorry this escaped my go through. Fixed

blarghmatey:
:+1:

PR #952, opened 2022-07-15
Translate ami and deploy pipeline for Concourse to Python

blarghmatey:
Translate the Concourse pipelines for deploying Concourse to Python

## Description

This translates the pipeline definition for building the Concourse AMIs and then
deploying them with Pulumi into our Python DSL. This is a first step to abstracting
these operations into reusable methods to speed up pipeline development.

## Motivation and Context

Being able to automate all of our infrastructure workflows via Concourse is a useful means of reducing toil related to ensuring code changes and system updates are performed routinely. This is cumbersome when having to maintain the pipelines as YAML code. This adds the beginning steps of having easily reusable pipeline definitions.

## How Has This Been Tested?

The generated pipeline has been loaded into the production Concourse environment https://cicd.odl.mit.edu/teams/infrastructure/pipelines/concourse-build-and-deploy

## Types of changes

## Checklist:


Ardiea (approved):

src/concourse/lib/resources.py
```diff
@@ -1,10 +1,67 @@
-from concourse.lib.models import Identifier, Resource
```
shaidar, line 1: Think what's a bit misleading/confusing about this file is why is there a `Git` model resource but no model resources for the rest of the functions?

src/concourse/pipelines/infrastructure/concourse/pipeline.py
```diff
@@ -0,0 +1,216 @@
+from concourse.lib.constants import REGISTRY_IMAGE
```
shaidar, line 1: Don't think I'm very clear on what's going on in this file

blarghmatey: The idea of this module is to store values that are used frequently and are immutable. For instance `"registry-image" is a string we will be using often and this way we can avoid accidental typos that lead to enigmatic bugs.

Issue #945, opened 2022-07-11
Upgrade Redash deployment to latest release


blarghmatey:
We have been running Redash via Salt for several years at this point and the current installation is using an out of date version. The upstream Redash project recently published a v10 release that we would like to migrate to.

As part of this upgrade we also want to remove the dependency on SaltStack and use our new Pulumi/Packer/PyInfra toolchain to support it. The upstream project publishes a docker-compose file that we can use as the basis for our installation.

Some things to consider for this deployment are:
- The release that we are running is still using Celery workers but the newer versions migrated to use rq which may necessitate some changes to our configuration
- The workers are the processes that handle any query execution against the various data sources, meaning that they are going to be the most resource intensive part of the installation. We may want to consider running those separately from the web nodes.
- We currently have Redash running in our operations VPC but we would like to move this to our Data VPC for the new installation. This will likely require some updates to security groups to allow access to the various data sources being queried. The new security group should be labelled something like `business-intelligence` so that it can be applied seamlessly to any other BI service that we may migrate to in the future

Ardiea:
## Current State

Two new sites created with NGINX infront of a docker container running redash.

https://redash-qa.odl.mit.edu/
https://redash-production.mit.odl.edu

Everything seems to be working well? No complaints from users who have tried things so far. I believe Peter found some things that were hard-coded to 'bi.odl.mit.edu' in the dashboard but taht is okay because we will change the DNS (see below). 

## TODO 

1. Pick a date and time to formally cut over the environment. 
2. Message users about 'Forgot Password' link to use going forward rather than touchstone.
3. Take a look at @pdpinch's ticket here https://github.com/mitodl/ol-data-platform/issues/289 to remove the 'Remote Login' button. Should be possible. 
4. Do a final dump and load from https://bi.odl.mit.edu -> both new envs. Follow instructions in ol-infrastructure/src/ol-infrastructure/applications/redash/README.md on order of operations (involves invoking a manage command to upgrade v8 database to v10). 
5. Delete DNS entrie for bi.odl.mit.edu in Route53
6. Redeploy https://redash-production.odl.mit.edu -> https://bi.odl.mit.edu (Should be just a pulumi stack deployment).
7. Verify site functionality @ https://bi.odl.mit.edu
8. Rejoice.

Ferdi:
What is the reason for removing touchstone integration ?

Ardiea:
I defer to Tobias but I believe it came down to:

[...]

PR #963, opened 2022-07-20
Redash -> Docker/AMI/PyInfra/Pulumi stack

Ardiea:
## Description
This implements redash using our standard stack into AWS

## Motivation and Context
#945 

## How Has This Been Tested?
Slowly and by hand.

## Types of changes

## Checklist:


src/ol_infrastructure/applications/redash/__main__.py
```diff
@@ -85,10 +117,39 @@
...
+    bound_vpc_ids=[data_vpc["id"]],
+    token_policies=[redash_server_vault_policy.name],
+)
+
 redash_instance_security_group = ec2.SecurityGroup(
     f"redash-instance-{stack_info.env_suffix}",
     description="Security group to assign to Redash application to control "
     "inter-service access",
+    ingress=[
+        ec2.SecurityGroupIngressArgs(
```
blarghmatey, line 144: Rather than creating a specific security group for allowing web traffic we can just use the pre-made `web` security group from the network stack output. We may also want to assign the `data-integrator` security group for allowing access to the various databases until we sever those connections and move things into the lake/dbt workflow.

blarghmatey (commented):
A couple of minor comments but overall looks great

src/ol_infrastructure/applications/redash/__main__.py
Ardiea: Ummm, well that isn't the only thing this group is used for. I think maybe it is just convenient and explicit to include the HTTPS port here. It is used as an ingress for redash db security group, lines 158-187 and then in a similar manner for redis on lines 238-255.

If I were to use the default web group as an ingress target for this, then I would be opening up any web server to talk to Postgres + Redis, which I don't think we necessarily want, right? 

👍 on the data-integrator thing tho.

blarghmatey: All valid points :+1:

blarghmatey:
:+1:

PR #965, opened 2022-07-20
feat: Migrate Fastly management for edX into Pulumi

blarghmatey:
This moves the logic for managing the Fastly distribution for Open edX installations
into Pulumi.

## Description

- Adds a service definition for Fastly into edX Pulumi app
- Adds support for templating the MFE regex fields in the Fastly distribution

## Motivation and Context

We have been managing the Fastly distribution via click-ops up to now. As we work on enabling more MFEs it will be easier to keep things aligned by managing the Fastly configuration as code.

## How Has This Been Tested?

Imported the state of the MITx CI environment into the relevant stack and previewed the proposed diff to ensure it was reasonable.

## Types of changes

## Checklist:


src/ol_infrastructure/applications/edxapp/__main__.py
```diff
@@ -1269,6 +1272,159 @@ def cloud_init_user_data_func(
...
+            "backends",
+            "cache_settings",
+            "conditions",
+            "dictionaries",
+            "domains",
+            "name",
+            "snippets",
+        ],
+    }
+fastly_mitxonline_qa = fastly.ServiceVcl(
```
shaidar, line 1295: Shouldn't this be generic/templated?

src/ol_infrastructure/applications/edxapp/__main__.py
```diff
@@ -1269,6 +1272,159 @@ def cloud_init_user_data_func(
...
+            statement=f'req.url.path ~ "{mfe_regex}"',
+            type="REQUEST",
+        ),
+    ],
+    dictionaries=[
+        fastly.ServiceVclDictionaryArgs(
+            name="marketing_redirects",
+        )
+    ],
+    domains=[
```
shaidar, line 1342: Do we not want the domain name to be pulled from the stack config?

src/ol_infrastructure/applications/edxapp/__main__.py
blarghmatey: Good catch, fixing now...

src/ol_infrastructure/applications/edxapp/__main__.py
blarghmatey: They should come from the config, I'll fix that

shaidar:
Looks good now. I didn't compare the fastly config to what's in place now as I'm assuming u ran an export to get all that. 👍

PR #1008, opened 2022-08-18
Tm/airbyte fixup

blarghmatey:
This adds some fixes related to recovery after the Vault QA outage, as well as for integrating Airbyte with AWS Lake Formation

## Description
Update the Airbyte deployment to be more fully automated and update IAM permissions for interacting with necessary data lake resources.

## How Has This Been Tested?

The changes have been applied to the QA instance and the full end-to-end flow of data from source to lake and querying with Trino has been tested manually.

## Types of changes

## Checklist:

### DIFF
--- a/src/bilder/components/pomerium/files/config.yaml.tmpl
+++ b/src/bilder/components/pomerium/files/config.yaml.tmpl
@@ -3,7 +3,7 @@
 
 # this is the domain the identity provider will callback after a user authenticates
 # form: https://< whatever >
-authenticate_service_url: "https://{{ .Data.data.authenticate_server_address }}"
+authenticate_service_url: "https://{{ .Data.authenticate_server_address }}"
 
 # certificate settings:  https://www.pomerium.com/docs/reference/certificates.html
 # These paths are within the docker container, not the host machine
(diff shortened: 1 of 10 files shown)


blarghmatey:
One thing to consider here is that I updated the consul/vault templates to read from a v2 KV mount in Vault, but we currently have a v1 mounted. I think it would be useful to be able to upgrade if everything will work, but I'm also happy to back out those template changes.

Ardiea:
>One thing to consider here is that I updated the consul/vault templates to read from a v2 KV mount in Vault, but we currently have a v1 mounted. I think it would be useful to be able to upgrade if everything will work, but I'm also happy to back out those template changes. 

Just so I understand correctly, with **this PR**, it will work with either a v1 or a v2 mount. However, as the code is right now on `main`, it will only work with a v1 mount? 

That is, you could merge this and everything would be great and you could update the mount to v2 and everything would still be great.

blarghmatey:
Unfortunately that's not correct. I'm sure it would be possible to add some logic in the template to support V1 and V2, but as it stands right now the template will only function with v2 if this PR merges. Probably something worth investing in to paper over those differences since we currently don't have much rhyme or reason of which mounts are v1 and which are v2 and services are liable to span multiple mounts. I'll take a stab at what that might look like.

Ardiea:
Okay, I spent the better part of the past several hours playing with this as requested and I'm not sure it is ready to be merged yet until we have a good grasp on how to make the changes needed to the data layout and secrets engine in vault first.

blarghmatey:
I removed the v2 style lookups in the templates so that the rest of the changes don't get blocked as a result.

Ardiea (approved):

Issue #1020, opened 2022-08-22
Create Concourse pipelines for all Pulumi projects


blarghmatey:
# User Story
As a platform engineer I want to ensure that all Pulumi code is being exercised regularly and that I don't have to run the projects manually from my laptop.

# Description/Context
Once the work from #1019 is done, create pipelines for controlling our various Pulumi projects so that non-production stacks are executed automatically on relevant code changes.

# Acceptance Criteria
- [ ] A list of Pulumi projects that can and should be automated has been compiled (see below)
- [ ] A pipeline definition has been defined, uploaded to Concourse, and validated for each project in the list

## Pulumi Project Pipelines
- [ ] applications/ocw_studio
- [ ] applications/airbyte
- [ ] applications/ocw_site
- [ ] applications/tika
- [ ] applications/redash
- [ ] infrastructure/consul
- [ ] infrastructure/vector_log_proxy
- [ ] infrastructure/grafana_cloud
- [ ] infrastructure/aws/network
- [ ] infrastructure/aws/opensearch
- [ ] infrastructure/aws/dns
- [ ] infrastructure/aws/kms
- [ ] infrastructure/aws/policies

pdpinch:
So when this is done, where there be python code somewhere that defines each of these deployments? Where can I find it?

Ardiea:
`ol-infrastructure/src/concourse/pipelines/{open_edx,infrastructure}/...`

Ardiea:
I have a ton of work from this week that is not committed yet but I will do so at EOD.

PR #1028, opened 2022-08-25
Create reusable functions for building infrastructure Concourse pipelines

### LINKED ISSUE #1019: Simplify Concourse pipeline creation for infrastructure deployments
# User Story
As a platform engineer I want to have an easy way to create Concourse pipelines for building and deploying infrastructure resources. This mostly takes the form of EC2 AMIs and Pulumi projects

# Description/Context
Create a set of function definitions in the Concourse Python wrapper that allows for generating the common job definitions to support AMI builds with Packer and execution of Pulumi projects. This includes identifying the minimal set of inputs that are necessary to address the majority of our jobs, while also allowing for flexibility when we have a case that doesn't fit the exact same pattern.

# Acceptance Criteria
- [ ] A function exists in the Concourse library that generates a set of job definitions for validating Packer templates and building an AMI with Packer
- [ ] A function exists in the Concourse library that generates a set of job definitions for running a Pulumi project and sequences execution of related stacks appropriately
- [ ] An existing set of YAML pipeline definitions for building and deploying an AMI has been translated to use the new function definitions and deployed to Concourse

blarghmatey:
This is the next step in simplifying the work of building deployment pipelines for infrastructure resources

## Description
Writing pipelines for infrastructure resources has largely been simplified to "build an AMI" and "run the Pulumi project". Unfortunately, that still requires a non-trivial amount of YAML to make it happen in Concourse. In order to simplify our pipeline authoring the goal is to move that into pure Python. This takes the next step of adding helper functions for building those stages of the pipeline and wiring them together.

## Motivation and Context
This closes #1019 
This includes aspects of #1020 

## How Has This Been Tested?
The Airbyte service was used as a test case to exercise the new helper functions. The resulting pipeline has been loaded to https://cicd.odl.mit.edu/teams/infrastructure/pipelines/airbyte-build-and-deploy

## Types of changes

## Checklist:


src/concourse/pipelines/infrastructure/grafana_cloud/grafana_cloud_pipeline.py
```diff
@@ -0,0 +1,173 @@
+from concourse.lib.constants import REGISTRY_IMAGE
```
Ardiea, line 1: I think this is probably an example of a pipeline that is more complicated via python than just doing it in yaml. It was a good exercise but it didn't really lend itself to this framework well. Granted there may be opportunities to optimize this that I didn't notice/take.

src/concourse/pipelines/infrastructure/tika/packer_pulumi_pipeline.py
```diff
@@ -0,0 +1,68 @@
+from concourse.lib.constants import PULUMI_CODE_PATH
```
Ardiea, line 1: Packer+Pulumi pipelines, however do lend themselves well to this framework and it was fairly easy to stamp out pipelines like this with minimal effort once I got going. 

I think we could go even further and condense many of these into one pipeline that is built based on a data structure that is something like this 

```
{ "tika" : { "project_source_path" : "applications/tika/",
           "watch_paths" : ["src/bilder/components/","src/bilder/images/tika/","src/bridge/lib/versions.py"],
           "stages": ["CI, "QA", "Production"],
           "other_piece_of_relevant_information": "relevant information"}
"vector_log_proxy": { ... } 
}
```

src/concourse/pipelines/infrastructure/airbyte/pipeline.py
```diff
@@ -0,0 +1,71 @@
...
+)
+
+airbyte_pulumi_code = git_repo(
+    name=Identifier("ol-infrastructure-pulumi"),
+    uri="https://github.com/mitodl/ol-infrastructure",
+    paths=[
+        "src/ol_infrastructure/applications/airbyte/",
+        "pipelines/infrastructure/scripts/",
+    ],
+)
```
Ardiea, line 24: I think we are going to need to re-evaluate the `paths` that we setup as part of these packer+pulumi jobs.

I'm not sure if I can articulate, at the moment, what the issue here is but it isn't _quite_ right. 

I think the AMI paths need to be expanded to include other 'shared' things like `vector`, `vault`, `consul`, `docker-baseline-ami`, but of course this would have the side effect of trigger more builds and more builds simultaneously. 

Second, The link between the AMI bake and the pulumi state refresh + update makes sense for some situations but not others. I believe, as this pipeline is structured now, the pulumi states will be updated when a new AMI is baked but NOT when there is a change to say `src/ol-infrastructure/applications/airbyte/*` Which I think we might want, too. 

Third there is a blind break between the AMI build and the pulumi state updates. That is, there is nothing saying that the AMI made in the packer steps is what will be used in the pulumi steps. I don't think there is necesarily a resolution to this or that it is a serious issue at the moment, but I think it is something to be aware of for debugging. There really is no relationship between the two except temporal locality or something cleverly phrased like that. 

There are more problem but I can't think of them at the moment.

Ardiea:
_Almost_ Sigh. 

```
Source      Dest
10          10
43843       0            <<<<<<<  discussions-rc_cb9d3b5b4f0f4a5783598e5ab6d6104f
159501      159501
1051178     1051169      <<<<<<<  discussions-rc_course_2ad6f3356f6d4d31b6ce2a352918b270
38          38
1880        1880
84240       84240
216229      216229
30          30
4           4
59          59
9253        9253
1945        1945
1426        1426
59          59
```

PR #1062, opened 2022-09-09
Add pipeline for ocw_site deployment

blarghmatey:
This adds a pipeline definition for the ocw_site Pulumi project.

## Description
As part of that this addresses a behavioral aspect of Pydantic where validators only run on attribute assignment, not attribute mutation. This meant that duplicate resources and resource_types could be created when the attribute was mutated, so a fix is added in the pulumi_jobs_chain to assign on each iteration to trigger the validators. As part of running in the context of a Concourse pipeline the Fastly API key is needed, so this also addresses that need by using an explicity Fastly provider and aliases it to avoid triggering replacement of child resources.

## Motivation and Context
We want to allow developers to manage deployment/updates of Fastly configuration for the OCW site.

## How Has This Been Tested?
The changes have been run manually from my laptop, and the pipeline is visible at https://cicd.odl.mit.edu/teams/infrastructure/pipelines/pulumi-ocw-site

## Types of changes

## Checklist:

### DIFF
--- a/src/concourse/lib/models/fragment.py
+++ b/src/concourse/lib/models/fragment.py
@@ -11,7 +11,7 @@ class PipelineFragment(BaseModel):
     @validator("resource_types")
     def deduplicate_resource_types(
         cls: "PipelineFragment", resource_types: list[ResourceType]
-    ):
+    ) -> list[ResourceType]:
         """Ensure that there are no duplicate resource type definitions.
 
         Concourse pipelines don't support duplicate definitions of resource types, where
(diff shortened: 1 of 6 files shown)


src/concourse/lib/models/fragment.py
```diff
@@ -11,43 +11,57 @@ class PipelineFragment(BaseModel):
...
         the `name` identifier is used to determine uniqueness.  This ensurs that
         `PipelineFragment` objects can be composed together without violating that
         requirement.
 
+        NOTE: This validator is only triggered on new assignments to the attribute, not
+        mutations of existing instances of the attribute.  In other words, if you
+        `.append` or `.extend` this list the validator will *not* trigger.  See
+        `Pydantic #496 <https://github.com/pydantic/pydantic/issues/496>`_ for more
+        details
+
```
Ardiea, line 27: 👍

Ardiea (approved):

PR #1070, opened 2022-09-13
Remove salt integration and other dead code

blarghmatey:
This removes code for Pulumi projects that are no longer in use, as well as deleting the SaltStack integration that is no longer required.

## Description
We have been eliminating our dependency on SaltStack in favor of using Packer and PyInfra for building deployable images and managing the environments with Pulumi. During the initial implementation of the new stack we needed to maintain compatibility with Salt until we could build up enough other supporting infrastructure. Now that we are no longer using Salt to manage any of the systems deployed from this repository we can remove the code that was necessary during that transition.

## Motivation and Context
See above.

## How Has This Been Tested?
N/A - this is deleting code that is no longer in use. All affected projects/stacks have been destroyed.

## Types of changes

## Checklist:


blarghmatey:
This won't be ready to merge until we destroy the apps.elasticsearch.{QA,Production} stacks, which is ready to happen once we are sufficiently confident with the usage of the AWS managed ES service running with the MIT Open app as of this morning.

shaidar (approved):

PR #1073, opened 2022-09-15
Ol-Infra-1007 Add configs for user retirement

asadali145:
## Types of changes

## Checklist:


asadali145:
@blarghmatey @pdpinch 
Here are a few more things that we need to do but we are not sure how to manage:

- Install https://github.com/edx/tubular in edX instance at `/src`. Not sure how we are going to install it at `/src`.
- Create a virtual environment and install requirements for https://github.com/edx/tubular.
- Create a config.yml that we are going to need for https://github.com/edx/tubular. Once tubular is cloned we will have to move this to tubular.

Once all of this is set up. Here are the steps to retire a user. 

- Run the `retire_users ` command in MITxOnline to retire it from MITxOnline.
- Run the `retire_user` command in edX to retire the user in the edX. This will create a Retirement Status with a `PENDING` state.
- Activate the `tubular` virtual environment.
- Run `retire_one_learner.py <username>`. This will take a user through the retirement pipeline and eventually it will mark the retirement status as `COMPLETED` if everything goes well.
PS: For the last 3 steps in retirement, we can create a shell script that will execute these steps.

blarghmatey (commented):
Small typo, but otherwise looks good.

blarghmatey:
As far as the usage of Tubular, when I looked at the retirement scripts it seemed to be using the REST APIs, so we should be able to run that in a Concourse pipeline, removing the need to package that project into the same instance as the edX application.

For the first two steps, are those only available as management commands, or can those also be triggered via an API?

asadali145:
> For the first two steps, are those only available as management commands, or can those also be triggered via an API?

- In MITxOnline, We do not have an API for user retirement. We can create one if we need.
- In edX, we have 2 APIs to start user retirement.
  - `api/v1/accounts/deactivate_logout/` is the one built for the users to delete the accounts.
  - `v1/accounts/bulk_retire_users` can be used by service users but it has authentication issues in edX, If we add `api/` at the start of this URL then it works fine.

Let me know if this still does not answer your question.

blarghmatey:
Closing as a duplicate. This change was added in a previous PR.

PR #1112, opened 2022-09-30
OVS static assets phase one 

Ardiea:
The pipeline will probably require revisiting, definitely for the prefix where it is placing files at the moment: `mike-testing`.

## Description
- Concourse pipeline that generates OVS static assets and puts them into S3 (CI)
- Pulumi stacks for each new OVS environment
  - Only creates the S3 static asset buckets at the moment.
- Let concourse write to the S3 static asset buckets. 

## Motivation and Context
Migration from Salt-stack to docker-compose framework.

## How Has This Been Tested?
- Pulumi stacks are all already created. 
- Pipeline was tested and iterated over as it was developed. 

## Types of changes

## Checklist:


src/concourse/pipelines/infrastructure/ovs/static_assets_pipeline.py
```diff
@@ -0,0 +1,171 @@
...
+                        args=[
+                            "bash",
+                            "-ceux",
+                            textwrap.dedent(
+                                """mount -t cgroup -o none,name=systemd cgroup /sys/fs/cgroup/systemd
+                                 cd odl-video-service
+                                 chmod 777 .
+                                 env > .env
+                                 docker-compose build
+                                 whoami
```
Ardiea, line 116: Woops. This can go as well. Debugging remnant.

src/concourse/pipelines/infrastructure/ovs/static_assets_pipeline.py
```diff
@@ -0,0 +1,171 @@
...
+                                 docker-compose run -u root watch yarn install --frozen-lockfile --ignore-engines --prefer-offline
+                                 docker-compose run -u root watch node node_modules/webpack/bin/webpack.js --config webpack.config.prod.js --bail
+                                 docker-compose run web python manage.py collectstatic --no-input
```
blarghmatey, line 119: Do we need to use a full docker-compose stack for running the `collectstatic` step? Can we run that in just the OVS container?

src/concourse/pipelines/infrastructure/ovs/static_assets_pipeline.py
```diff
@@ -0,0 +1,171 @@
...
+                            "-ceux",
+                            textwrap.dedent(
+                                """mount -t cgroup -o none,name=systemd cgroup /sys/fs/cgroup/systemd
+                                 cd odl-video-service
+                                 chmod 777 .
+                                 env > .env
+                                 docker-compose build
+                                 whoami
+                                 docker-compose run -u root watch yarn install --frozen-lockfile --ignore-engines --prefer-offline
+                                 docker-compose run -u root watch node node_modules/webpack/bin/webpack.js --config webpack.config.prod.js --bail
```
blarghmatey, line 118: Do we need the docker-compose stack for running this step? It seems like we could just use a `node` container to run the webpack compilation.

src/ol_infrastructure/applications/ovs/Pulumi.yaml
```diff
@@ -0,0 +1,6 @@
+---
+name: ol-infrastructure-ovs-env
```
blarghmatey, line 2: I wonder if we should expand `ovs` -> `odl-video-service` for sake of clarity? It's easy to get confused about what the acronym is supposed to represent for anyone who isn't familiar with this app.

src/concourse/pipelines/infrastructure/ovs/static_assets_pipeline.py
```diff
@@ -0,0 +1,171 @@
...
+                        "VIDEO_S3_TRANSCODE_BUCKET": "video-s3-transcodes",
+                        "VIDEO_S3_THUMBNAIL_BUCKET": "video-s3-thumbs",
+                        "VIDEO_S3_SUBTITLE_BUCKET": "video-s3-subtitles",
+                        "VIDEO_S3_WATCH_BUCKET": "video-s3-watch",
+                        "VIDEO_CLOUDFRONT_DIST": "video-cf",
+                        "YT_ACCESS_TOKEN": "fake",
+                        "YT_REFRESH_TOKEN": "fake",
+                        "YT_CLIENT_ID": "fake",
+                        "YT_CLIENT_SECRET": "false",  # pragma: allowlist secret
+                        "YT_PROJECT_ID": "fake",
```
blarghmatey, line 103: Are we able to trim/simplify this at all for the asset compilation step? /cc @mbertrand

src/concourse/pipelines/infrastructure/ovs/static_assets_pipeline.py
Ardiea: That is a good question. I figured since we were already here with a good amount of the rest of the stack for the node steps, we could just bang out the `collectstatic` as well. I will look into it further.

src/concourse/pipelines/infrastructure/ovs/static_assets_pipeline.py
Ardiea: Yeah, I'm not sure. I will look into it.

PR #1122, opened 2022-10-05
Added mitx production file

mitangelo:
Some updates to the codejail dockerfile. Some dependencies were updated. This has been successfully tested. 
## Types of changes

## Checklist:

### DIFF
--- a/dockerfiles/openedx-codejail/Dockerfile
+++ b/dockerfiles/openedx-codejail/Dockerfile
@@ -1,9 +1,10 @@
 FROM ubuntu:focal
 SHELL ["/bin/bash", "-c"]
 
-# Install Codejail Packages
+ENV DEBIAN_FRONTEND=noninteractive
+ENV DEBCONF_NONINTERACTIVE_SEEN=true
 RUN apt-get update && apt-get upgrade -y
-RUN apt-get install -y vim python3-virtualenv python3-pip git sudo
+RUN apt-get install -y vim python3-virtualenv python3-pip git sudo libxslt-dev
 
 # Define Environment Variables
 ENV CODEJAIL_GROUP=sandbox
@@ -11,7 +12,6 @@ ENV CODEJAIL_SANDBOX_CALLER=ubuntu
 ENV CODEJAIL_USER=sandbox
 ENV CODEJAIL_VENV=/sandbox/venv/
 
-# Create Virtualenv for sandbox user
 RUN virtualenv -p python3.8 --always-copy $CODEJAIL_VENV
 ENV PATH="$CODEJAIL_VENV/bin:$PATH"
 
@@ -26,13 +26,13 @@ RUN adduser --disabled-login --disabled-password ubuntu --ingroup ubuntu
 # Give Ownership of sandbox env to sandbox group and user
 RUN chown -R $CODEJAIL_USER:$CODEJAIL_GROUP $CODEJAIL_VENV
 
-# ADD . ./codejail
 WORKDIR /codejail
 RUN git clone https://github.com/eduNEXT/codejailservice/ --branch main --depth 1 /codejail
+ADD requirements.txt requirements.txt
 
 # Install dependencies
-RUN source $CODEJAIL_VENV/bin/activate && pip install -r requirements/base.txt -r requirements/dev.txt && \
-    pip install gunicorn random2 && deactivate
+RUN source $CODEJAIL_VENV/bin/activate && pip install --no-cache-dir --no-cache-dir -r requirements/base.txt  && \
+    pip install --no-cache-dir -r  https://github.com/openedx/edx-platform/blob/master/requirements/edx-sandbox/py38.txt && deactivate
 
 # Setup sudoers file
 ADD sudoers-file/01-sandbox /etc/sudoers.d/01-sandbox


dockerfiles/openedx-codejail/Dockerfile
```diff
@@ -32,7 +32,9 @@ RUN git clone https://github.com/eduNEXT/codejailservice/ --branch main --depth
 
 # Install dependencies
 RUN source $CODEJAIL_VENV/bin/activate && pip install -r requirements/base.txt -r requirements/dev.txt && \
-    pip install gunicorn random2 && deactivate
+    pip install gunicorn MarkupSafe==2.1.1 Pillow==9.1.0 cffi==1.15.0 chem==1.2.0 click==8.1.2 cryptography==36.0.2 cycler==0.11.0 joblib==1.1.0 \
```
Ardiea, line 35: my only concern he is that dependabot or rennovate may not keep an eye on these? Could we add a `requirements.txt` to this repo and then copy it into the container as `/codejail/requirements/mit.txt` immediately after the `git clone` on line 31? Then just include it with another -r on Line 34?

mitangelo: The thing is we are not using a repo for this one....The Dockerfile clones the repo and pip install the requirements from there...These additional dependencies are ours. You are suggesting to put them in a requirements file in the same folder location?

Ardiea: yeah just a reqs file in this directory with this Dockerfile and then copy it into the edx repo that cloned. I guess it doesn't really matter where into the container you copy it, just seems easiest to put it next to the ones that come from edx.

blarghmatey: So, there already is a requirements file for all of this. We should just be using [this file](https://github.com/openedx/edx-platform/blob/master/requirements/edx-sandbox/py38.txt) directly for building the image. Ideally the Concourse pipeline would clone the edx-platform repository as a dependency, using the release branch that we are deploying for a given installation, and add that repository to the build context for the docker image so that we can copy in whatever we need during the build (e.g. the requirements file and the in-repo library that is being installed [here](https://github.com/openedx/edx-platform/blob/open-release/nutmeg.master/requirements/edx-sandbox/py38.txt#L7), though that is no longer present in the `master` branch)

dockerfiles/openedx-codejail/Dockerfile
```diff
@@ -32,7 +32,9 @@ RUN git clone https://github.com/eduNEXT/codejailservice/ --branch main --depth
 
 # Install dependencies
 RUN source $CODEJAIL_VENV/bin/activate && pip install -r requirements/base.txt -r requirements/dev.txt && \
```
blarghmatey, line 34: We shouldn't need to install the dev requirements since this image won't be used for development. We will only need the `base.txt` requirements file to run this service.

Also, in order to reduce the overall image size we should use the `--no-cache-dir` flag for `pip install` so that the downloaded files are discarded after installing.

dockerfiles/openedx-codejail/Dockerfile
mitangelo: @blarghmatey how can we use that file? i already tested the code with the dockerfile i created. do u want to replace those changes with the reqwuirements file u provided?

mitangelo: Also the directions that you mentioned seem a bit overly complex. Why would we need to clone the edx-platform while building a codejail docker image? Only for dependencies?

Ardiea: 👍

blarghmatey: Yes, we will need to use that file because it is maintained by the edX community, and will change with the different versions of edX that we are using this codejail service alongside. We don't necessarily have to clone the edX repository, but we do need to use that requirements file for managing the dependencies that are included in the codejail environment.

blarghmatey: @mitangelo we don't want to arbitrarily update the library versions using e.g. Renovate, because we need to match what course instructors are expecting and what other edX environments (e.g. edx.org) are using, because course authors are likely to run their same course in multiple installations. This is why we want to pull the requirements file from the edx-platform repository, because that is the list of currently installed and supported libraries that is used in other codejail environments.

Since this codejail service will be deployed alongside each of our Open edX installations, we will need to ensure that the version of the requirements file that we use to build the container launched into that environment, we will also need to ensure that the branch of edx-platform that we pull the requirements file from matches the version of Open edX that we are running in that environment (e.g. `open-release/nutmeg.master` for MITx and MITx-staging, `master` for MITx Online, `open-release/maple.master` for xPro).

mitangelo: @blarghmatey how can we implement this?

mitangelo: @blarghmatey  we would need to track this right. If versions change.

blarghmatey: The `extra_files` doesn't handle this, because that is something that is specified on a per-course basis by each of the course authors. They are all expecting that the libraries listed in the linked requirements file are already present so they aren't going to bundle them in again for their own extra dependencies.

As to how to implement this, I think it would make the most sense to write it as a [multi-stage build](https://docs.docker.com/build/building/multi-stage/), with the base stage doing the work of installing the `eduNext/codejailservice`, and then the additional stage(s) would handle the installation of the edx-platform sandbox requirements. That could be done by referencing the remote requirements file as described [here](https://stackoverflow.com/questions/70422673/pip-install-python-packages-from-requirements-txt-file-on-github), with the blob reference pointing to the appropriate branch as mentioned above. We could even feasibly write a single multi-stage build that references each of the versions that we support at a given time (currently open-release/maple.master, open-release/nutmeg.master, and master)

blarghmatey: In terms of tracking when the file changes, Concourse makes that trivial with the git resource combined with the `paths` parameter to watch that `requirements/python38.txt` path.

mitangelo: ok so the dockerfile could have the url to the res files.

mitangelo: I did not know that the requirements you had given me the other day were installed as a part of edx. I thought we were installing them ourselves.

mitangelo: > In terms of tracking when the file changes, Concourse makes that trivial with the git resource combined with the `paths` parameter to watch that `requirements/python38.txt` path.

Can you show me an example of this?

blarghmatey: There's an example of triggering based on path [here](https://github.com/mitodl/ol-infrastructure/blob/main/src/concourse/pipelines/infrastructure/ocw_site/pipeline.py#L6-L15)

dockerfiles/openedx-codejail/Dockerfile
```diff
@@ -26,13 +26,13 @@ RUN adduser --disabled-login --disabled-password ubuntu --ingroup ubuntu
...
 
-# ADD . ./codejail
 WORKDIR /codejail
 RUN git clone https://github.com/eduNEXT/codejailservice/ --branch main --depth 1 /codejail
+ADD requirements.txt requirements.txt
 
 # Install dependencies
-RUN source $CODEJAIL_VENV/bin/activate && pip install -r requirements/base.txt -r requirements/dev.txt && \
-    pip install gunicorn random2 && deactivate
+RUN source $CODEJAIL_VENV/bin/activate && pip install --no-cache-dir --no-cache-dir -r requirements/base.txt  && \
```
blarghmatey, line 34:
```suggestion
RUN source $CODEJAIL_VENV/bin/activate && pip install --no-cache-dir -r requirements/base.txt  && \
```

mitangelo:
Need to test this now that its approved. Will merge once successfully tested.

PR #1124, opened 2022-10-06
PyInfra component definition for Traefik proxy server

blarghmatey:
This adds the component definition and configuration models for installing [Traefik](https://traefik.io/) in a binary format.

## Description
The Traefik proxy is widely adopted and stable, with a large community and consistent interfaces for configuration. This adds support for installing it using our PyInfra code.

## Motivation and Context
We have recently run into issues with the Route53 DNS resolver plugin on Caddy where it fails to generate the challenge properly, leading to problems in our deployment of Vault and Consul servers (constrained to CI stages so far). Despite some upstream fixes, the problem persists. With a brief attempt at experimenting with Traefik I was able to replicate the functionality that we had with Caddy and without requiring any additional plugins. This adds the logic to allow for installing and configuring the Traefik binary as a proxy service.

## How Has This Been Tested?
The `sample_deploy.py` file was run against a local Vagrant host to ensure that the sample configuration is rendered properly and the binary is placed at the proper location on disk.

## Types of changes

## Checklist:


src/bilder/components/traefik/templates/conf.ini.j2
```diff
@@ -0,0 +1,3 @@
+[{{ settings.section }}]
+foo = {{ settings.foo }}
```
Ardiea, line 2: ?

Ardiea (approved):
👍

PR #1135, opened 2022-10-11
Md/ovs migration2 - Checkpoint

Ardiea:
## Description
- Completed configuration for building an OVS AMI, the associated AWS resources, and deploying an OVS environment. 

## How Has This Been Tested?
Deployed a QA environment with a pg_dump and pg_restore from the existing environment. Next step is to replace the existing environment. 

## Types of changes

- [X] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:


src/bilder/images/odl_video_service/files/.env.tmpl
```diff
@@ -0,0 +1,75 @@
+VERSION={{ file "/etc/default/ovs-version" }}
+
+{{ with secret "postgres-odl-video-service/creds/app" }}
+DATABASE_URL="postgres://{{ .Data.username }}:{{ .Data.password }}@{{ keyOrDefault "ovs/database_endpoint" "missing_value_database_endpoint" }}/odlvideo"
+{{- end }}
+
+{{ with secret "secret-operations" }}
+CLOUDFRONT_KEY_ID="{{ index .Data "global.cloudfront-private-key.id" }}"
+CLOUDFRONT_PRIVATE_KEY="{{ index .Data "global.cloudfront-private-key.value" }}"
+MAILGUN_KEY="{{ index .Data "global.mailgun-api-key.value" }}"
```
blarghmatey, line 10: Is this a syntax that we can use to abstract the difference between kv-v1 and kv-v2?

Ardiea: we could? but that isn't formally what it is for. This was because of the hyphens in the key names. The normal dot notation would not work.

src/bilder/lib/template_helpers.py
```diff
@@ -0,0 +1,133 @@
...
+    name: str,
+    repo_path: Path,
+    template_path: Path,
+    destination_path: Path,
+    consul_templates: list[ConsulTemplateTemplate],
+    watched_files: list[Path],
+    mode: str = "0664",
+    user: str = "consul-template",
+    group: str = "consul-template",
+):
```
blarghmatey, line 75: This might be better (from a design perspective) if it returns the `ConsulTemplateTemplate` object. The `watched_files` argument for the service watches can then just be generated as `watched_files=[template.destination for template in consul_template_config.template]`

That produces a more functional interface without implicit state mutation that can lead to "spooky action at a distance".

src/bilder/lib/template_helpers.py
```diff
@@ -0,0 +1,133 @@
...
+
+CONSUL_TEMPLATE_DIRECTORY = "/etc/consul-template"
+
+
+def place_jinja_template_file(  # noqa: WPS211
+    name: str,
+    repo_path: Path,
+    destination_path: Path,
+    context: dict,
+    watched_files: list[Path],
```
blarghmatey, line 15: I think that it's best if we don't mutate the `watched_files` as part of the body of the function. What might work better is if the return value of the function is the destination path of the file on disk, and that can then be used in a pattern like:
```
watched_files.append(place_jinja_template_file(...))
```

src/ol_infrastructure/applications/concourse/iam_policies/infra.py
```diff
@@ -49,6 +49,20 @@
...
+                "s3:PutObject",
+                "s3:PutObjectTagging",
+                "s3:DeleteObject",
+                "s3:ListBucket*",
+            ],
+            "Resource": [
+                "arn:aws:s3:::ovs-static-assets-*",
+                "arn:aws:s3:::ovs-static-assets-*/",
+            ],
+        },
```
blarghmatey, line 65: This can get backed out now that we're not going the pre-built assets approach.

src/ol_infrastructure/applications/odl_video_service/Pulumi.applications.odl_video_service.QA.yaml
```diff
@@ -0,0 +1,28 @@
...
+  - ttv-dev.mit.edu;
+  ovs:edx_base_url: https://courses-rc.xpro/mit.edu
+  ovs:log_level: INFO
+  ovs:rds_password:
+    secure: v1:/n8kQRfM6Od/xUKw:wwUBWVPgtmgRrN9jOA5djHdkizHjPYsUI9UDZFf5eYeOGUjZXTqSYStSK8lizXRyujf5SLfKiWqXYUeQP97QumhPuTuNbDyw+6ByeKT49P8=
+  ovs:s3_bucket_name: odl-video-service-rc
+  ovs:s3_subtitle_bucket_name: odl-video-service-subtitles-rc
+  ovs:s3_thumbnail_bucket_name: odl-video-service-thumbnail-rc
+  ovs:s3_transcode_bucket_name: odl-video-service-transcode-rc
+  ovs:s3_watch_bucket_name: odl-video-service-watch-rc
```
blarghmatey, line 23: It might be good to just create the buckets in Pulumi and force a bucket name for consistency and then migrate contents to the intended buckets using `s3 sync`. That way we can ensure that they are set up properly and keep them in line with what the code says.

Ardiea: I like the sound of this but I think there is a complication. I am fairly sure that the bucket name is recorded in the database somewhere. I say this because at one point during my testing I had the configuration using an invalid bucket name, 'uploaded' a file, fixed the configuration, went to delete the file and the logs included the old bucket name, that theoretically the app shouldn't have known about any longer if its only reference for bucket names was the configuration items. 

That experience makes me think it is going to be more than just a s3 sync and it is going to require some analysis and SQL development.

blarghmatey (approved):
:+1: Looks great!

Issue #1044, opened 2022-09-02
Set up database replication from production to RC for ocw-studio


gumaerc:
# User Story
When testing OCW related PR's in RC, there has been some trouble with outdated data. Now that the OCW team is fully utilizing `ocw-studio` in production and making all their changes there, we should backpopulate RC's database with the contents of the production database. This will ensure that our testing is run against relevant, up-to-date data.

# Description/Context
A database replication job needs to be set up to do a `pgdump` of the prod database and restore it to RC on a regular basis. I'll leave it up to someone in devops to add details on the best way to accomplish this.

# Acceptance Criteria
 - The RC database is restored from production on a regular basis (weekly? monthly?)
 - All ODL developers need to be able to still use their touchstone credentials to log in and be staff / superuser
 - Actions taken on RC after the restore cannot affect production data

PR #1151, opened 2022-10-17
ocw-studio replication pipeline

gumaerc:
## Description
Closes https://github.com/mitodl/ol-infrastructure/issues/1044

This PR adds a Python based pipeline definition to the repo for performing a Postgres DB replication on the `ocw-studio` database, based on a source and destination.

## Motivation and Context
The primary motivation behind this work is to be able to push a pipeline up into QA Concourse to perform a replication from production to QA for `ocw-studio`. This pipeline can also be used with the local Concourse instance spun up by `ocw-studio` to perform a restore of either QA or production to your local database. Initially [Replibyte](https://github.com/Qovery/Replibyte) was trialed for this work, but a number of issues were run into that prevented its use:

 - Multiple issues with escape characters in markdown causing the Replibyte dump creation to fail
 - Even with those "corrected," I encountered the worst memory leak issue I have ever seen

## How Has This Been Tested?
The pipeline generation code has been tested locally using the `poetry` setup, and the pipeline itself has been tested against a locally running version of `ocw-studio` with its own Concourse instance.

## Types of changes

## Checklist:

### DIFF
--- a/src/concourse/pipelines/infrastructure/ocw_studio_db_replication/pipeline.py
+++ b/src/concourse/pipelines/infrastructure/ocw_studio_db_replication/pipeline.py
@@ -1,0 +1,121 @@
+from calendar import timegm
+from datetime import datetime
+
+from concourse.lib.constants import REGISTRY_IMAGE
+from concourse.lib.models.pipeline import (
+    AnonymousResource,
+    Command,
+    GetStep,
+    Identifier,
+    Input,
+    Job,
+    Output,
+    Pipeline,
+    Platform,
+    RegistryImage,
+    TaskConfig,
+    TaskStep,
+)
+from concourse.lib.resources import git_repo
+
+ocw_studio = git_repo(
+    Identifier("ocw-studio"),
+    uri="https://github.com/mitodl/ocw-studio",
+    branch="cg/replibyte-config",
+    check_every="60m",
+)
+
+alpine_resource = AnonymousResource(
+    type=REGISTRY_IMAGE,
+    source=RegistryImage(repository="marsblockchain/curl-git-jq-wget"),
+)
+
+postgres_resource = AnonymousResource(
+    type=REGISTRY_IMAGE, source=RegistryImage(repository="postgres", tag="latest")
+)
+
+
+def db_replication_pipeline() -> Pipeline:
+    db_replication_job = Job(
+        name="ocw-studio-db-replication",
+        build_log_retention={"builds": 10},
+        plan=[
+            GetStep(get=ocw_studio.name, trigger=False),
+            TaskStep(
+                task=Identifier("install-replibyte"),
+                privileged=False,
+                config=TaskConfig(
+                    platform=Platform.linux,
+                    image_resource=alpine_resource,
+                    inputs=[Input(name=ocw_studio.name)],
+                    outputs=[Output(name=ocw_studio.name)],
+                    run=Command(
+                        path="/bin/sh",
+                        dir="./ocw-studio/replibyte",
+                        args=["./install.sh"],
+                    ),
+                ),
+            ),
+            TaskStep(
+                task=Identifier("replibyte-dump-create"),
+                privileged=False,
+                config=TaskConfig(
+                    platform=Platform.linux,
+                    image_resource=postgres_resource,
+                    inputs=[Input(name=ocw_studio.name)],
+                    outputs=[Output(name="dump")],
+                    params={
+                        "SOURCE_DB_URI": "((ocw-studio-production-database-uri))",
+                        "DESTINATION_DB_URI": "((ocw-studio-qa-database-uri))",
+                    },
+                    run=Command(
+                        path="./ocw-studio/replibyte/replibyte",
+                        args=[
+                            "--config",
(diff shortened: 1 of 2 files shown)


src/concourse/pipelines/infrastructure/ocw_studio_db_replication/pipeline.py
```diff
@@ -0,0 +1,84 @@
...
+        plan=[
+            TaskStep(
+                task=Identifier("run-pg-dump"),
+                privileged=False,
+                config=TaskConfig(
+                    platform=Platform.linux,
+                    image_resource=postgres_resource,
+                    outputs=[Output(name="dump")],
+                    params={
+                        "PGHOST": "((source.host))",
```
blarghmatey, line 34: Is this credential being written to a Vault path scoped to the pipeline? Otherwise I'd suggest that the var name be changed to something more descriptive/distinctive. (same for the `destination` name below)

Ardiea: So based on where this pipeline is run (team + name), the secret path is actually @ 
`ocw/ocw-studio-db-replication/source` and `
ocw/ocw-studio-db-replication/destination` which is kind of puzzling where that middle part comes from. It is obviously the pipeline name but I don't feel like I've seen it before.

src/concourse/pipelines/infrastructure/ocw_studio_db_replication/pipeline.py
```diff
@@ -0,0 +1,84 @@
...
+
+if __name__ == "__main__":
+    import sys  # noqa: WPS433
+
+    with open("definition.json", "w") as definition:
+        definition.write(db_replication_pipeline().json(indent=2))
+    sys.stdout.write(db_replication_pipeline().json(indent=2))
+    sys.stdout.write("\n")
+    sys.stdout.write(
+        "fly -t local set-pipeline -p ocw-studio-db-replication -c definition.json\n"
```
Ardiea, line 83:
```suggestion
        "fly -t qa-ocw set-pipeline -p ocw-studio-db-replication -c definition.json\n"
```

src/concourse/pipelines/infrastructure/ocw_studio_db_replication/pipeline.py
```diff
@@ -0,0 +1,84 @@
...
+                        "PGHOST": "((source.host))",
+                        "PGPORT": "((source.port))",
+                        "PGUSER": "((source.username))",
+                        "PGPASSWORD": "((source.password))",
+                    },
+                    run=Command(
+                        path="/bin/sh",
+                        args=[
+                            "-c",
+                            "pg_dump -Fc -d ((source.database)) > ./dump/db.dump",
```
Ardiea, line 43:
```suggestion
                            "pg_dump -v -Fc -d ((source.database)) > ./dump/db.dump",
```

src/concourse/pipelines/infrastructure/ocw_studio_db_replication/pipeline.py
```diff
@@ -0,0 +1,84 @@
...
+                        "PGHOST": "((destination.host))",
+                        "PGPORT": "((destination.port))",
+                        "PGUSER": "((destination.username))",
+                        "PGPASSWORD": "((destination.password))",
+                    },
+                    run=Command(
+                        path="/bin/sh",
+                        args=[
+                            "-c",
+                            "pg_restore --clean --no-privileges --no-owner -d ((destination.database)) ./dump/db.dump",
```
Ardiea, line 65:
```suggestion
                            "pg_restore -v --clean --no-privileges --no-owner -d ((destination.database)) ./dump/db.dump",
```

src/concourse/pipelines/infrastructure/ocw_studio_db_replication/pipeline.py
Ardiea: I checked this output with adding -v (verbose), it doesn't dump everything onto the console, just table + index names, which I think is a good compromise between "nothing" and "everything".

Ardiea (commented):
Looks good 👍 Just a couple suggestions

gumaerc:
@Ardiea all set, thanks

Ardiea (approved):

src/concourse/pipelines/infrastructure/ocw_studio_db_replication/pipeline.py
gumaerc: This is inferred from the pipeline name at the time `set-pipeline` is run based on the command printed at the end: `fly -t qa-ocw set-pipeline -p ocw-studio-db-replication -c definition.json`. It could be clearer that the pipeline name is actually important, but either way if it's not changed then it should take vars set on the pipeline name as described [here](https://concourse-ci.org/vault-credential-manager.html#vault-credential-lookup-rules) with `PIPELINE_NAME` being the name of the pipeline when pushed up. Do y'all think I should still be more explicit and have it not depend on the pipeline name?

Ardiea: I think it is probably okay how it is. It makes sense to me now.

Issue #1009, opened 2022-08-19
Migrate management of Bootcamps resources into Pulumi


blarghmatey:
# User Story
As a platform engineer I want to make sure that the Bootcamps application and its infrastructure resources are being managed in the same way as all of our other systems.

# Description/Context
The Bootcamps application was set up before our current usage of Pulumi and before we had established some of the patterns that led to the creation of the `apps` VPC. The result is that the RDS instance used in production was created in its own VPC and the Vault policies are managed via Salt code which is being actively deprecated. There are also associated S3 buckets that were created either manually or via Salt and are not being actively managed by Pulumi.

# Acceptance Criteria
- [x] Create a Pulumi project for managing the Bootcamps infrastructure resources
- [x] Either import or recreate the Bootcamps RDS instance in the applications VPC using Pulumi
- [x] Import the S3 buckets used by Bootcamps into the stack state for the Pulumi project
- [ ] Remove the associated code (e.g. Vault policies, S3 bucket management, etc.) from the salt-ops project

mitangelo:
There should be a stack definition for each environment that the bootcamp app is deployed to (namely CI, QA, and Production). This will also need a Vault policy definition for the AWS resources that it is using and the RDS instance.

PR #1170, opened 2022-10-24
Added bootcamp. Imported

mitangelo:
Imported bootcamp infra. 
## Types of changes

## Checklist:

### DIFF
--- a/src/ol_infrastructure/applications/bootcamp/__main__.py
+++ b/src/ol_infrastructure/applications/bootcamp/__main__.py
@@ -1,17 +1,16 @@
 import pulumi
 import pulumi_aws as aws
 from pulumi_aws import Provider
+
+from bridge.lib.magic_numbers import AWS_RDS_DEFAULT_DATABASE_CAPACITY
 from ol_infrastructure.components.aws.database import OLAmazonDB, OLPostgresDBConfig
-from ol_infrastructure.lib.pulumi_helper import parse_stack
-from pulumi import Config, Output, ResourceOptions, StackReference
-from ol_infrastructure.lib.stack_defaults import defaults
 from ol_infrastructure.components.services.vault import (
     OLVaultDatabaseBackend,
     OLVaultPostgresDatabaseConfig,
 )
-from bridge.lib.magic_numbers import (
-    AWS_RDS_DEFAULT_DATABASE_CAPACITY,
-)
+from ol_infrastructure.lib.pulumi_helper import parse_stack
+from ol_infrastructure.lib.stack_defaults import defaults
+
 stack_info = parse_stack()
 config = pulumi.Config("ol-infrastructure-bootcamps")
 region = config.get("region")
--- a/src/ol_infrastructure/applications/bootcamp/__main__.py
+++ b/src/ol_infrastructure/applications/bootcamp/__main__.py
@@ -1,0 +1,88 @@
+import pulumi
+import pulumi_aws as aws
+from pulumi_aws import Provider
+from ol_infrastructure.components.aws.database import OLAmazonDB, OLPostgresDBConfig
+from ol_infrastructure.lib.pulumi_helper import parse_stack
+from pulumi import Config, Output, ResourceOptions, StackReference
+from ol_infrastructure.lib.stack_defaults import defaults
+from ol_infrastructure.components.services.vault import (
+    OLVaultDatabaseBackend,
+    OLVaultPostgresDatabaseConfig,
+)
+from bridge.lib.magic_numbers import (
+    AWS_RDS_DEFAULT_DATABASE_CAPACITY,
+)
+stack_info = parse_stack()
+config = pulumi.Config("ol-infrastructure-bootcamps")
+region = config.get("region")
+provider = Provider("provider", region=region)
+env = config.get("env")
+app = config.get("app")
+hosted_zone_id = config.get("hosted_zone_id")
+rds_subnet_group_name = config.get("rds_subnet_groups")
+# network_stack = StackReference(f"infrastructure.aws.network.{stack_info.name}")
+# target_vpc_name = config.get("target_vpc") or f"{stack_info.env_prefix}_vpc"
+# target_vpc = network_stack.require_output(target_vpc_name)
+ol_bootcamps_app = aws.s3.Bucket(
+    f"ol_{app}_app_{env}",
+    arn=f"arn:aws:s3:::ol-{app}-app-{env}",
+    bucket=f"ol-{app}-app-{env}",
+    hosted_zone_id=f"{hosted_zone_id}",
+    request_payer="BucketOwner",
+    tags={
+        "Department": f"{app}",
+        "Environment": f"{env}",
+        "OU": f"{app}",
+        "business_unit": f"{app}",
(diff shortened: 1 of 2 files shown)


src/ol_infrastructure/applications/bootcamp/__main__.py
```diff
@@ -0,0 +1,113 @@
...
+        "business_unit": "bootcamps",
+    },
+    versioning=aws.s3.BucketVersioningArgs(
+        enabled=True,
+    ),
+    opts=pulumi.ResourceOptions(protect=True),
+)
+
+
+bootcamp = aws.rds.Instance(
```
blarghmatey, line 72: This should be using the `OLAmazonDB` component resource that we are using everywhere else so that it uses the proper configuration that we are supporting across our RDS instances.

src/ol_infrastructure/applications/bootcamp/__main__.py
```diff
@@ -0,0 +1,113 @@
...
+    },
+    versioning=aws.s3.BucketVersioningArgs(
+        enabled=True,
+    ),
+    opts=pulumi.ResourceOptions(protect=True),
+)
+
+ol_bootcamps_app_rc = aws.s3.Bucket(
+    "ol_bootcamps_app_rc",
+    arn="arn:aws:s3:::ol-bootcamps-app-rc",
```
blarghmatey, line 55: We shouldn't be hardcoding the ARN

src/ol_infrastructure/applications/bootcamp/__main__.py
```diff
@@ -0,0 +1,113 @@
...
+    bucket="ol-bootcamps-app-rc",
+    hosted_zone_id="Z3AQBSTGFYJSTF",
+    request_payer="BucketOwner",
+    tags={
+        "Department": "bootcamps",
+        "Environment": "rc",
```
blarghmatey, line 61: Why are the CI, RC and production buckets being defined (and hardcoded) in the same stack?

blarghmatey (commented):
There should be a stack definition for each environment that the bootcamp app is deployed to (namely CI, QA, and Production). This will also need a Vault policy definition for the AWS resources that it is using and the RDS instance.

src/ol_infrastructure/applications/bootcamp/__main__.py
scoutinhobh: I thought that since we are importing these buckets we can keep  them in the same stack.

src/ol_infrastructure/applications/bootcamp/__main__.py
scoutinhobh: :+1:

src/ol_infrastructure/applications/bootcamp/__main__.py
scoutinhobh: :+1: will add that...did not know that we had a component. Could you add that here please...

scoutinhobh:
"This will also need a Vault policy definition for the AWS resources that it is using and the RDS instance." Why would we need a vault policy? Is the RDS secret being stored in vault?

mitangelo:
"This will also need a Vault policy definition for the AWS resources that it is using and the RDS instance." Why would we need a vault policy? Is the RDS secret being stored in vault?

src/ol_infrastructure/applications/bootcamp/__main__.py
mitangelo: The names should not be hardcoded? How else can we put names for these? I put them in the same stack since they were imported. Can you define the definition of done on these? That way I can get this done neatly instead of doing it and then you commenting on it. I thought all you wanted was an import.

mitangelo:
https://github.com/mitodl/ol-infrastructure/issues/1009

src/ol_infrastructure/applications/bootcamp/__main__.py
mitangelo: ok we can add a format for it.

src/ol_infrastructure/applications/bootcamp/__main__.py
mitangelo: Can you add that in the  issue or post a link here. I did not know that we had this since this is my first time working on a RDS instance.

mitangelo:
@blarghmatey  we have a rds instance of bootcamp for each stack? I was only able to find one rds named bootcamp.

src/ol_infrastructure/applications/bootcamp/__main__.py
mitangelo: Can we add like a prefix based on the env?

src/ol_infrastructure/applications/bootcamp/__main__.py
blarghmatey: The RDS component resource is defined [here](https://github.com/mitodl/ol-infrastructure/blob/main/src/ol_infrastructure/components/aws/database.py) and you can see an example of its use [here](https://github.com/mitodl/ol-infrastructure/blob/main/src/ol_infrastructure/applications/ocw_studio/__main__.py)

blarghmatey:
We currently only have an RDS instance for the production environment. We can keep that as-is and use a conditional of something like `if stack_info.env_suffix == "production"` so that we can manage the buckets and any other resources per-stack and only deploy the database for production.

As to the Vault policy, we need a role defined for the AWS resource(s) that the app uses (likely just S3), as well as the role definition for dynamic database credentials. We have a component resource for that as well which is defined [here](https://github.com/mitodl/ol-infrastructure/blob/main/src/ol_infrastructure/components/services/vault.py#L108) and you can see an example of its use [here](https://github.com/mitodl/ol-infrastructure/blob/main/src/ol_infrastructure/applications/ocw_studio/__main__.py#L248-L256).

blarghmatey:
We don't necessarily want to use the imported RDS instance because it is connected to a VPC that predates a lot of our other work and we would like to consolidate that into the `apps` VPC so that it matches the pattern of our other resources. That can be achieved by creating a new instance and running a dump/load of the data, or migrating the RDS instance to the target VPC as described [here](https://docs.aws.amazon.com/prescriptive-guidance/latest/patterns/migrate-an-amazon-rds-db-instance-to-another-vpc-or-account.html).

mitangelo:
@blarghmatey  can you review once again please.

src/ol_infrastructure/applications/bootcamp/__main__.py
```diff
@@ -0,0 +1,87 @@
...
+region = config.get("region")
+provider = Provider("provider", region=region)
+env = config.get("env")
+app = config.get("app")
+hosted_zone_id = config.get("hosted_zone_id")
+rds_subnet_group_name = config.get("rds_subnet_groups")
+# network_stack = StackReference(f"infrastructure.aws.network.{stack_info.name}")
+# target_vpc_name = config.get("target_vpc") or f"{stack_info.env_prefix}_vpc"
+# target_vpc = network_stack.require_output(target_vpc_name)
+ol_bootcamps_app = aws.s3.Bucket(
```
shaidar, line 25: Bucket should be encrypted

PR #1196, opened 2022-11-03
config: Use Fastly for edX studio to support MFEs

blarghmatey:
The MFE delivery logic is all managed in Fastly. In order for MFEs to work properly in
studio (e.g. course-authoring and library-authoring) we will need to use Fastly as the
frontend for CMS as well as LMS.

## Description
Set the Fastly distribution to serve the Studio hostname as well as LMS

## Motivation and Context
We are starting to deploy MFEs that are used in Studio which means we need to use Fastly for that hostname.

## How Has This Been Tested?
Manually applied the changes in Fastly to validate functionality.

## Types of changes

## Checklist:

### DIFF
--- a/src/ol_infrastructure/applications/edxapp/__main__.py
+++ b/src/ol_infrastructure/applications/edxapp/__main__.py
@@ -1338,6 +1338,17 @@ edxapp_fastly_service = fastly.ServiceVcl(
             first_byte_timeout=60000,  # noqa: WPS432
             between_bytes_timeout=15000,  # noqa: WPS432
         ),
+        fastly.ServiceVclBackendArgs(
+            address=web_lb.dns_name,
+            name="AWS ALB for edX Studio",
+            port=DEFAULT_HTTPS_PORT,
+            ssl_cert_hostname=edxapp_domains["studio"],
+            ssl_sni_hostname=edxapp_domains["studio"],
+            use_ssl=True,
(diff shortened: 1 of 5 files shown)


src/ol_infrastructure/applications/edxapp/__main__.py
```diff
@@ -1338,6 +1338,17 @@ def cloud_init_user_data_func(
             first_byte_timeout=60000,  # noqa: WPS432
             between_bytes_timeout=15000,  # noqa: WPS432
         ),
+        fastly.ServiceVclBackendArgs(
+            address=web_lb.dns_name,
+            name="AWS ALB for edX Studio",
+            port=DEFAULT_HTTPS_PORT,
+            ssl_cert_hostname=edxapp_domains["cms"],
+            ssl_sni_hostname=edxapp_domains["cms"],
```
shaidar, line 1346: Should this be `lms`?

blarghmatey: No, because we need to add CMS as a backend for Fastly so that we can serve studio MFEs through that layer.

blarghmatey: But thank you for pointing that out because I was using the wrong key. Just force-pushed to go from cms -> studio

shaidar (approved):

Ardiea (approved):

PR #1214, opened 2022-11-16
Expire Vault backups out of S3 after 1 year

### LINKED ISSUE #1040: Vault raft-backup lifecycle management
# User Story
We need to be able to limit the amount / number of raft snapshots we store in s3. Right now it is unbounded and that could add up fast (like for QA, each snapshot is 400MB+...) 

We want it to be constrained in two regards:

1. By age. For example, expire a snapshot after 90 days.
2. By Count. For example, retain at least 90 snapshots, IRREGARDLESS of how old they maybe.

This accounts for circumstances where the backups break and stop getting pushed, we wouldn't want the last backup to end up being expired simply because it was old where there is no newer backup. 

Maybe we can accomplish this by simply using the same file name / object key and using bucketing versions and lifecycle expirations? That may be the easiest way to do this. 

# Description/Context

# Acceptance Criteria
- [ ]

Ardiea:
## Description
Expire vault backups after 1 calendar year. 

When used in conjunction with an alert to detect when vault backups are failing, this should close #1040 

## Motivation and Context
There shouldn't really be a need to keep vault backups for more than 1 calendar year. This PR adds an expiration policy to objects in the vault-raft-backup-* buckets that will remove the objects after a year. 

"For version-enabled buckets, Amazon S3 adds a delete marker and the current version of an object is retained as a noncurrent version. **For non-versioned buckets, Amazon S3 permanently removes the object**."

## How Has This Been Tested?
Verified the policy in CI.

## Types of changes

## Checklist:

- [X] My change requires a change to the documentation.

### DIFF
--- a/src/bridge/settings/openedx/types.py
+++ b/src/bridge/settings/openedx/types.py
@@ -8,7 +8,7 @@ class OpenEdxApplication(str, Enum):  # noqa: WPS600
     repository: str
 
     def __new__(cls, app: str, repository: str):
-        enum_element = str.__new__(cls, app)  # noqa: WPS609
+        enum_element = str.__new__(cls, app)
         enum_element._value_ = app  # noqa: WPS437
         enum_element.repository = repository
         return enum_element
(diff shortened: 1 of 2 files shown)


Ardiea:
Verified that healthchecks.io checks are working for raft backups to notify us if backups stop running successfully.

PR #1292, opened 2023-01-03
POC: Instance Refresh jobs in concourse --  

### LINKED ISSUE #692: on-demand Instance refresh job in concourse for workers
Instance refresh pipeline for each class of concourse nodes. 

So in the ocw-team, there is an on-demand job that will kick off an instance refresh of just the ocw-workers, and so on and so on. 

> Should we consider writing a pipeline that will trigger an instance refresh on OCW workers for this kind of situation? We would need to add a well-scoped permission to the instance profile to only allow cycling the specific set of instances.

Ardiea:
## Description
- Added tags to concourse autoscale groups to add in identifiying them.
- Added permissions to infra concourse workers to allow them to trigger an asg instance refresh. Cleaned up some old comments.
- Fixed concourse worker group names in CI to make consistent with other envs.
- Updated the IAM policy for infra workers to allow them to describe all autoscale groups.
- Created a shared function for generating an instance refresh task component. Created a simple pipeline for running instances refreshes against concourse autoscale groups.

## Motivation and Context
This makes an easy-mode trigger for instance refreshes against concourse environments. Mostly a proof of concept that could be expanded to other applications easily. 

Sort of closes #692 ... Didn't set up the workers / teams to allow them to do their own instance refreshes. Just seems excessive. 

## How Has This Been Tested?
Tested in CI. The Stack changes (for the tagging) need to be applied to higher envs before they can be tested there. No reason to think there will be issues. 

## Types of changes

## Checklist:

### DIFF
--- a/src/concourse/lib/tasks.py
+++ b/src/concourse/lib/tasks.py
@@ -1,0 +1,31 @@
+from concourse.lib.constants import REGISTRY_IMAGE
+from concourse.lib.models.pipeline import Command, Identifier, TaskConfig, TaskStep
+
+
+# Generates a TaskStep to perform an instance refresh from a given set of filters and queires.
+# The combination of filters and queries should be trusted to return one, and only one,
+# autoscale group name.
+def instance_refresh_task(
+    filters: str,
+    queries: str,
+) -> TaskStep:
+    return TaskStep(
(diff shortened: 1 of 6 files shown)


blarghmatey (approved):
:+1:

PR #1294, opened 2023-01-06
config: enable certificate display behaviour in studio (MITxOnline)

arslanashraf7:
## Description
We wanted to enable `Certificates Display Behaviour` in our edX studio for MITxOnline. In order to do that need to set up a couple of things(TL;DR A waffle switch and a feature flag/config) mentioned https://github.com/mitodl/mitxonline/issues/1257#issuecomment-1346463078.
https://github.com/mitodl/mitxonline/issues/1257

## Motivation and Context
Part of https://github.com/mitodl/mitxonline/issues/1257
This is for enabling certificate display behavior in edX studio.

## How Has This Been Tested?
I tested it with:
1. Added a waffle switch `certificates.auto_certificate_generation` through Django admin.
2. Enabled the `ENABLE_V2_CERT_DISPLAY_SETTINGS` in your `FEATURES` for cms (`private.py` or `cms.yml`).
3. Open a course in Studio > Settings > Schedule and Details > Notice that you can now see the display behavior settings as below 
image

## Checklist:


pdpinch:
I am concerned about the potential for side-effects from this setting. We'll want to test it carefully on QA/RC before setting it in production.

pdpinch:
wow, that is a horrible name for a waffle flag and the documentation for it is even worse. 

from https://edx.readthedocs.io/projects/edx-platform-technical/en/latest/featuretoggles.html#featuretoggle-FEATURES['ENABLE_V2_CERT_DISPLAY_SETTINGS']

all it says: 

> Whether to use the reimagined certificates_display_behavior and certificate_available_date

src/bilder/images/edxapp/templates/edxapp/mitxonline/studio_only.yml.tmpl
```diff
@@ -2,6 +2,8 @@
 ADDL_INSTALLED_APPS:  # ADDED KEY
   - git_auto_export
 COURSE_AUTHORING_MICROFRONTEND_URL: /course-authoring
+FEATURES:
+    ENABLE_V2_CERT_DISPLAY_SETTINGS: true # ADDED KEY
```
blarghmatey, line 6: This will need to go in the `common_settings` file because it already has a `FEATURES` dict and is shared between LMS and studio. Adding this in the studio config will generate invalid YAML since it will be redefining the name further down in the YAML file. The way these files are used is that `common` and `studio` templates get concatenated and then the unified template is rendered. Same for `common` and `lms` templates.

arslanashraf7: Ah, @blarghmatey thanks for the clarification. I thought about adding it to common first but wasn't clear about that with something that's only for Studio. Now, this makes sense. I'll change this.

arslanashraf7: I've updated the PR.

Issue #1256, opened 2022-12-08
Log and metrics tracking for Docker services


blarghmatey:
# User Story
- As a platform engineer I need visibility into the services that I run via logs and metrics

# Description/Context
As we migrate more of our services to run via docker-compose on EC2 instances we need a standardized way of collecting logs and metrics. We use Grafana as the storage/query/visualization layer for those metrics and Vector as the collector. We need a solid default configuration for Vector that collects information that aids us in monitoring, alerting, and debugging.

# Acceptance Criteria
- [ ] Vector collects the logs from the running containers and sends them to Grafana's installation of Loki associated with the running instance (e.g. CI -> CI, etc.)
- [ ] Vector collects metrics for the running containers and sends them to Grafana's installation of Prometheus. One option is to use [cAdvisor](https://github.com/google/cadvisor) for exposing those metrics via a Prometheus endpoint.
- [ ] Vector has a default configuration snippet managed via Vault agent templates or consul-template that manages the credentials for Grafana so that we don't have to pass them through EC2 userdata

# References
Some interesting resources that I came across when looking at cAdvisor for use with Grafana:
- https://github.com/vegasbrianc/grafana_dashboard
- https://github.com/vegasbrianc/docker-traefik-prometheus
- https://github.com/vegasbrianc/prometheus

Ardiea:
Partially addressed here:
https://github.com/mitodl/ol-infrastructure/pull/1281
https://github.com/mitodl/ol-infrastructure/pull/1280

blarghmatey:
@Ardiea can you weigh on on the status of this and whether we can consider it complete? If it's not complete, should we create a new issue to capture any remaining work, or just keep this one open?

Ardiea:
Log collection is greatly improved. There are no metric collections to speak of and I've not even considered it. So I can close this and open a new one for investigating that. I think there is cadvisor or something?

Issue #899, opened 2022-06-14
Support our installation of Airbyte - Phase 2 


Ardiea:
Tech Debt and Hold-over tasks from the initial build out. Subject to change:

- [ ] Update load balancer config to hit an airbyte health check endpoint directly (via nginx/caddy, bypass pomerium)
- [x] Concourse job to make the AMI (follow standard pattern)
- [x] S3 Bucket access added to the instance policy
- [x] [OPTIONAL] Cull the db service from docker-compose file. 
- [x] Finish integration of AIRBYTE_VERSION variable through to the AMI pipeline.
- [ ] Make the RDS config more secure
  - [ ] Turn force TLS back on, (could be tricky)
  - [ ] Get the app to run as just the app, not admin.
  - `temporal` introduces complications here
- [ ] Allow access on port 443 from concourse using basic auth by adding a secgroup in addition to using lets encrypt and adding `airbyte.service.consul` as a SAN
- [ ] Observability stuff
  - [ ] Metrics endpoints for Pomerium?
  - [ ] Metrics endpoints for Airbyte?
  - [x] Logs from Pomerium?
  - [x] Logs from airbyte? 
- [ ] https://github.com/mitodl/ol-infrastructure/pull/866/files#r897307555 
- [x] https://github.com/mitodl/ol-infrastructure/pull/866/#discussion_r897314070

PR #1328, opened 2023-01-26
Docker log collection overhaul

Ardiea:
## Description
Refactor on how logging configuration is deployed for most docker based applications. Vector installation and configuration from the docker_baseline_ami build process. We will be better served moving this out into a more app specific location.

## Motivation and Context
While specifying a generic log collection configuration in the docker baseline AMI is definitely easier and results in less configuration in fewer places, it doesn't lend itself well to classifying the logs correctly when they are shipped to loki.

Relates to #1256 #899 

## How Has This Been Tested?
TBD Need to verify these all build AMIs that work first and then verify logs are shipped in from CI/RC envs as needed

## Types of changes

## Checklist:


blarghmatey:
Can we look at splitting these configs into separate files/snippets so that they are more composable and reduces duplication? That could also allow us to move the management of the Grafana credentials into a shared vault or consul template that gets laid down in the baseline AMI.

Ardiea:
Yeah I will experiment with that. That would be a good compromise between the two paths.

Ardiea:
@blarghmatey Yes that does work but it will require some careful thought, since each subsequent step is dependent on the one before it and you need to know the name of the one before it. So probably standardize on 'enrich_logs' as the last transform in ALL log pipelines, and then in the second file that is deployed everywhere with the sink is is configured with the one input 'enrich_logs'. It is kind of tricky but I think we can make it work.

src/bilder/images/dagster/templates/vector/dagster_logs.yaml
```diff
@@ -0,0 +1,14 @@
+---
+transforms:
+  parse_dagster_logs:
+    type: remap
+    inputs:
+    - "parse_docker_stdout_stderr"
+    source: |
+      .application = "dagster"
```
blarghmatey, line 8: This is probably going to require another pass to be able to pull in logs from the different pipeline containers. We'll likely want to apply a common label to all of them so that we can set the proper source filtering.

src/bilder/images/dagster/deploy.py
```diff
@@ -267,11 +268,17 @@
 for product in hashicorp_products:
     configure_hashicorp_product(product)
 
+# Install and configure vector
+vector_config = VectorConfig(is_docker=True, use_global_log_sink=True)
+vector_config.configuration_templates[
+    TEMPLATES_DIRECTORY.joinpath("vector", "dagster_logs.yaml")
+] = {}
+install_vector(vector_config)
+configure_vector(vector_config)
```
blarghmatey, line 277: Should we wrap all of this as a single deploy function to cut down on boilerplate? (not right now, but as a TODO)

src/bilder/images/dagster/templates/vector/dagster_logs.yaml
Ardiea: So actually under service it will be the name of the docker compose service as written in docker-compose.yaml for example:

```
{
  "application": "dagster",
  "service": "lakehouse-assets-graph",
  "environment": "data-production"
  ...
```

and 
```
{
  "application": "dagster",
  "service": "edx-gcs-courses-pipeline",
  "environment": "data-production"
  ...
```
https://github.com/mitodl/ol-infrastructure/blob/main/src/bilder/images/dagster/templates/docker-compose.yaml.j2

Ardiea: https://github.com/mitodl/ol-infrastructure/pull/1328/files#diff-4869c0991783135c961f7f43ce25ca110148cc7c7fa2c621528c17036e3f43efR16

src/bilder/components/vector/templates/global_metric_sink.yaml
```diff
@@ -0,0 +1,13 @@
...
+    inputs:
+    # Expects the final step of ALL metric pipelines to be named 'noop_metric_global_funnel'
+    - 'noop_metric_global_funnel'
+    type: prometheus_remote_write
+    endpoint: https://prometheus-prod-10-prod-us-central-0.grafana.net/api/prom/push
+    healthcheck: false
+    auth:
+      strategy: basic
+      user: "${GRAFANA_CLOUD_PROMETHEUS_API_USER}"
+      password: ${GRAFANA_CLOUD_API_KEY}
```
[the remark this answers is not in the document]
Ardiea: I think it might be better to convert EVERYTHING to use these shared sinks before converting them to consul/vault templates. There is a validate step during install+config that will break with switching to templates and having both co-existing will be complicated, I think.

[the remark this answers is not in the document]
blarghmatey: What I mean is to use the vault/consul templates to write the values into `/etc/default/vector` and leave the global sink yaml files as-is

blarghmatey (approved):

src/bilder/components/vector/templates/global_metric_sink.yaml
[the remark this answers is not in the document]
Ardiea: I've been poking at this for a bit now and it keeps getting larger and larger in scope and consequences so I think I'm going to keep it for another day. We will need an `infrastructure` stack to store the grafana-cloud secrets into vault. Getting the `datacenter` is not entirely trivial, which is going to be used for 'environment' in the final log / metric document shipped off to Grafana. There are a few other things as well.

PR #1342, opened 2023-02-02
Updated dockerhub creds to use token

shaidar:
## Description

This updates the dockerhub password to use a separate token for QA and production. Looking at [docker docs](https://docs.docker.com/docker-hub/access-tokens/) it appears that we can use the same username but the token as the password. Additionally, this removes dockerhub references from old files.
## Motivation and Context

General updates and cleanup.
## How Has This Been Tested?

I tried using a token to login locally and it seems to work, but the ultimate test would be to deploy on QA and make sure the pipelines are still functional.
## Types of changes

## Checklist:


Ardiea:
> I tried using a token to login locally and it seems to work, but the ultimate test would be to deploy on QA and make sure the pipelines are still functional.

I honestly don't know the status of the QA concourse these days and if there is a valid docker-build job you could expect to be working there. May just need to deploy to prod 🤷 It is nothing more than a pulumi stack update so it is quick to revert.

PR #1368, opened 2023-02-15
Fastly Logging to s3 for OCW

### LINKED ISSUE #1224: Enable fastly logging for OCW
# User Story
- We need to enable logging on Fastly to get a sense of how much usage is the download course content feature is getting.

# Description/Context

For now, it's good enough to simply enable logging and collect the data in S3 for later analysis.

# Acceptance Criteria
- [ ]

Ardiea:
## Description
This PR sets up an S3 bucket for receiving logs from fastly as well as configures IAM resources that will be needed to configure an s3 logging destination within a fastly config. Finally it then configures the OCW fastly code to send access log data to the aformentioned bucket.

## Motivation and Context
Closes #1224 

## How Has This Been Tested?
Manually

## Types of changes

## Checklist:


src/ol_infrastructure/infrastructure/monitoring/__main__.py
```diff
@@ -0,0 +1,84 @@
...
+        {
+            "Effect": "Allow",
+            "Action": [
+                "s3:PutObject",
+            ],
+            "Resource": f"arn:aws:s3:::{fastly_logging_bucket_name}/*",
+        }
+    ],
+}
+fastly_logging_iam_policy_name = "ol-fast-access-logs-policy"
```
blarghmatey, line 44:
```suggestion
fastly_logging_iam_policy_name = "ol-fastly-access-logs-policy"
```

src/ol_infrastructure/infrastructure/monitoring/__main__.py
```diff
@@ -0,0 +1,84 @@
...
+        "Version": IAM_POLICY_VERSION,
+        "Statement": {
+            "Condition": {
+                "StringEquals": {
+                    "sts:ExternalId": monitoring_config.get("fastly_customer_id")
+                },
+            },
+            "Effect": "Allow",
+            "Action": "sts:AssumeRole",
+            "Principal": {"AWS": "717331877981"},
```
shaidar, line 65: Is this right?

Ardiea: Yes.

Ardiea: And to elaborate, that is fastly AWS acct # from their config. We allow their AWS account to assume this role and then via the attached policy (which isn't actually attached yet in this commit), they can write (and only write, no read) to our S3 bucket.

Ardiea:
So I've tested this vs QA and everything seems to be working as expected. One thing that may need a closer look is the permissions on the s3 bucket. I **think** I have it correct but I'm not sure. I have confirmed that you can't access any of the files without authentication.

src/ol_infrastructure/applications/ocw_site/__main__.py
```diff
@@ -455,6 +464,15 @@
                 type="error",
             ),
         ],
+        logging_s3s=[
+            fastly.ServiceVclLoggingS3Args(
+                bucket_name=fastly_access_logging_bucket["bucket_name"],
+                name=f"ocw-{purpose}-{stack_info.env_suffix}-logging-args",
+                path=f"/ocw/{purpose}/{stack_info.env_suffix}/",
```
blarghmatey, line 471: Where is the `purpose` coming from?

Ardiea: line 260 02b4e324687890ce84c5ac4b69ecca040da35c50c91f9ba8e9c290ee6c21774cR260

"live" or "draft", a distinction only important for OCW I believe.

blarghmatey: I'd say probably put `env_suffix` as the second tier rather than 3rd

src/ol_infrastructure/lib/fastly.py
```diff
@@ -24,3 +24,23 @@ def get_fastly_provider(
...
+    "request_method": "%{json.escape(req.method)}V",
+    "request_protocol": "%{json.escape(req.proto)}V",
+    "request_referer": "%{json.escape(req.http.referer)}V",
+    "request_user_agent": "%{json.escape(req.http.User-Agent)}V",
+    "response_state": "%{json.escape(fastly_info.state)}V",
+    "response_status": %{resp.status}V,
+    "response_reason": %{if(resp.response, "%22"+json.escape(resp.response)+"%22", "null")}V,
+    "response_body_size": %{resp.body_bytes_written}V,
+    "fastly_server": "%{json.escape(server.identity)}V",
+    "fastly_is_edge": %{if(fastly.ff.visits_this_service == 0, "true", "false")}V
```
blarghmatey, line 45: Is it possible to get any information from Fastly on response time?

Ardiea: oh let me look, I just stole the default one from manually setting it up in the UI.

blarghmatey: So, we actually have S3 logging set up on the legacy OCW distribution and the format string that was used there is:
```
{"timestamp": "%{%Y-%m-%dT%H:%M:%S+00:00}t", "client": "%h", "method": "%m", "request": "%{json.escape(req.url)}V", "status": "%>s", "header_bytes_sent": %{resp.header_bytes_written}V, "body_bytes_sent": %{resp.body_bytes_written}V, "referer": "%{json.escape(req.http.Referer)}V", "user_agent": "%{json.escape(req.http.User-Agent)}V", "request_time": %{time.elapsed}V, "time_to_first_byte": %{time.to_first_byte}V, "cache_status": "%{regsub(fastly_info.state, "^(HIT-(SYNTH)|(HITPASS|HIT|MISS|PASS|ERROR|PIPE)).*", "\\2\\3") }V", "city": "%{client.geo.city.utf8}V", "region": "%{client.geo.region.utf8}V", "country": "%{client.geo.country_name.utf8}V", "country_code": "%{client.geo.country_code}V", "continent_code": "%{client.geo.continent_code}V", "client_conn_type": "%{client.geo.conn_type}V", "client_geo_conn_speed": "%{client.geo.conn_speed}V"}
```
There are some interesting bits in there, so it's probably worth melding the two format strings to be a superset of all of this information so we can capture as much as possible.

blarghmatey: In general I like the field naming that you used for the attributes (e.g. `request_`, `response_`, `geo_`, etc. to make it clear what each attribute pertains to)

blarghmatey: Looks great to me

blarghmatey (approved):

src/ol_infrastructure/lib/fastly.py
Ardiea: Oh follow up question: would it hurt/hinder to have these compressed when they show up? I suspect it is going to be A LOT of data from production (60-100 rps). I can set a flag and they will show up gzipped instead of just raw json but I feel like that might add a complication to whatever future consumer of the data.

blarghmatey: GZIP is fine. The likely consumer is probably Trino which can handle compression without issue.

PR #1443, opened 2023-04-05
refactor: Clean up Fargate model and add attribute details

blarghmatey:
The AWS API has a lot of detail that is needed to use it effectively. This adds well-documented `Field` objects that use the contents of the relevant AWS documentation for those attributes. It also cleans up the log and container definitions to be easier to use.

## Description
This was done in an attempt to make this component more usable if/when we end up needing it.

## Motivation and Context
This code was contributed as part of an experimental approach to deployments, but hasn't been used. The cleanup was opportunistic but would be good to merge rather than deleting or letting languish in a branch.

## How Has This Been Tested?
N/A

## Types of changes

## Checklist:


feoh (approved):
LGTM 👍

PR #1447, opened 2023-04-07
Md/issue 1414

### LINKED ISSUE #1414: Send Fastly logs into Grafana for debugging and alerting
# User Story
- As a platform engineer I would like to be alerted when a Fastly property is experiencing a spike of error conditions
- As an application engineer I would like to be able to diagnose errors all the way from Fastly through to the backend system

# Description/Context
We need a path for sending Fastly logs into a data source connected to Grafana. The most likely destination is to use Vector as a router for proxying log data into the Grana cloud managed Loki storage system.

# Acceptance Criteria
- [ ] Fastly properties all log as JSON to S3 with a standard message schema
- [ ] Fastly logs are sent to Loki via a Vector proxy using HTTPS or ~Syslog inputs~ absolutely not. 
- [ ] Vector rejects log data generated by bot traffic as identified by user agent information
- [ ] Fastly logs are available in grafana via the established Loki datasources

Ardiea:
## Description
This adds configuration to the OCW fastly configurations to ship logs to Grafana via our existing vector log proxy services. It adds the fastly config changes as well as needed configuration changes to vector. It installs traefik infront of vector to route ownership challnege requests to an S3 bucket associated with each vector-log-proxy that contains the required response that fastly is looking for. 

## Motivation and Context
Closes #1414 

## How Has This Been Tested?
Tested in QA

## Types of changes

## Checklist:


src/ol_infrastructure/lib/fastly.py
```diff
@@ -6,6 +6,44 @@
...
+"timestamp":"%{strftime(\\{"%Y-%m-%dT%H:%M:%S%z"\\},time.start)}V",
+"url":"%{json.escape(req.url)}V"
+}"""
+
+
+# A fastly logformat string isn't actually json and we can't treat it as such in code
+# but we do need it to *produce* valid and minimized json at the end of the day once
+# it is installed into the fastly log configurations.
+def build_fastly_log_format_string(additional_static_fields: dict[str, str]) -> str:
+    split_format_string = __base_fastly_log_format_string.split("\n")
```
blarghmatey, line 42: Is this because it's not possible to treat the format string as valid JSON due to the special characters?

Ardiea: Yeah I tried changing this to just a regular python-sh data structure and then using json.dumps but fastly refused to accept it as a valid logformat string. I didn't want to spend too much energy on it because I figured it didn't hurt to much to have it as just one big string above and the do some cleverness in this function to add additional fields while keeping the format string still producing valid json.

Issue #1332, opened 2023-01-26
Manage Keycloak configuration objects via Pulumi


blarghmatey:
# User Story
As a platform engineer I want to manage the configuration of our Keycloak service as code

# Description/Context
Keycloak has an available Pulumi plugin for managing the different configuration objects (e.g. realms, integrations, etc.). We want to use this to keep our different Keycloak installations aligned with each other so that they don't suffer configuration drift through overuse of "click-ops".

# Acceptance Criteria
- [ ] There is one or more Pulumi projects in the substructure namespace that are used to manage the relevant Keycloak configurations as code.
- [ ] The Pulumi projects are executed as part of a Concourse pipeline.

# Functional Requirements
- [ ] Determine appropriate groupings and scope of realm definitions and implement code to manage them
  - [ ] Operations (e.g. Vault, Consul, maybe something like Boundary?)
  - [ ] Data (e.g. Airbyte, Dagster, Starburst, etc.)
  - [ ] Applications (e.g. Micromasters, MITx Online, OCW, OVS, etc.)
- [ ] Integrate with identity providers
  - [ ] Touchstone
  - [ ] GitHub
  - [ ] LDAP?
- [ ] Determine proper roles and attributes for different realms and implement code to manage them

shaidar:
The keycloack configs are in this gist https://gist.github.com/collinpreston/b13af969bd90ff166df7f4624678a069. Some of the settings are loose given that it's a POC and will need to be tweaked prior to rolling out to QA and Prod. Started going through those configs and finding them in [Pulumi](https://www.pulumi.com/registry/packages/keycloak/api-docs/). As always, the docs (keycloak) require some fishing around to find out how they can be configured.

shaidar:
Setup a local environment with Keycloak and Vault to test out auth in Vault using Keycloak. Added a new Realm separate from the master realm that is created as part of the Keycloak deployment. Created a few users with different permission in vault and I was able to login to vault using Keycloak. Ran into a minor issue regarding the keycloak pop-up that was blocked by the browser but didn't see a warning or anything hinting to that, so might look at some options to see about that, but not a major issue for now.
Today, will be setting up another realm and testing different permissions and then writing a Pulumi PR for a basic Keycloak config with multiple realms.

shaidar:
[...]

PR #1460, opened 2023-04-13
Keycloak realm config

shaidar:
## Description

Creates a new realm in Keycloak that platform engineering will be using for their apps along with some best practices to lock things down as an initial run.
## Motivation and Context

https://github.com/mitodl/ol-infrastructure/issues/1332
## How Has This Been Tested?

Ran it locally and made expected changes on keycloak-qa instance.
## Types of changes

## Checklist:


src/ol_infrastructure/substructure/keycloak/__main__.py
```diff
@@ -0,0 +1,133 @@
...
+    "update_password",
+    realm_id=ol_platform_engineering_realm.realm,
+    alias="UPDATE_PASSWORD",
+    default_action=True,
+    enabled=True,
+    opts=resource_options,
+)
+
+github_oauth_client_id = keycloak_config.require("github_oauth_client_id")
+github_oauth_client_secret = keycloak_config.require("github_oauth_client_secret")
```
blarghmatey, line 121: In a second pass it might be good to generate the OAuth app in the Pulumi code so that we don't have to manage that as a separate manual step.

src/ol_infrastructure/substructure/keycloak/__main__.py
```diff
@@ -0,0 +1,133 @@
...
+
+realm_identity_provider = keycloak.oidc.IdentityProvider(
+    "github-idp",
+    realm=ol_platform_engineering_realm.id,
+    alias="github-idp",
+    display_name="GitHub",
+    client_id=github_oauth_client_id,
+    client_secret=github_oauth_client_secret,
+    authorization_url=f"https://{env_name}.odl.mit.edu/auth/realm/ol-platform-engineering/github/endpoint",
```
blarghmatey, line 130: It looks like this will result in a domain of `operations-ci.odl.mit.edu`. Is that the domain of the instance? This is probably a good spot to pull a stack output from the infrastructure layer to avoid any stringly typed errors with a mismatch.

blarghmatey (commented):
Only the one question about the domain in the GitHub provider block, otherwise this looks great!

shaidar:
@blarghmatey This should be ready to merge. Can you please take another quick look.

blarghmatey (approved):
Looks great!

PR #1511, opened 2023-05-09
add ENABLE_NEW_BULK_EMAIL_EXPERIENCE ...

pdpinch:
... and set it to true

## Description

Enable the Communications MFE, frontend-app-communications, based on the documentation at https://edx.readthedocs.io/projects/edx-platform-technical/en/latest/featuretoggles.html#featuretoggle-FEATURES['ENABLE_NEW_BULK_EMAIL_EXPERIENCE']

This PR adds the settings to Residential MITx and MITx Online. We would like the feature for both of those sites. 

If all goes well, we can include it in xPRO with the next release (presumably Quince in December 2023).

## Motivation and Context

The MFE allows instructors to schedule the sending of bulk emails. 

## How Has This Been Tested?

This hasn't been tested at all. 

## Types of changes


blarghmatey:
It looks like we also need to set the value for `COMMUNICATIONS_MICROFRONTEND_URL` to allow for routing it properly. The value should be `communications` as defined here: https://github.com/mitodl/ol-infrastructure/blob/main/src/bridge/settings/openedx/types.py#L39

blarghmatey:
That's based on https://github.com/openedx/edx-platform/blob/9036c0613fcee7a806490b885388403b5bcca04c/lms/djangoapps/instructor/views/instructor_dashboard.py#L705

blarghmatey (approved):

PR #1537, opened 2023-05-26
Add framework tagging to edxapp AMI builds and deploys 

Ardiea:
Adding a framework tag to edxapp images to differentiate between old ansible AMIs and new docker AMIs. Practice safe deployments.

N/A

While we work on moving edxapp into docker, it is possible for an AMI built around docker to sneak its way through into environments where it is not welcome should a `pulumi up` be run against an env without a corresponding `packer build` for the same env. This is pretty unlikely to happen since we mostly depend on the concourse pipeliens to do the needful at the moment but it is still a possibility. 

This PR does two things: 

1. Adds a tag to the legacy AMIs 'framework: ansible`
2. Adds an AMI filter to the pulumi stacks that defaults to requiring `framework: ansible` unless a configuration option of `edxapp:use_docker: True` is set for an environment.

### DIFF
--- a/src/bilder/images/edxapp/custom_install.pkr.hcl
+++ b/src/bilder/images/edxapp/custom_install.pkr.hcl
@@ -83,6 +83,7 @@ source "amazon-ebs" "edxapp" {
     OU              = var.business_unit
     app             = local.app_name
     deployment      = var.installation_target
+    framework       = "ansible"
     purpose         = "${local.app_name}-${var.node_type}"
     openedx_release = var.openedx_release
   }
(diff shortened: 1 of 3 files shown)


src/ol_infrastructure/applications/edxapp/__main__.py
```diff
@@ -120,11 +120,14 @@
 edxapp_vpc_id = edxapp_vpc["id"]
 data_vpc = network_stack.require_output("data_vpc")
 data_integrator_secgroup = data_vpc["security_groups"]["integrator"]
+
+framework = "docker" if edxapp_config.get("use_docker") else "ansible"
```
blarghmatey, line 124: Maybe make that `get_bool`?

src/bilder/images/edxapp/custom_install.pkr.hcl
```diff
@@ -83,6 +83,7 @@ source "amazon-ebs" "edxapp" {
     OU              = var.business_unit
     app             = local.app_name
     deployment      = var.installation_target
+    framework       = "ansible"
```
blarghmatey, line 86: Probably worth adding this to the `base` template as well just for belt and suspenders.

blarghmatey (approved):

PR #1544, opened 2023-05-31
Put bootcamps-ecommerce under Pulumi management

### LINKED ISSUE #1009: Migrate management of Bootcamps resources into Pulumi

feoh:

Fixes #1009 

Puts the bootcamps-ecommerce app under Pulumi resource management and creates the relevant RDS DB, S3 bucket,
and VPCs.

# How can this be tested?

Bootcamps-Ecommerce CI environment was flipped over to the new Pulumi managed RDS DB with no discernable ill effects.

Further testing in RC is needed before rolling to production.


feoh:
I have somehow once again wound up in git hell and managed to revert the contents of my change to a prior commit.

I will untangle this mess and get the right revisions in place... Tomorrow.

feoh:
Now having rescued myself from a mess of my own making, I think this is looking good :)

src/ol_infrastructure/applications/bootcamps_ecommerce/__main__.py
```diff
@@ -0,0 +1,203 @@
...
+)
+
+
+bootcamps_iam_policy = iam.Policy(
+    f"bootcamps-{stack_info.env_suffix}-policy",
+    description="AWS access controls for the bootcamps application in the "
+    f"{stack_info.name} environment",
+    path=f"/ol-applications/bootcamps/{stack_info.env_suffix}/",
+    name_prefix="aws-permissions-",
+    policy=lint_iam_policy(
```
feoh, line 75: Please pay careful attention to this IAM rule. The bucket already exists so will remain unchanged for now, but in a disaster recovery scenario, would these be the right perms with which to re-create?

Is there more testing I can do to validate my assertions here? I mostly copied/pasted from OCW at @blarghmatey's suggestion.

src/ol_infrastructure/applications/bootcamps_ecommerce/__main__.py
```diff
@@ -0,0 +1,203 @@
...
+                    ],
+                    "Resource": [
+                        f"arn:aws:s3:::{bootcamps_storage_bucket_name}",
+                        f"arn:aws:s3:::{bootcamps_storage_bucket_name}/*",
+                    ],
+                },
+                {
+                    "Effect": "Allow",
+                    "Action": ["execute-api:Invoke", "execute-api:ManageConnections"],
+                    "Resource": "arn:aws:execute-api:*:*:*",
```
shaidar, line 101: What is this used for in this context?

src/ol_infrastructure/applications/bootcamps_ecommerce/__main__.py
```diff
@@ -0,0 +1,203 @@
...
+            description="Allow access over the public internet from Heroku",
+        ),
+        ec2.SecurityGroupIngressArgs(
+            protocol="tcp",
+            from_port=DEFAULT_POSTGRES_PORT,
+            to_port=DEFAULT_POSTGRES_PORT,
+            security_groups=[data_vpc["security_groups"]["integrator"]],
+        ),
+    ],
+    egress=[
```
shaidar, line 143: Should we look into tightening things and if that's the case, is this egress rule necessary? My understanding is that sec groups are stateful and since the Heroku app will be initiating the connection, there's no need for an egress rule. Think it's worth trying out and verifying.

shaidar:
Just a nitpick, but we're mainly calling the application `bootcamps` so should be stick with that number across the board and get rid of the `ecommerce` bit or should be stick with `bootcamps_ecommerce` across the board. Don't mind it either way, but think it would be good to stick with one.

src/ol_infrastructure/applications/bootcamps_ecommerce/__main__.py
blarghmatey: Yeah, I'm unclear what the execute-api permissions would be used for.

blarghmatey: Yeah, I'm unclear what the execute-api permissions would be used for.

src/ol_infrastructure/applications/bootcamps_ecommerce/__main__.py
```diff
@@ -0,0 +1,203 @@
...
+        ec2.SecurityGroupEgressArgs(
+            from_port=0,
+            to_port=0,
+            protocol="-1",
+            cidr_blocks=["0.0.0.0/0"],
+            ipv6_cidr_blocks=["::/0"],
+        )
+    ],
+    tags=aws_config.merged_tags(
+        {"name": "bootcamps-db-access-applications-{stack_info.env_suffix}"}
```
blarghmatey, line 153: This is missing the `f` prefix for the string to actually interpolate the value

src/ol_infrastructure/applications/bootcamps_ecommerce/__main__.py
```diff
@@ -0,0 +1,203 @@
...
+    tags=aws_config.tags,
+    db_name="bootcamps",
+    public_access=True,
+    **defaults(stack_info)["rds"],
+)
+bootcamps_db = OLAmazonDB(bootcamps_db_config)
+
+bootcamps_vault_backend_config = OLVaultPostgresDatabaseConfig(
+    db_name=bootcamps_db_config.db_name,
+    mount_point=f"{bootcamps_db_config.engine}-bootcamps-applications-{stack_info.env_suffix}",  # noqa: E501
```
blarghmatey, line 172: Because we have a different Vault cluster per env stage it's actually preferred to leave the env info out of the mount point. This should just be `f"{bootcamps_db_config.engine}-bootcamps"`

src/ol_infrastructure/applications/bootcamps_ecommerce/__main__.py
feoh: Good catch! Nuking it. 💥

src/ol_infrastructure/applications/bootcamps/__main__.py
```diff
@@ -0,0 +1,189 @@
...
+                "ignore_locations": [{"actions": ["s3:putobjectacl"]}]
+            },
+            "RESOURCE_EFFECTIVELY_STAR": {},
+        },
+    ),
+)
+
+bootcamps_vault_backend_role = vault.aws.SecretBackendRole(
+    f"bootcamps-app-{stack_info.env_suffix}",
+    name=f"bootcamps-app-{stack_info.env_suffix}",
```
blarghmatey, line 112: The role doesn't need to have the env information in it because of the fact that we have distinct Vault clusters per env-stage

src/ol_infrastructure/applications/bootcamps/__main__.py
```diff
@@ -0,0 +1,189 @@
...
+)
+
+vault_secrets = read_yaml_secrets(
+    Path(f"bootcamps/bootcamps.{stack_info.env_suffix}.yaml")
+)
+vault.generic.Secret(
+    "bootcamps-vault-secrets",
+    path=bootcamps_secrets.path.apply("{}/app-config".format),
+    data_json=json.dumps(vault_secrets),
+)
```
blarghmatey, line 183: Until we get these secrets into SOPS these lines will need to be removed.

PR #1545, opened 2023-05-31
config: enable email scheduling in communications MFE (MITxOnline)

arslanashraf7:

https://github.com/mitodl/hq/issues/1503

- Adds a config setting for Communications MFE 
- The setting enables the `Schedule Email` feature in the MFE

image

# How can this be tested?
- I've only tested it locally by setting up the communications MFE. For more details take a look at 
image
- To set up the MFE locally, You can also take a look at feature toggles docs [here](https://edx.readthedocs.io/projects/edx-platform-technical/en/latest/featuretoggles.html#featuretoggle-FEATURES['ENABLE_NEW_BULK_EMAIL_EXPERIENCE']) 
- The schedule email functionality can be enabled with https://github.com/openedx/frontend-app-communications/blob/4493eb71fa4fdaef3f5c541cff2d9f95d91ec14c/.env.development#L22

# Additional Context

- To see the working of Communications MFE, please take a look at https://github.com/openedx/frontend-app-communications/blob/4493eb71fa4fdaef3f5c541cff2d9f95d91ec14c/.env.development#L22 
- In this PR I'm only enabling this feat for the QA instance of MITxOnline, Once we test it there I'll create a followup PR to enable it in production.


asadali145 (approved):
LGTM!

blarghmatey (approved):

src/concourse/pipelines/open_edx/mfe/values.py
```diff
@@ -97,6 +97,7 @@
         site_name="MITx Online QA",
         studio_domain="studio-qa.mitxonline.mit.edu",
         support_url="mitxonline.zendesk.com/hc/",
+        schedule_email_section="true",  # Because the communication MFE treats this boolean as string  # noqa: E501
```
blarghmatey, line 100: It's probably worth copying this to the production block as well (line 120)

arslanashraf7: I intentionally kept its scope to QA so that we can test this first and then create a follow-up PR and add this to Prod. But let's hear @pdpinch 's thoughts on this. 

@pdpinch Should we enable it for both QA and Prod?

pdpinch:
I think doing both at the same time is OK in this case. 

The feature is additive. And we would review it on the QA stack before deploying the production stack anyway.

arslanashraf7:
@blarghmatey I've enabled it for production instance as well in https://github.com/mitodl/ol-infrastructure/pull/1545/commits/ad8c69e4ff71dfa25e5b058475d946a1c1e96555. Could you take another look and give a thumbs up if it looks good to you?

blarghmatey (approved):
:+1:

blarghmatey:
The pipeline update failed to apply, but this is pulled in now with the changes that I did to the communications MFE for fixing its link handling. Let me know if you see the expected changes.

arslanashraf7:
> The pipeline update failed to apply, but this is pulled in now with the changes that I did to the communications MFE for fixing its link handling. Let me know if you see the expected changes.

Looking at https://courses-qa.mitxonline.mit.edu/communications/courses/course-v1:MITx+Template+11_15_2022/bulk_email, Looks like it didn't show the expected changes. For reference we were looking for [this config](https://github.com/openedx/frontend-app-communications/blob/4493eb71fa4fdaef3f5c541cff2d9f95d91ec14c/.env#L21) to have a value(`true`) in the communications MFE.

Issue #1473, opened 2023-04-21
Integrate Keycloak with Traefik via ForwardAuth


blarghmatey:
# Description/Context
We are standardizing on Traefik as our proxy for all services. As part of that we need a way to add authentication and authorization to applications/services that don't include that capability (e.g. Dagster or Airbyte). Traefik supports a middleware called ForwardAuth that allows it to delegate to systems like KeyCloak for this purpose. We would like to have a reusable component that easily sets up ForwardAuth via Keycloak for any system being proxied through Traefik.

# Acceptance Criteria
- [x] Traefik is configured to use ForwardAuth with Keycloak as the authentication provider

# References
- https://doc.traefik.io/traefik/middlewares/http/forwardauth/
- https://stackoverflow.com/questions/75168350/traefik-forward-auth-middleware-for-openid-connect
- https://geek-cookbook.funkypenguin.co.nz/docker-swarm/traefik-forward-auth/

shaidar:
As the issue describes, the idea is have a request be proxied through traefik, make its way to the forward-auth middleware and be sent to keycloak and then return to traefik and be forwarded to the app listening on the requested URL. 

#### Initial trial
Tried getting this to work with Airbyte. The complications I ran into were:
- The Airbyte docker images include an nginx component which made things fairly complex routing traffic and in production would not be needed. I did some digging to see about removing nginx but based on the limited research I did, it appeared that we might need to build our own docker image of airbyte for that to work.
- Our pulumi stack is tightly coupled with consul and vault and spinning up a dev stack for testing an airbyte deployment wasn't very straightforward. Additionally, given that we were testing with keycloak as well and haven't fully verified/secured the config, I didn't want to be using our Vault CI cluster as it contains usable secrets. The same applies to keycloak. The devs were experimenting with the keycloak stack and didn't want anything coupled with any of our vault clusters, so had to setup a keycloak EC2 instance manually with DNS and a valid cert.
- The docs (traefik) aren't very clear on using the forward-auth middleware. Initially I didn't fully understand that the middleware is yet another docker image that's listening on another port and has traffic forwarded to it. Additionally there are a few different forward-auth containers, but most of the online references mention this https://github.com/thomseddon/traefik-forward-auth however the last commit to that repo was two years ago which is a bit concerning.

After running into those few issues, decided to go back to the basics and test out a simple app to verify the flow and make sure we have the basics configured (keycloak and traefik forward-auth)

#### Simple application
[...]

PR #1563, opened 2023-06-06
Added a basic keyloak/dagster client

shaidar:

This is somewhat related to https://github.com/mitodl/ol-infrastructure/issues/1473

Does two things:
- Renames keycloak instances to service name as opposed to product name
- Sets up a very basic dagster OIDC client which will be used by the traefik-forward-auth container config to communicate with keycloak
  - Initially, I was adding a bunch of configurable parameters to the client config, but then changed it back to keep it simple for now so that we can test things out and make sure the flow works before tweaking things further.

- [ ] Desktop screenshots

# How can this be tested?

I couldn't test the substructure bit as that was looking for the renamed keycloak domain name, however, I tried previewing the minor renaming change and that seems to be working.
# Additional Context

### DIFF
--- a/src/ol_infrastructure/substructure/keycloak/__main__.py
+++ b/src/ol_infrastructure/substructure/keycloak/__main__.py
@@ -116,3 +116,15 @@ required_action_update_password = keycloak.RequiredAction(
     enabled=True,
     opts=resource_options,
 )
+
+# Create Dagster OIDC client
+keycloak_domain_name = keycloak_config.require("url")
+dagster_openid_client = keycloak.openid.Client(
+    "ol-dagster-client",
+    realm_id=ol_platform_engineering_realm.realm,
+    client_id="ol-dagster-client",
+    enabled=True,
+    access_type="CONFIDENTIAL",
+    valid_redirect_uris=[f"https://{keycloak_domain_name}/*"],
+    login_theme="keycloak",
+)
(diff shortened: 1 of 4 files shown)


src/ol_infrastructure/substructure/keycloak/__main__.py
```diff
@@ -116,3 +116,15 @@
     enabled=True,
     opts=resource_options,
 )
+
+# Create Dagster OIDC client
+keycloak_domain_name = keycloak_config.require("url")
```
blarghmatey, line 121: Should we make this a `StackReference`?

src/ol_infrastructure/substructure/keycloak/__main__.py
```diff
@@ -116,3 +116,15 @@
...
+dagster_openid_client = keycloak.openid.Client(
+    "ol-dagster-client",
+    realm_id=ol_platform_engineering_realm.realm,
+    client_id="ol-dagster-client",
+    enabled=True,
+    access_type="CONFIDENTIAL",
+    valid_redirect_uris=[f"https://{keycloak_domain_name}/*"],
```
blarghmatey, line 128: If the domain gets moved to a stack reference this would become `keycloak_domain.format('https://{}/*')`

src/ol_infrastructure/substructure/keycloak/__main__.py
shaidar: I went with this cause I noticed we had a comment on line 12 about issues with pulumi reading stack defiinition

blarghmatey: I think those are disparate issues, but this works for today. We can revisit later.

PR #1566, opened 2023-06-08
Initial pass at adding production deployment for keycloak

shaidar:

This is related to the work done here https://github.com/mitodl/ol-infrastructure/issues/1473

Adds a few necessary stacks to deploy a production instance of keycloak.

- [ ] Desktop screenshots

# How can this be tested?

# Additional Context


blarghmatey (approved):

src/bridge/secrets/keycloak/data.production.yaml
blarghmatey, line 1: Shouldn't this be `operations.production`?

blarghmatey: Nevermind, I just saw that all of the other files have the same prefix. Does that mean that we're running Keycloak in the data VPC? If not then I'd suggest we either do away with the prefix and just make it `ci`, `qa`, and `production`, or replace the `data` prefix with `operations`

blarghmatey (approved):

PR #1575, opened 2023-06-13
Treat lock files as binary for git merge purposes.

feoh:
N/A

This .gitattributes file tells git to treat all the various lock files we know about as binary, because hand merging things like yarn.lock is an error prone operation.

# How can this be tested?

Ran git check-attributes binary poetry.lock

```
# cpatti @ rocinante in ~/src/mit/ol-infrastructure on git:main x [10:10:07] 
$ git check-attr binary poetry.lock
poetry.lock: binary: set
```

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,11 @@
+poetry.lock binary
+yarn.lock binary
+black.lockfile binary
+build-support-bin.lockfile binary
+flake8.lockfile binary
+ipython.lockfile binary
+isort.lockfile binary
+mypy.lockfile binary
+pytest.lockfile binary
+python-default.lockfile binary
+testing.lockfile binary


feoh:
@blarghmatey As near as I can tell, .gitattributes are, as you suggested, a Git level mechanism, and so putting this in .github wouldn't have any affect, but I'll put it there too because it can't hurt :)

.gitattributes
```diff
@@ -0,0 +1,12 @@
+poetry.lock binary
+yarn.lock binary
```
ChristopherChudzicki, line 2: In my experience resolving issues with yarn, the lockfile is sometimes useful to look at (either in whole or in diffs). My reccommendation/preference for `yarn.lock` (and probably the others) would be to use
```gitattributes
yarn.lock merge=binary linguist-generated
```
By using `merge=binary`, git shows a normal textual diff, but `git merge` shows something like
```
warning: Cannot merge binary files: a.txt (HEAD vs. c)
Auto-merging a.txt
CONFLICT (content): Merge conflict in a.txt
Automatic merge failed; fix conflicts and then commit the result.
```
And `linguist-generated` is what github recommends (https://docs.github.com/en/repositories/working-with-files/managing-files/customizing-how-changed-files-appear-on-github) for auto-generated files; it causes github to collapse the diff by default.

feoh:
@ChristopherChudzicki Good catch! See latest commit. Tested with:
```
# cpatti @ rocinante in ~/src/mit/ol-infrastructure on git:cpatti_add_gitattrs x [12:12:52] 
$ git check-attr linguist-generated poetry.lock
poetry.lock: linguist-generated: set
```

PR #1602, opened 2023-06-26
refactor: Add a means of overriding branding components in MFEs

blarghmatey:
N/A

This adds a common mechanism for specifying which MFE components to override with which plugins.

# How can this be tested?
This has been manually applied to the mitx-staging MFE pipeline


arslanashraf7 (commented):
The changes overall look clean.

But you mentioned that this is manually deployed on mitx-staging. But I don't think the learner dashboard brand override is working on https://staging.mitx.mit.edu/dashboard. Maybe because the learner dashboard isn't enabled there.

I have seen the learner dashboard enabled on the CI instance though, https://lms-ci.mitx.mit.edu/dashboard/.

shaidar (approved):

PR #1650, opened 2023-07-25
Enabling scheduled emails in MITXOnline Production.

Ardiea:
https://github.com/mitodl/hq/issues/1503

Enabling scheduled emails in MITXOnline Production


pdpinch (approved):
LGTM

PR #1687, opened 2023-08-16
Keycloak config for MIT Open

shaidar:

https://github.com/mitodl/mit-open/issues/21

This is an initial pass to setup the following:
- New realm
- Enables Google reCAPTHCA v2 as part of the registration flow
- Sets up a password policy
- Basic setup to configure Touchstone as an IDP. Still missing some configs for it and need to do some more digging and reach out to IST for a bit more information/help on configuration.

# How can this be tested?

Run it on QA and test out the added features listed above.

### DIFF
--- a/src/ol_infrastructure/substructure/keycloak/__main__.py
+++ b/src/ol_infrastructure/substructure/keycloak/__main__.py
@@ -158,7 +158,7 @@ ol_apps_realm = keycloak.Realm(
         "business_unit": f"operations-{env_name}",
     },
     display_name="OL Apps",
-    display_name_html="<b>MIT Open Learning Apps</b>",
+    display_name_html="<b>MIT Open Learning Applications</b>",
     enabled=True,
     login_theme="base",
     duplicate_emails_allowed=False,
--- a/src/ol_infrastructure/substructure/keycloak/__main__.py
+++ b/src/ol_infrastructure/substructure/keycloak/__main__.py
@@ -173,7 +173,7 @@ ol_apps_realm = keycloak.Realm(
     realm="olapps",
     reset_password_allowed=True,
     verify_email=True,
-    password_policy="upperCase(2) and digits(2) and length(12) and specialChars(2) and forceExpiredPasswordChange(365) and notUsername and notEmail",  # noqa: E501,S106 # pragma: allowlist secret
+    password_policy="upperCase(1) and digits(1) and specialChars(1) and length(8) and notUsername and notEmail",  # noqa: E501,S106 # pragma: allowlist secret
     security_defenses=keycloak.RealmSecurityDefensesArgs(
         brute_force_detection=keycloak.RealmSecurityDefensesBruteForceDetectionArgs(
             failure_reset_time_seconds=43200,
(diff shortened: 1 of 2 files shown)


src/ol_infrastructure/substructure/keycloak/__main__.py
```diff
@@ -145,3 +148,140 @@
...
+
+# Create OL Public Realm
+ol_apps_realm = keycloak.Realm(
+    "olapps",
+    access_code_lifespan="30m",
+    access_code_lifespan_user_action="15m",
+    attributes={
+        "business_unit": f"operations-{env_name}",
+    },
+    display_name="OLApps",
```
blarghmatey, line 160:
```suggestion
    display_name="OL Apps",
```

src/ol_infrastructure/substructure/keycloak/__main__.py
```diff
@@ -145,3 +148,140 @@
...
+    display_name_html="<b>MIT Open Learning Apps</b>",
```
blarghmatey, line 161:
```suggestion
    display_name_html="<b>MIT Open Learning Applications</b>",
```

src/ol_infrastructure/substructure/keycloak/__main__.py
```diff
@@ -145,3 +148,140 @@
...
+        digits=6,
+        initial_counter=2,
+        look_ahead_window=1,
+        period=30,
+        type="totp",
+    ),
+    realm="olapps",
+    reset_password_allowed=True,
+    verify_email=True,
+    password_policy="upperCase(2) and digits(2) and length(12) and specialChars(2) and forceExpiredPasswordChange(365) and notUsername and notEmail",  # noqa: E501,S106 # pragma: allowlist secret
```
blarghmatey, line 176: I'm not sure that we want to force end-users to rotate their passwords anually. Maybe also change the requirements for uppercase/digits/special characters to 1? And maybe require at least 2 of the 3?

src/ol_infrastructure/substructure/keycloak/__main__.py
```diff
@@ -145,3 +148,140 @@
...
+        starttls=True,
+    ),
+    ssl_required="external",
+    offline_session_idle_timeout="168h",
+    sso_session_idle_timeout="2h",
+    sso_session_max_lifespan="24h",
+    opts=resource_options,
+)
+
+required_action_configure_otp = keycloak.RequiredAction(
```
blarghmatey, line 217: Is this setting MFA as a required step for users? I'm not sure that we want to force that on everyone. I think it's a good idea, but we might want to give folks an option of not configuring it.

blarghmatey: Does that allow for the use of TOTP but not require it?

shaidar: So new users won't be forced to enable TOTP, but they do have the option to go to their account settings and add an authenticator app.

blarghmatey: Sounds good, thanks :+1:

blarghmatey (approved):

PR #1693, opened 2023-08-18
revert: certificate html view setting in Studio

arslanashraf7:
Part of https://github.com/mitodl/hq/issues/1630#issuecomment-1679566242

- This removes a setting that we recently added to enable the `Certificates` setting in edX Studio. Now that we have deactivated the certificates we want to close this path so that the course authors don't have this option to create certificates in studio. The certificates should be accessed from MITxOnline only.

# How can this be tested?
Once deployed, We should not see `Certificates` setting in Studio. e.g. https://studio.mitxonline.mit.edu/certificates/course-v1:MITxT+14.73x+1T2023


src/bilder/images/edxapp/templates/edxapp/mitxonline/common_values.yml.tmpl
```diff
@@ -270,7 +270,6 @@ FEATURES:
     AUTH_USE_OPENID_PROVIDER: true
     AUTOMATIC_AUTH_FOR_TESTING: false
     CUSTOM_COURSES_EDX: false
-    CERTIFICATES_HTML_VIEW: true  # ADDED KEY
```
arslanashraf7, line 273: I'm not sure If removing the key has the desired effect. Or should I set it to `false` instead?

 Asking this because i recently had an issue somewhere when removing the key didn't remove the key from Heroku. I'm not sure if that's the case here.

arslanashraf7:
pre-commit is failing but I believe that's not related to this PR since that's also failing with the same error on the `main` branch.

src/bilder/images/edxapp/templates/edxapp/mitxonline/common_values.yml.tmpl
Ardiea: That is actually a very good point. @shaidar care to comment? I think you may know more about this functionality than I do. 

I agree, now that it exists, it might be better to set it to false. I don't know if the salt code is smart enough to remove something it no longer "knows about".

Ardiea (approved):

Issue #989, opened 2022-08-02
edxapp monitoring data lacks some key metadata 


Ardiea:
# User Story
- Viewing monitoring data for specific instances of edxapp lack the required metadata to perform any meaningful aggregations or trend analysis. 

# Description/Context
edxapp installs need to have their vector configurations modified to populate the 'application' label for all node types, not just CAS and xqueue.  

# Acceptance Criteria
- [ ] edxapp installs need to have their vector configurations modified to populate the 'application' label for all node types, not just CAS and xqueue.

blarghmatey:
@Ardiea is this still an open issue, or has some of your other work already addressed this?

PR #1695, opened 2023-08-18
Misc Stuff 

Ardiea:
Closes https://github.com/mitodl/ol-infrastructure/issues/989

Three things
- Set the storage device on all packer builds to /dev/xdva rather than /dev/sda1 ref: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/device_naming.html
- Sets APPLICATION and SERVICE env vars on all installs of vector which will then get added to host metrics shipped into prometheus. 
- Adjusts the size of concourse worker disks.
-

### DIFF
--- a/src/ol_infrastructure/applications/airbyte/__main__.py
+++ b/src/ol_infrastructure/applications/airbyte/__main__.py
@@ -542,6 +542,8 @@ lt_config = OLLaunchTemplateConfig(
                                 "content": textwrap.dedent(
                                     f"""\
                             ENVIRONMENT={consul_dc}
+                            APPLICATION=data-platform
+                            SERVICE=airbyte
                             VECTOR_CONFIG_DIR=/etc/vector/
                             AWS_REGION={aws_config.region}
                             GRAFANA_CLOUD_API_KEY={grafana_credentials['api_key']}
--- a/src/ol_infrastructure/applications/airbyte/__main__.py
+++ b/src/ol_infrastructure/applications/airbyte/__main__.py
@@ -542,8 +542,8 @@ lt_config = OLLaunchTemplateConfig(
                                 "content": textwrap.dedent(
                                     f"""\
                             ENVIRONMENT={consul_dc}
-                            APPLICATION=data-platform
-                            SERVICE=airbyte
+                            APPLICATION=air-byte
+                            SERVICE=data-platform
                             VECTOR_CONFIG_DIR=/etc/vector/
                             AWS_REGION={aws_config.region}
                             GRAFANA_CLOUD_API_KEY={grafana_credentials['api_key']}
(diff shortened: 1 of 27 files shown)


src/ol_infrastructure/applications/airbyte/__main__.py
```diff
@@ -536,6 +536,8 @@
                                 "content": textwrap.dedent(
                                     f"""\
                             ENVIRONMENT={consul_dc}
+                            APPLICATION=data-platform
+                            SERVICE=airbyte
```
blarghmatey, line 540: I would flip these. The application is airbyte and it is a component of the data-platform service.

src/ol_infrastructure/applications/codejail/__main__.py
```diff
@@ -165,6 +165,8 @@
                                 "content": textwrap.dedent(
                                     f"""\
                             ENVIRONMENT={consul_dc}
+                            APPLICATION=edx
+                            SERVICE=codejail
```
blarghmatey, line 169: I would flip these as well. The application is the specific component and they are composed into an overarching service. For edX I would set the service name as `openedx`

src/ol_infrastructure/applications/keycloak/__main__.py
```diff
@@ -335,6 +335,8 @@
                                 "content": textwrap.dedent(
                                     f"""\
                             ENVIRONMENT={consul_dc}
+                            APPLICATION=keycloak
+                            SERVICE=keycloak
```
blarghmatey, line 339: I would argue service would be `sso` or `single-signon`

src/ol_infrastructure/applications/airbyte/__main__.py
```diff
@@ -536,8 +536,8 @@
                                 "content": textwrap.dedent(
                                     f"""\
                             ENVIRONMENT={consul_dc}
-                            APPLICATION=data-platform
-                            SERVICE=airbyte
+                            APPLICATION=air-byte
```
blarghmatey, line 539:
```suggestion
                            APPLICATION=airbyte
```

src/ol_infrastructure/applications/edxapp/__main__.py
```diff
@@ -1104,8 +1104,8 @@ def cloud_init_user_data_func(
                 "content": textwrap.dedent(
                     f"""\
                     ENVIRONMENT={consul_env_name}
-                    APPLICATION=edx
-                    SERVICE=edxapp
+                    APPLICATION=edxpp
```
blarghmatey, line 1107:
```suggestion
                    APPLICATION=edxapp
```

blarghmatey (approved):

PR #1704, opened 2023-08-21
Replace Caddy with Traffik for Concourse HTTPS

### LINKED ISSUE #1425: Replace caddy with traefik for tika and concourse
# User Story
- As a platform owner I want to reduce the number of technologies in use

# Description/Context
We are standardizing on Traefik for all of our proxying. Caddy is still in use for Concourse and Tika images so we would like to remove it from our infrastructure.

# Acceptance Criteria
- [ ] Concourse is using Traefik as its proxy
- [ ] Tika is using Traefik as its proxy
- [ ] Code for managing Caddy is removed from ol-infrastructure

feoh:
Fixes #1425

This change replaces Caddy's Let's Encrypt certificate generation and proxying with Traefik due to reliability concerns with Caddy and a desire for consistency across all our products and infrastructure.

# How can this be tested?

**TODO**
I'm concerned about this myself. I've build the container locally with pyinfra and that succeeds, and I now wonder what next steps are for testing this without actually deploying it to production.

Manual testing would include:
- In the resultant, running container, verifying that the Let's Encrypt was generated successfully, and that the host is listening on port 443.
- Validate that the correct local port is being proxied successfully


blarghmatey (commented):
We should probably also delete the caddyfile template

src/bilder/images/concourse/files/traefik/dynamic_config.yaml
```diff
@@ -0,0 +1,17 @@
...
+http:
+  routers:
+    concourse-router:
+      rule: Host("{{env "DOMAIN"}}")
+      service: concourse-service
+      tls:
+        certResolver: letsencrypt_resolver
+    concourse-healthcheck:
+      rule: Path("/api/v1/info")
+      service: concourse-service
```
Ardiea, line 12: You need to add the 

```
      tls:
        certResolver: letsencrypt_resolver
```

clause to the healthcheck router as well.

Ardiea (approved):

Issue #1707, opened 2023-08-22
Convert remaining edxapp envs to docker


Ardiea:
# Description/Context

Before anything:

- [x] reconcile configurations between edxapp and edxapp_v2 image build code

The following environments needs to be switched over to use the docker framework
- [x] residential production
- [x] residential staging production
- [ ] xpro production
- [x] mitxonline production (config already in place, will take effect next deployment)

Additionally
- [x] remove old image building code 
- [x] remove / archive old deployment pipelines

PR #1708, opened 2023-08-22
Config reconcile for edxapp dockerization 

Ardiea:
Partially addresses #1707 

Updated configuratiosn for dockerzied edxapp to match changes that took place in not-dockerized edxapp. Added a script that is helpful for reconconciling those configs.


Ardiea:
@blarghmatey 
In the mitxonline common config I found one item present in the _new_ configs that is not present in the old configs (which is kind of weird, IMO) Do you know anything about this and should it be removed from the new configs as well? 

```diff
287d299
<     ENABLE_COURSE_HOME_REDIRECT: true
```

All other differences are either reconciled in this PR or are expected (things like paths which change a bit inside containers, all the STATIC_ROOT_BASE stuff recently added for PR, etc).

blarghmatey:
It looks like that is because there was somehow a duplicate entry placed in the features dict there. I think we can just remove the second entry and leave the other. Not sure how it creeped in, but I'm not worried about it.

blarghmatey (approved):

PR #1719, opened 2023-08-28
Keycloak CI OpenID client for local testing

shaidar:

Added the following:
- Changed stack info and removed some redundant code when creating openid clients for different realms
- Enabled registration form
- Removed password change as part of initial flow
- Added saml config to QA and prod (still not fully functional)

# How can this be tested?

Ran it using the QA stack and ran without any errors.

### DIFF
--- a/src/ol_infrastructure/substructure/keycloak/__main__.py
+++ b/src/ol_infrastructure/substructure/keycloak/__main__.py
@@ -1,4 +1,3 @@
-import json
 import pulumi_keycloak as keycloak
 
 from pulumi import Config, ResourceOptions
@@ -8,7 +7,6 @@ env_config = Config("environment")
 stack_info = parse_stack()
 env_name = f"{stack_info.env_prefix}-{stack_info.env_suffix}"
 keycloak_config = Config("keycloak")
-keycloak_clients = keycloak_config.get("clients")
 
 
 # Create a Keycloak provider cause we ran into an issue with pulumi reading
@@ -122,22 +120,6 @@ required_action_update_password = keycloak.RequiredAction(
     opts=resource_options,
 )
 
-# Check if any OIDC clients exist in config and create them
-if keycloak_clients:
-    keycloak_clients = json.loads(keycloak_clients)
-    for client_name, client_domain in keycloak_clients.items():
-        openid_client = keycloak.openid.Client(
-            f"ol-{client_name}-client",
-            realm_id=ol_platform_engineering_realm.realm,
-            client_id=f"ol-{client_name}-client",
-            enabled=True,
-            access_type="CONFIDENTIAL",
-            standard_flow_enabled=True,
-            valid_redirect_uris=[f"{client_domain}/*"],
-            login_theme="keycloak",
-            opts=resource_options,
-        )
-
 # Create OL Public Realm
 ol_apps_realm = keycloak.Realm(
     "olapps",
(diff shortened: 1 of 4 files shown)


src/ol_infrastructure/substructure/keycloak/__main__.py
```diff
@@ -224,24 +207,44 @@
...
-    opts=resource_options,
-)
+
+# Check if any Openid clients exist in config and create them
+for openid_clients in keycloak_config.get_object("openid_clients"):
+    realm_name = openid_clients.get("realm_name")
+    client_details = openid_clients.get("client_info")
+    for client_name, client_url in client_details.items():
+        openid_client = keycloak.openid.Client(
+            f"ol-{client_name}-client",
```
blarghmatey, line 222: Maybe include the realm name in the `f` string to avoid collisions if we want to have the same app in multiple realms.

src/ol_infrastructure/substructure/keycloak/__main__.py
```diff
@@ -224,24 +207,44 @@
...
+            realm_id=realm_name,
+            client_id=f"ol-{client_name}-client",
```
blarghmatey, line 224: Maybe include the realm in the client_id as noted above.

src/ol_infrastructure/substructure/keycloak/__main__.py
```diff
@@ -224,24 +207,44 @@
...
+            enabled=True,
+            access_type="CONFIDENTIAL",
```
blarghmatey, line 226: Does this need to be configurable at all?

src/ol_infrastructure/substructure/keycloak/__main__.py
```diff
@@ -224,24 +207,44 @@
...
+            standard_flow_enabled=True,
+            valid_redirect_uris=[f"{client_url}/*"],
+            login_theme="keycloak",
```
blarghmatey, line 229: Should this be configurable as well?

src/ol_infrastructure/substructure/keycloak/__main__.py
```diff
@@ -224,24 +207,44 @@
...
+            opts=resource_options,
+        )
+
+if stack_info.env_suffix != "ci":
+    ol_apps_touchstone_saml_identity_provider = keycloak.saml.IdentityProvider(
```
blarghmatey, line 234: Can/should we also enable touchstone on the platform realm?

src/ol_infrastructure/substructure/keycloak/__main__.py
shaidar: Will do.

src/ol_infrastructure/substructure/keycloak/__main__.py
shaidar: Not sure if it's useful to add realm_name here as the clients in keycloak will be listed under the individual realms, so it'll be clear what realm they belong to. Not sure it gives us anything to add it here.

blarghmatey: Makes sense.

src/ol_infrastructure/substructure/keycloak/__main__.py
shaidar: Maybe down the line, but won't worry about it for now.

src/ol_infrastructure/substructure/keycloak/__main__.py
shaidar: Maybe down the line, but won't worry about it for now.

src/ol_infrastructure/substructure/keycloak/__main__.py
shaidar: Probably not as we either want to create the developer accounts ourselves or integrate with Github and write some code to adjust the action in the login flow to check for org membership

blarghmatey (approved):

Issue #1683, opened 2023-08-14
Global Secrets Management


Ardiea:
# Description/Context
`secret-operations` is a .... disaster. We need a managed space for our secrets that are shared between applications / stacks / whatever. 

# Plan/Design
- Create a substructure stack that makes a kv2 vault mount and populates it with various keys / data structures that will be shared between applications.
- Pick a few small apps / stacks and migrate them
  - That means tracking down the existing values in `secret-operations` and ensuring they get populated into the new mount via pulumi code. 
  - Update the application stacks to retrieve secrets from the new mount and not the old mount.
  - Make sure to update the .hcl files for the stacks so they have permission to access the new mount.
 
# Questions
- What, if anything should be exported from the substructure stack? Maybe paths to the root keys / groups
- Root keys / groups. Come up with a sane orgnaizsation system for the root keys /objects /paths in the new vault mount and make sure it is sane with the rest of the team.

shaidar:
@blarghmatey Now that the mount and a sample app (bootcamps) is complete, should we go ahead and close this issue and open separate issue to update pillar for the individual apps?

blarghmatey:
Yes, I think it makes sense to have an issue per app so that we can pick it off opportunistically.

PR #1742, opened 2023-09-11
Setting up vault global secrets

shaidar:

https://github.com/mitodl/ol-infrastructure/issues/1683

Creates a new key-value vault secrets mount and populates some of the global configs that apps need access to


src/ol_infrastructure/substructure/vault/secrets/__main__.py
```diff
@@ -0,0 +1,28 @@
...
+    options={"version": 2},
+    description="Storage of global secrets used across our applications.",
+    opts=ResourceOptions(delete_before_replace=True),
+)
+
+vault.generic.Secret(
+    "global-vault-secrets",
+    path=global_vault_mount.path.apply("{}/data".format),
+    data_json=json.dumps(global_vault_secrets),
+)
```
blarghmatey, line 28: This will just write all of the contents of the YAML file to a single path in the mount. If you want it to be one path per top-level key it would need to look like:
```suggestion
for key, data in global_vault_secrets.items():
    vault.generic.Secret(
        "global-vault-secrets",
        path=global_vault_mount.path.apply(lambda path: "{path}/{key}".format(path=path, key=key)),
        data_json=json.dumps(data),
    )
```

shaidar: Yeah I thought about, but don't see the point of splitting it into multiple paths at this point. Majority are global with the exception of Heroku (hoping that we'll be moving away from that in a year or so).

blarghmatey: I think that the Heroku API keys is a good reason to split it out. We don't want to make those randomly available to apps, even if we are the ones ultimately controlling the reads from Vault.

src/ol_infrastructure/substructure/vault/secrets/__main__.py
```diff
@@ -0,0 +1,31 @@
...
+    path="secret-global",
+    type="kv-v2",
+    options={"version": 2},
+    description="Storage of global secrets used across our applications.",
+    opts=ResourceOptions(delete_before_replace=True),
+)
+
+for key, data in global_vault_secrets.items():
+    vault.generic.Secret(
+        "global-vault-secrets",
```
blarghmatey, line 26: This will need to be changed because it will throw a duplicate ID error from being in the loop

blarghmatey (approved):

PR #1762, opened 2023-09-20
Add BULK_EMAIL_RETRY_DELAY_BETWEEN_SENDS

shaidar:
This should resolve SES timeouts that we're seeing since it currently configured at 19 email per second based on this https://github.com/openedx/edx-platform/blob/32a1820360e8c0908ee5c68031b44bdc599ab80f/lms/envs/common.py#L2941


blarghmatey:
Looks good :+1:

PR #1786, opened 2023-09-28
SAML Pull page fix

Ardiea:
Created a utility script for running the saml pull management command in edxapp every 24-48h. Should stop the paging every two-ish weeks for residential and residential staging. Doesn't do anything in environments w/o saml providers configured.


shaidar:
Should we spend some time trying to verify if the celery task is actually running but failing or not running at all - https://github.com/openedx/edx-platform/blob/master/common/djangoapps/third_party_auth/tasks.py#L28. If we're missing something simple or there's an easy fix using the existing task, I would imagine that would be the better route to take.

blarghmatey:
This might be obviated by https://github.com/mitodl/ol-infrastructure/commit/8c617823c768c10b49a62cff67ffa40479ca539b. The reference for this is found at https://github.com/mitodl/edx-platform/blob/master/lms/envs/production.py#L685-L689

PR #1806, opened 2023-10-06
Added Keycloak CAS plugin

shaidar:

This should install the following plugin - https://github.com/jacekkow/keycloak-protocol-cas and use the same version as the running instance of Keycloak.


blarghmatey:
It might be worth updating the other plugins to use the github release resource as well. Looking at them, they already have the JAR file pre-built, so we can probably just pull that into the container during context during build time and skip the JAR building step.

blarghmatey (approved):

PR #1810, opened 2023-10-06
Added resource def to get release tag

shaidar:

Adds a `tag_filter` resource to the concourse github-release type so that we can get a specific tagged release for some of the Keycloak SPI's.


src/ol_concourse/pipelines/infrastructure/keycloak/docker_packer_pulumi_pipeline.py
```diff
@@ -86,7 +86,10 @@ def build_keycloak_pipeline() -> Pipeline:
     # Repo: https://github.com/daniel-frak/keycloak-user-migration
     # Use: Migrating users from open discussions
     user_migration_spi = github_release(
-        Identifier("user-migration-spi"), "daniel-frak", "keycloak-user-migration"
+        Identifier("user-migration-spi"),
+        "daniel-frak",
+        "keycloak-user-migration",
+        KEYCLOAK_VERSION,
```
blarghmatey, line 92: We should set the version to be `22.0` so that patch releases will get pulled in.

shaidar: If we set it to that, wouldn't this break as it would be unable to find the keycloak version to deploy? https://github.com/mitodl/ol-infrastructure/blob/main/src/bilder/images/keycloak/deploy.py#L75

blarghmatey: No, because that is getting set based on the value that we are setting on line 185.

src/ol_concourse/pipelines/infrastructure/keycloak/docker_packer_pulumi_pipeline.py
```diff
@@ -183,9 +182,7 @@ def build_keycloak_pipeline() -> Pipeline:
         ],
         image_code=keycloak_packer_code,
         packer_template_path="src/bilder/images/keycloak/keycloak.pkr.hcl",
-        env_vars_from_files={
-            "KEYCLOAK_VERSION": f"{keycloak_customization_repo.name}/.git/describe_ref"
-        },
+        env_vars_from_files={"KEYCLOAK_VERSION": KEYCLOAK_VERSION},
```
blarghmatey, line 185: We should probably keep the previous behavior because that is the value that we are using to set the tag on the image that we upload to dockerhub.

blarghmatey: https://github.com/mitodl/ol-infrastructure/blob/b6cf504e5badacb09483a45ea8de5dae991222d4/src/ol_concourse/pipelines/infrastructure/keycloak/docker_packer_pulumi_pipeline.py#L155-L157

blarghmatey (approved):

PR #1840, opened 2023-10-13
fix: fix default xmodule class name

asadali145:
Might fix https://github.com/mitodl/hq/issues/2687 & https://github.com/mitodl/hq/issues/2460. Also, this [sentry issue](https://mit-office-of-digital-learning.sentry.io/issues/3825401912/events/57d9dc51546d46eca0cba6a2d451e84d/?project=5939801).

edX has renamed one of the module store classes. Also, they have updated the settings in edx-platform. We are using the old name as defined in our configurations. This PR updates the MODULESTORE setting for MITx Online. Here is the edX PR https://github.com/openedx/edx-platform/pull/31113

# How can this be tested?

We can test the grading and check sentry.

### DIFF
--- a/src/bilder/images/edxapp/templates/edxapp/mitxonline/common_values.yml.tmpl
+++ b/src/bilder/images/edxapp/templates/edxapp/mitxonline/common_values.yml.tmpl
@@ -392,7 +392,7 @@ MODULESTORE:
               DOC_STORE_CONFIG:
                 <<: *docstore_config
               OPTIONS:
-                default_class: xmodule.hidden_module.HiddenDescriptor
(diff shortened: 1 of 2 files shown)


blarghmatey:
Should these changes also be applied in xPro and residential MITx?

asadali145:
> Should these changes also be applied in xPro and residential MITx?

Yes, should be applied to residential MITx at least as we are on Palm release. No need for xPro as we are using Olive release. We will create a PR once we test things for MITx Online.

asadali145:
These changes are required from Palm onwards.

Issue #1628, opened 2023-07-12
Edxapp deployment pipeline enhancments


Ardiea:
# Description/Context
Implement a wait-on-instance-refresh step for edxapp deployments
Implement a cache invalidation on edxapp static resources in fastly.
Other enhancements go here  

# Plan/Design

PR #1842, opened 2023-10-13
Issue 1629 - Log shipping for edxapp fastly distributions 

Ardiea:
Partially addresses #1628 

This enables log shipping from fastly to Grafana and S3 following the same pattern we use for OCW.

Additionally it makes a few things more consistent with how the log shipping is setup with both edxapp and ocw. 

It also addresses a bug introduced by the `get_monitoring_sns_arn` helper function which silently creates stack references in stacks that did not explicitly ask for it. The helper function will continue to create that stack reference, but it is now given a special `implicit` name that will prevent collisions when stacks later explicitly ask for references to the `infrastructure.monitoring` stack. Also documented this behavior.


blarghmatey (approved):
Nicely done

PR #1853, opened 2023-10-17
Downgrade Keycloak to resolve IDP issue

shaidar:

Touchstone auth stopped working using Keycloak. Seeing a `400 Bad Request` error when a `POST` request takes place on the keycloak saml endpoint despite the IDP sending a success messages in its SAMLResponse. The error in the [logs](https://mitolqa.grafana.net/explore?panes=%7B%22EoO%22:%7B%22datasource%22:%22grafanacloud-logs%22,%22queries%22:%5B%7B%22refId%22:%22A%22,%22expr%22:%22%7Bapplication%3D%5C%22keycloak%5C%22%7D%20%7C%3D%20%60Not%20possible%20to%20decrypt%20SAML%20assertion%60%22,%22queryType%22:%22range%22,%22datasource%22:%7B%22type%22:%22loki%22,%22uid%22:%22grafanacloud-logs%22%7D,%22editorMode%22:%22builder%22%7D%5D,%22range%22:%7B%22from%22:%22now-6h%22,%22to%22:%22now%22%7D%7D%7D&schemaVersion=1&orgId=1) is `Not possible to decrypt SAML assertion`. Checked with IST and the keycloak side and verified that nothing changed as far as the the public key that the IDP is using to encrypt assertions. Came across the following closed [issue](https://github.com/keycloak/keycloak/issues/19689) on the keycloak repo which might be what's causing our issue. So testing it out to see if a downgrade resolves the problem.


collinpreston:
Did we recently update the Keycloak version we are on?

src/bridge/lib/versions.py
```diff
@@ -3,7 +3,7 @@
 CONSUL_TEMPLATE_VERSION = "0.29.5"
 CONSUL_VERSION = "1.14.0"
 DAGSTER_VERSION = "1.0.17"
-KEYCLOAK_VERSION = "22.0"
```
blarghmatey, line 6: This will break the plugins that we are installing because Keycloak 20 is using an older version of Java than what the plugin Jars were compiled with.

shaidar:
Closing PR in favor of this change - https://github.com/mitodl/ol-keycloak-customization/commit/2b34beacca8c4a8b3ebea4a6ebb9409ad596ba88

Issue #1855, opened 2023-10-18
Define user migration plug-in configuration for Open Discussions in QA


collinpreston:
# Description/Context
A user federation configuration needs to be defined for the User migration plug-in in Keycloak to work with Open Discussions in QA.

# Plan/Design
https://docs.google.com/document/d/17tJ-C2EwWoSpJWZKjuhMVgsqGtyPH0IN9KakXvSKU0M/edit#heading=h.geba7zl9ukna

REST API details: https://github.com/mitodl/open-discussions/pull/4231

PR #1859, opened 2023-10-20
1855: define user migration plug in configuration for open discussions in qa

collinpreston:
https://github.com/mitodl/ol-infrastructure/issues/1855

This defines the Keycloak configuration to support and configure the Keycloak user migration plug-in described here:
https://docs.google.com/document/d/17tJ-C2EwWoSpJWZKjuhMVgsqGtyPH0IN9KakXvSKU0M/edit#heading=h.geba7zl9ukna

The corresponding functionality for this plug-in was added to Open Discussions here https://github.com/mitodl/open-discussions/issues/4227

There are bearer token values that will need to be created in Vault and referenced in the substructure yaml files.

# How can this be tested?
Once this configuration is deployed to QA, Open Discussion users should be able to authenticate with Keycloak using their Open Discussion credentials.

### DIFF
--- a/src/ol_infrastructure/substructure/keycloak/__main__.py
+++ b/src/ol_infrastructure/substructure/keycloak/__main__.py
@@ -457,3 +457,23 @@ if stack_info.env_suffix != "production":
         ),
     )
     # Okta TEST SAML [END]
+
+keycloak.CustomUserFederation(
+    "ol-open-discussions-qa-user-migration-federation",
+    cache_policy="DEFAULT",
+    config={
+        "API_TOKEN_ENABLED": True,
+        "API_HTTP_BASIC_ENABLED": False,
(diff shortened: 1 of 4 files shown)


src/ol_infrastructure/substructure/keycloak/__main__.py
```diff
@@ -457,3 +457,26 @@
         ),
     )
     # Okta TEST SAML [END]
+
+if stack_info.env_suffix in ["qa", "rc"]:
```
shaidar, line 461: U mean qa and ci?

collinpreston: No, I just want this to be in QA/RC.  I thought I saw somewhere that the value of `env_suffix` could be "qa" or "rc".

shaidar: In that case, then it just needs to be qa

shaidar (approved):

Issue #1856, opened 2023-10-18
Keycloak SAML config using an Okta application that replicates Touchstone in QA and CI


collinpreston:
# Description/Context
In order to test changes and to the Touchstone SAML configuration in Keycloak, without impacting developers in RC and CI, we should create mock instances of Touchstone using a developer Okta application. 

# Plan/Design
mock instance of Touchstone in Okta should be as close as possible to the Toucshtone configuration in QA and CI.  The configuration should be added to `src/ol_infrastructure/substructure/keycloak/__main__.py`

PR #1860, opened 2023-10-20
1856: Add fake touchstone and okta in CI and QA

collinpreston:
https://github.com/mitodl/ol-infrastructure/issues/1856

This PR contributes the following changes to the Keycloak configuration:

1. Adds two new Okta SAML identity providers which are replicated to be copies of Touchstone.  2 providers have been created, 1 for CI and 1 for QA.
2. Adds an Okta SAML identity provider in CI.  This same one already exists in QA.  The intention is that this Okta IDp can be modified whereas the Touchstone copies should never be modified.
3. Adds some naming conventions such as prepending "ol-" for OOTB functionality that has been duplicated and customized (first login flow for example).  This is explained here: https://docs.google.com/document/d/17tJ-C2EwWoSpJWZKjuhMVgsqGtyPH0IN9KakXvSKU0M/edit#heading=h.qb4vm0tanb33

# How can this be tested?
Once deployed:

1. The new Okta IDp's (2 in CI, 1 in QA) should allow for authentication with Keycloak.
2. Existing Touchstone login should function.


shaidar:
To avoid some duplication, can't we use the same code for fake-touchstone and Okta-test for both CI and QA?

src/ol_infrastructure/substructure/keycloak/__main__.py
```diff
@@ -334,85 +333,147 @@
...
         want_assertions_encrypted=True,
         want_assertions_signed=True,
         opts=resource_options,
         first_broker_login_flow_alias=ol_touchstone_first_login_flow.alias,
+        hide_on_login_page=False,
     )
     oidc_attribute_importer_identity_provider_mapper = (
         keycloak.AttributeImporterIdentityProviderMapper(
-            "map-touchstone-saml-email-attribute",
+            "map-fake-touchstone-ci-saml-email-attribute",
```
shaidar: This probably needs to be otherwise it'll throw an error for using the same name multiple times.
```suggestion
            f"map-fake-touchstone-{stack_info.env_suffix}-saml-email-attribute",
```

PR #1876, opened 2023-10-26
Revert "Revert "Revert "feat: set COURSES_INVITE_ONLY to true"""

asadali145:
Reverts mitodl/ol-infrastructure#1789


asadali145:
We need to revert this change. Our changes for forced enrollments in Open edX are available from Palm onwards, but we are on Olive so forced enrollments cannot work.

We should either backport forced enrollments in Olive or We should wait for Quince.

Here is the commit for forced enrollments https://github.com/openedx/edx-platform/commit/3e35e3af8d895c14f05a55a134dbf8a068cd584a.
CC: @blarghmatey @arslanashraf7

blarghmatey (approved):

arslanashraf7:
@asadali145 I'm merging this on your behalf to get the enrollment back to working in xPRO RC. I hope that's fine with you. 

@blarghmatey I'm assuming that merging this would auto-deploy this change on QA/RC.

PR #1888, opened 2023-10-31
Giving mitxonline edxapp workers dedicated CPU

Ardiea:
https://github.com/mitodl/hq/issues/2879

This switches edxapp workers in the mitxonline production env to c6a.large which has two dedicated vCPU rather than burstable instances that the stacks default to. This does reduce the amount of memory available but after confirming in grafana, the workers only ever utulize about 2.2GB of memory so we are still well within the resources. 

Costs about 0.003/h * 5 more.

### DIFF
--- a/src/ol_infrastructure/applications/edxapp/Pulumi.applications.edxapp.mitxonline.Production.yaml
+++ b/src/ol_infrastructure/applications/edxapp/Pulumi.applications.edxapp.mitxonline.Production.yaml
@@ -33,6 +33,7 @@ config:
   edxapp:use_docker: true
   edxapp:web_node_capacity: "9"
   edxapp:worker_node_capacity: "5"
+  edxapp:worker_instance_type: "c6a.large"
   mongodb:atlas_project_id: 61f0a63f8bc1f86a073a7148
   redis:instance_type: cache.r6g.xlarge
   vault:address: https://vault-production.odl.mit.edu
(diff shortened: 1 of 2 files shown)


src/ol_infrastructure/applications/edxapp/Pulumi.applications.edxapp.mitxonline.Production.yaml
```diff
@@ -33,6 +33,7 @@ config:
   edxapp:use_docker: true
   edxapp:web_node_capacity: "9"
   edxapp:worker_node_capacity: "5"
+  edxapp:worker_instance_type: "c6a.large"
```
blarghmatey, line 36: Should we set this as the default for all production deployments?

blarghmatey: They should all have similar usage. It's definitely worth running on non-burstable instances at the least.

Ardiea: So actually it is interesting, mitxonline is the only even moderately taxed production environment. I would say the default could actually be t3a.medium. They all sit nicely at ~2.5 GB memory and full CPU balances. That would save a couple hundred $ a month.

blarghmatey (approved):

PR #1902, opened 2023-11-07
Point to new keycloak customization repo

shaidar:

We moved theme assets and a mod to the login form to a new repo and this is a minor change to point to that new repo


blarghmatey (changes requested):
The Dockerfile name on line 149 will also need to be changed

blarghmatey (approved):

PR #1904, opened 2023-11-08
Added ol-spi

shaidar:

The new ol-keycloak repo has both a theme and some custom SPI for keycloak. The relevant theme files are copied and handled through the build process that we've followed previously using the old keycloak customization repo. The SPI however makes more sense to treat it as the other SPIs that we are using and this change does just that.
The one added thing that the developers would need to do is to create a `jar` if they're creating any SPIs and create a new release. This is something that I will document and add to the README in the ol-keycloak repo.


blarghmatey (approved):
I'm assuming that we're adding a GH workflow for compiling the jar?

Issue #1910, opened 2023-11-09
Keycloak - update browser login flow to use custom username-password form


collinpreston:
# Description/Context
In order to display the linked and unlinked IDPs for a user, as depicted in this process flow https://docs.google.com/document/d/17tJ-C2EwWoSpJWZKjuhMVgsqGtyPH0IN9KakXvSKU0M/edit#heading=h.tnkgkgxkelz4, a custom browser flow must be created in Keycloak which utilizes the custom form implemented for https://github.com/mitodl/ol-infrastructure/issues/1882. 

# Plan/Design
This form should be modeled as:
![image](https://github.com/mitodl/ol-infrastructure/assets/8311573/d4973b72-1ef5-47f5-90c5-cdfe4d22da87)

PR #1914, opened 2023-11-09
Define realm login theme, and add cookie to browser flow

collinpreston:
Related to https://github.com/mitodl/ol-infrastructure/issues/1910

Updates the olapps realm login theme to be the custom ol theme.  Adds the cookie check for the login flow.


src/ol_infrastructure/substructure/keycloak/__main__.py
```diff
@@ -46,7 +46,7 @@
     display_name="OL PLatform Engineering",
     display_name_html="<b>OL PLatform Engineering</b>",
     enabled=True,
-    login_theme="keycloak",
+    login_theme="ol",
```
shaidar, line 49: Assuming u wanted to change this on the `olapps` realm instead

shaidar (approved):

Issue #1881, opened 2023-10-30
earthly for edxapp docker images?


Ardiea:
as a devops engineer i want to have a modular `Earthfile` for building edxapp docker images so that I can build multiple final images from one build process. Primarily to allow us to parameterize the `PROCTORTRACK_CDN_URL` and `PROCTORTRACK_CONFIG_KEY` variables.

Need to determine how we handle passing the different artifacts to the right environments. And need to minimize HOW different the two artifacts are.

PR #1936, opened 2023-11-16
Md/issue 1881

Ardiea:
https://github.com/mitodl/ol-infrastructure/issues/1881

Does several thing:

1. Converts the edxapp docker file into an [Earthfile](https://docs.earthly.dev/). 
2. Uses said Earthfile to produce two produce two new build artifacts in addition to the largely un-altered original docker image.
  a. `staticfiles-nonprod-{docker image digest}.tar.gz`
  b. `staticfiles-production-{docker image digest}.tar.gz`
  c. **All static content is removed from the docker image now.** So there is no default static content shipped with the container.
3. Creates a new build+deploy pipeline, replacing the `docker build` steps with appropriate `earthly build` steps and adding tasks to ship the staticfiles content to s3. 
4. Makes modifications to the AMI build + docker compose configuration to use this new separated static content.
5. Makes modifications to the pulumi stack to populate a new environment var at runtime `ENVIRONMENT_TIER` to help docker-compose select the right staticfiles directory to mount into the container. 

# Remaining Tasks
1. ~Add this docker file to the repo and create a build and publish pipeline for it. This will replace `source={"repository": "ardiea/earthly", "tag": "v0.3"},` currently in use for the earthly build task in the pipeline. We may want to fork the original Dockerfile for this to ensure we're using a more up-to-date base.~
```
FROM --platform=linux/amd64 taylorsilva/dcind:20.10.8-20211022

RUN wget https://github.com/earthly/earthly/releases/download/v0.7.20/earthly-linux-amd64 -O /usr/local/bin/earthly && \
  chmod +x /usr/local/bin/earthly && \
  /usr/local/bin/earthly bootstrap

RUN apk --no-cache add git
```
2. ~Optimizations. I would say there is at least 10 minutes to save inside the build+deploy pipeline. There is a lot of copying stuff that doesn't always need to be copied that can be cleaned up. Additionally, the new docker image is somehow larger than the current one, despite actually containing fewer things. So we should look at that too.~ Shaved about 20 minutes off and several copies.  
3. ~Cleanup the old pipeline and dockerfile definitions. Cleanup in concoruse as well, once this is verified and working as expected.~
  a. Need to update the metapipeline at some point, as well.   
5. ~Look at moving the collecting of resources away from concourse and making it purely an 'earthly concern'. That is, don't spend so much time copying files from the various inputs -> the one place for earthly, just let earthly go get the stuff it needs. Would want a way to pass something like the commit hash from the concourse resource to earthly + have it default to a branch or whatever. This would make it easier to run the build otuside of concourse (which it very much is not easy at the moment earthly or dockerfile).~ 

# How can this be tested?
The build and deployment pipeline exists currently @ https://cicd.odl.mit.edu/teams/infrastructure/pipelines/earthly-packer-pulumi-edxapp-global but use caution unpausing it / the jobs within it as it will fight with the existing pipeline.

### DIFF
--- a/dockerfiles/openedx-edxapp/Earthfile
+++ b/dockerfiles/openedx-edxapp/Earthfile
@@ -1,0 +1,173 @@
+VERSION 0.7
+FROM python:3.8.12-buster
+
+apt-base:
+  ENV DEBIAN_FRONTEND=noninteractive
+  COPY ./apt-pkg-list /root/
+  COPY ./pip_package_lists /root/pip_package_lists
+  COPY ./pip_package_lists /root/pip_package_overrides
+  RUN apt update \
+    && apt install -y --no-install-recommends $(cat /root/apt-pkg-list) \
+    && apt autoremove -y \
+    && apt clean \
+    && rm -rf /var/lib/apt/lists/*
+  SAVE ARTIFACT /root/pip_package_lists
+
+locales:
+  FROM +apt-base
+  ARG OPENEDX_I18N_VERSION="open-release/olive.3"
+  WORKDIR /tmp
+  RUN curl -L -o openedx-i18n.tar.gz https://github.com/openedx/openedx-i18n/archive/$OPENEDX_I18N_VERSION.tar.gz \
+    && tar xzf /tmp/openedx-i18n.tar.gz \
+    && mkdir -p /openedx/locale/contrib \
+    && mv openedx-i18n-*/edx-platform/locale /openedx/locale/contrib
+  CACHE /openedx/locale
+  SAVE ARTIFACT /openedx/locale
+
+get-code:
+  FROM +apt-base
+  # Two required vars to specify the git repo and sha ref the build from.
+  ARG EDX_PLATFORM_GIT_REPO="invalid"
+  ARG EDX_PLATFORM_GIT_BRANCH="invalid"
+  GIT CLONE --branch $EDX_PLATFORM_GIT_BRANCH $EDX_PLATFORM_GIT_REPO /openedx/edx-platform
+  RUN cp /openedx/edx-platform/requirements/edx/base.txt /root/pip_package_lists/edx_base.txt
+  SAVE ARTIFACT /openedx/edx-platform
+
+install-deps:
+  FROM +get-code
+  ARG DEPLOYMENT_NAME="invalid"
+  ARG RELEASE_NAME="invalid"
+  RUN pip install --no-warn-script-location --user --no-cache-dir -r /root/pip_package_lists/edx_base.txt -r /root/pip_package_lists/$RELEASE_NAME/$DEPLOYMENT_NAME.txt
+  RUN pip install --no-warn-script-location --user --no-cache-dir -r /root/pip_package_overrides/$RELEASE_NAME/$DEPLOYMENT_NAME.txt
+  IF [ "$DEPLOYMENT_NAME" = "mitxonline" ]
+    RUN pip uninstall --yes "edx-name-affirmation"
+  END
+  WORKDIR /openedx/edx-platform
+  ENV PATH /root/.local/bin:/openedx/nodeenv/bin:${PATH}
+  ENV NPM_REGISTRY "https://registry.npmjs.org/"
+  RUN nodeenv /openedx/nodeenv --node=16.14.0 --prebuilt \
+    && npm clean-install -s --registry=$NPM_REGISTRY
+  SAVE ARTIFACT /openedx/edx-platform
+  SAVE ARTIFACT /openedx/nodeenv
+  SAVE ARTIFACT /root/.local
+
+themes:
+  FROM +apt-base
+  ARG THEME_GIT_REPO="invalid"
+  ARG THEME_GIT_BRANCH="invalid"
+  GIT CLONE --branch $THEME_GIT_BRANCH $THEME_GIT_REPO /openedx/themes/$RELEASE_NAME
+  SAVE ARTIFACT /openedx/themes
+
+tutor-utils:
+  FROM +apt-base
+  ARG TUTOR_REPO="https://github.com/overhangio/tutor.git"
+  ARG TUTOR_VERSION="v15.3.4"
+  GIT CLONE --branch $TUTOR_VERSION $TUTOR_REPO /openedx/tutor
+  SAVE ARTIFACT /openedx/tutor/tutor/templates/build/openedx/bin
+
+dockerize:
+  FROM docker.io/powerman/dockerize@sha256:983fd3e74c93b05f34533aec9ca0457f0748f173cd00acb119ec1983cbf6b62d
+  SAVE ARTIFACT /usr/local/bin/dockerize
+
+collected:
+  FROM +apt-base
+  ARG APP_USER_ID=1000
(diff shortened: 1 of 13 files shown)


dockerfiles/openedx-edxapp/Earthfile
```diff
@@ -0,0 +1,171 @@
...
+locales:
+  FROM +apt-base
+  ARG OPENEDX_I18N_VERSION="open-release/olive.3"
+  WORKDIR /tmp
+  RUN curl -L -o openedx-i18n.tar.gz https://github.com/openedx/openedx-i18n/archive/$OPENEDX_I18N_VERSION.tar.gz \
+    && tar xzf /tmp/openedx-i18n.tar.gz \
+    && mkdir -p /openedx/locale/contrib \
+    && mv openedx-i18n-*/edx-platform/locale /openedx/locale/contrib
+  CACHE /openedx/locale
+  SAVE ARTIFACT /openedx/locale
```
Ardiea, line 25: I don't think we need to do some of these early SAVE ARTIFACT steps.

src/bilder/images/edxapp_v2/files/docker-compose.yaml.tmpl
```diff
@@ -14,6 +14,8 @@ services:
     volumes:
     - /opt/data:/openedx/data
     - /opt/data/media:/openedx/media
+    - /opt/staticfiles-${ENVIRONMENT_TIER}:/openedx/staticfiles:ro
+    - /opt/staticfiles-${ENVIRONMENT_TIER}/bundles:/openedx/edx-platform/common/static/bundles:ro
```
Ardiea, line 18: Strictly speaking we don't need to mount the staticfiles into _all_ of these containers/services. Really they should only matter to two web containers. There may be performance gains by not mounting it when it isn't needed.

src/ol_concourse/pipelines/open_edx/edx_platform_v2/earthly_packer_pulumi_pipeline.py
```diff
@@ -0,0 +1,370 @@
...
+                                    f"""source /docker-lib.sh;
+                                    start_docker;
+                                    echo "{release_name}-{deployment_name}-$(cat ol-infrastructure-docker/dockerfiles/openedx-edxapp/edx_platform/.git/short_ref)" > artifacts/tag.txt;
+                                    cd {earthly_git_resource.name}/dockerfiles/openedx-edxapp;
+                                    RELEASE_NAME={release_name};
+                                    DEPLOYMENT_NAME={deployment_name};
+                                    earthly +docker-image --DEPLOYMENT_NAME=$DEPLOYMENT_NAME --RELEASE_NAME=$RELEASE_NAME;
+                                    DIGEST=$(docker inspect --format '{{{{.Id}}}}' mitodl/edxapp-$DEPLOYMENT_NAME-$RELEASE_NAME | cut -d ":" -f2);
+                                    echo "Saving docker image to tar file in the artifacts directory";
+                                    docker save -o ../../../artifacts/image.tar $DIGEST;
```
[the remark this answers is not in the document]
Ardiea: so create two different `artifacts` outputs.

src/ol_concourse/pipelines/open_edx/edx_platform_v2/earthly_packer_pulumi_pipeline.py
```diff
@@ -0,0 +1,370 @@
...
+                                path="bash",
+                                args=[
+                                    "-xc",
+                                    f"""source /docker-lib.sh;
+                                    start_docker;
+                                    echo "{release_name}-{deployment_name}-$(cat ol-infrastructure-docker/dockerfiles/openedx-edxapp/edx_platform/.git/short_ref)" > artifacts/tag.txt;
+                                    cd {earthly_git_resource.name}/dockerfiles/openedx-edxapp;
+                                    RELEASE_NAME={release_name};
+                                    DEPLOYMENT_NAME={deployment_name};
+                                    earthly +docker-image --DEPLOYMENT_NAME=$DEPLOYMENT_NAME --RELEASE_NAME=$RELEASE_NAME;
```
Ardiea, line 213: The digest that we retrieve here does NOT end up being the digest that is reported by dockerhub for whatever reason that I can't be asked to understand. So for purposes of building the AMI later in the pipeline, we get the digest from proper `registry_image` Put/Get. I suspect the Get part of that is happening one more times than really necessary but I haven't found the correct optimization for that yet.

src/ol_concourse/pipelines/container_images/dcind.py
```diff
@@ -0,0 +1,74 @@
...
+        "repository": "mitodl/dcind",
+        "tag": "latest",
+        "password": "((dockerhub.password))",
+        "username": "((dockerhub.username))",
+    },
+)
+
+build_task = container_build_task(
+    inputs=[Input(name=ol_inf_repo.name)],
+    build_parameters={"CONTEXT": f"{ol_inf_repo.name}/dockerfiles/dcind"},
```
blarghmatey, line 44: Shouldn't this also get the `BUILD_ARG` parameter for the Earthly release version?

dockerfiles/openedx-edxapp/Earthfile
```diff
@@ -0,0 +1,173 @@
...
+  RUN cp /openedx/edx-platform/requirements/edx/base.txt /root/pip_package_lists/edx_base.txt
+  SAVE ARTIFACT /openedx/edx-platform
+
+install-deps:
+  FROM +get-code
+  ARG DEPLOYMENT_NAME="invalid"
+  ARG RELEASE_NAME="invalid"
+  RUN pip install --no-warn-script-location --user --no-cache-dir -r /root/pip_package_lists/edx_base.txt -r /root/pip_package_lists/$RELEASE_NAME/$DEPLOYMENT_NAME.txt
+  RUN pip install --no-warn-script-location --user --no-cache-dir -r /root/pip_package_overrides/$RELEASE_NAME/$DEPLOYMENT_NAME.txt
+  IF [ "$DEPLOYMENT_NAME" = "mitxonline" ]
```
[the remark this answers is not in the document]
Ardiea: Yah it is cool to have it as a native construct rather than faking it in shell and ending up with an additional layer for essentially a noop.

src/ol_concourse/pipelines/container_images/dcind.py
Ardiea: This was surprisingly difficult to accomplish.

src/ol_concourse/pipelines/open_edx/edx_platform_v2/earthly_packer_pulumi_pipeline.py
```diff
@@ -0,0 +1,335 @@
...
+                    ),
+                    PutStep(
+                        put=edx_registry_image_resource.name,
+                        inputs=[Identifier("artifacts")],
+                        params={
+                            "image": "artifacts/image.tar",
+                            "additional_tags": "artifacts/tag.txt",
+                        },
+                    ),
+                    TaskStep(
```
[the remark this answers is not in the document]
Ardiea: ... Yes. I will find an example because I'm sure this is done elsewhere.

blarghmatey (approved):

PR #1946, opened 2023-11-29
Framework tagging enhancements for edxapp images

Ardiea:
Relates to https://github.com/mitodl/ol-infrastructure/issues/1881

Added dynamic framework tagging to ami generation for edxapp based a packer var.

# Additional Context
This will keep the earthly images from sneaking in pipelines where they are not welcomed.

### DIFF
--- a/src/ol_infrastructure/applications/edxapp/__main__.py
+++ b/src/ol_infrastructure/applications/edxapp/__main__.py
@@ -134,7 +134,7 @@ edxapp_vpc_id = edxapp_vpc["id"]
 data_vpc = network_stack.require_output("data_vpc")
 data_integrator_secgroup = data_vpc["security_groups"]["integrator"]
 
-framework = "docker" if edxapp_config.get_bool("use_docker") else "ansible"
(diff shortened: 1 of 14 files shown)


src/ol_infrastructure/applications/edxapp/__main__.py
```diff
@@ -134,7 +134,7 @@
 data_vpc = network_stack.require_output("data_vpc")
 data_integrator_secgroup = data_vpc["security_groups"]["integrator"]
 
-framework = "docker" if edxapp_config.get_bool("use_docker") else "ansible"
+framework = edxapp_config.get("framework") or "ansible"
```
blarghmatey, line 137: Given the current level of adoption it probably makes sense to default to docker rather than ansible.

blarghmatey (approved):

src/ol_infrastructure/applications/edxapp/__main__.py
Ardiea: Fair.

PR #1950, opened 2023-11-30
Keycloak: remove the custom username/password form from the browser flow

collinpreston:
Relates to https://github.com/mitodl/mit-open/issues/91

The custom username and password form, which provided templates with a variable indicating which SAML providers a user was not linked with, is no longer needed.  This PR updates the ol-browser flow to use the OOTB username/password form instead of the custom one.


src/ol_infrastructure/substructure/keycloak/__main__.py
```diff
@@ -496,7 +496,7 @@
     "ol-auth-username-password-form",
```
shaidar, line 496: Should we rename this as well?

collinpreston: Good catch, updated.

shaidar (approved):

PR #1951, opened 2023-12-01
Issue 1931

### LINKED ISSUE #1931: slack-url value not persisting in production vault
# Description/Context
Recently, we had some failing OCW pipelines in Concourse. When OCW pipelines fail, one of the things they do is use the Slack API to post a message in a channel so we can know the pipeline has failed. In the most recent failures, the Slack alert step also failed with `undefined vars: slack-url`. The var still exists in RC / QA, but had to be manually added back to production.

# Plan/Design
We should make sure that the `secret_concourse.ocw.slack-url` key persists and does not disappear from Vault.

Ardiea:
Closes #1931 

Manage the slack-url secret with pulumi for OCW concourse pipelines.

# How can this be tested?
Tested and verified in CI.


blarghmatey (approved):
Assuming that the pipeline code is using var syntax that looks like `((slack-url.value))` then this looks good.

blarghmatey:
According to the [Concourse docs](https://concourse-ci.org/vault-credential-manager.html#vault-credential-lookup-rules) it handles the automatic attribute access of `.value` if there is no attribute specified.

blarghmatey:
:shipit:

PR #1967, opened 2023-12-07
Add keycloak ratelimit though traefik

shaidar:

Our Open login form allows one to type a username and prior to typing a password, routes a user to the identity provider. This presents a risk of automated user enumeration. Keycloak does have mitigation strategies, however those only apply after a user tries to authenticate and thus don't work in this particular case. This PR adds a rate limit at the traefik layer to only allow 10 requests/second. There are additional settings that can be [leveraged](https://doc.traefik.io/traefik/middlewares/http/ratelimit/) but this should be a good start.

# How can this be tested?

Deploy to CI and test exceeding the rate limit.

### DIFF
--- a/src/bilder/images/keycloak/files/docker-compose.yaml
+++ b/src/bilder/images/keycloak/files/docker-compose.yaml
@@ -23,6 +23,8 @@ services:
     image: traefik:v2.10
     command:
     - "--configFile=/etc/traefik/traefik.yaml"
+    labels:
+    - "traefik.http.middlewares.keycloak-ratelimit.ratelimit.average=10"
     ports:
     - "80:80"
     - "443:443"


Ardiea (approved):

src/bilder/images/keycloak/files/docker-compose.yaml
```diff
@@ -23,6 +23,8 @@ services:
     image: traefik:v2.10
     command:
     - "--configFile=/etc/traefik/traefik.yaml"
+    labels:
+    - "traefik.http.middlewares.keycloak-ratelimit.ratelimit.average=10"
```
blarghmatey, line 27: Do we want this applied at the Traefik layer, or in the Keycloak service settings?

shaidar: Don't think it matters in this case given that it's only servicing keycloak in this case.

PR #1969, opened 2023-12-08
config: set Zendesk contact URLs for MIT xPRO and MITx Online

arslanashraf7:
https://github.com/mitodl/hq/issues/3094 & there is no ticket for MITx Online

- Sets the Zendesk new request URL as a contact URL for both xPRO and MITxOnline

# Why add this setting?

When a Marketing site is enabled for an edX instance, The platform uses `MKTG_URLS` dictionary to get the helpful links so that they can point to marketing sites. There are more details on https://github.com/openedx/edx-platform/wiki/Alternate-site-for-marketing-links.

In the case of the `CONTACT` link, it is also picked from `MKTG_URLS` so I created this PR to override the contact value.

None

# How can this be tested?
Once deployed, The `Contact Us` links in the 404 and other pages should take the users to our app's new request Zendesk page

### DIFF
--- a/src/bilder/images/edxapp_v2/templates/edxapp/mitxonline/common_values.yml.tmpl
+++ b/src/bilder/images/edxapp_v2/templates/edxapp/mitxonline/common_values.yml.tmpl
@@ -362,6 +362,7 @@ MKTG_URL_OVERRIDES:
     ABOUT: https://{{ key "edxapp/marketing-domain" }}/about-us/
     HONOR: https://{{ key "edxapp/marketing-domain" }}/honor-code/
     ACCESSIBILITY: https://accessibility.mit.edu/
+    CONTACT: https://mitxonline.zendesk.com/hc/en-us/requests/new/
     TOS_AND_HONOR: ''
 MOBILE_STORE_URLS: {}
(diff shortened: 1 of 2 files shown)


arslanashraf7:
> Which MFE/MFE components use `CONTACT`?
> 

I don't think the MFEs use this `CONTACT`, they have their setting `CONTACT_URL`. e.g. [Learning MFE](https://github.com/openedx/frontend-app-learning/blob/master/.env#L8) and all others that use `CONTACT_URL` would get it from [these settings in ol-infrastructure](https://github.com/search?q=repo%3Amitodl%2Fol-infrastructure%20CONTACT_URL&type=code).

> I'm concerned that in a year's time we won't know what this does or where it came from. Do you have any thoughts on where we could leave ourselves a breadcrumb? Is it enough to leave some details on this PR, or in the commit message?

I've updated the PR description with why we are adding this setting. There are more details on https://github.com/openedx/edx-platform/wiki/Alternate-site-for-marketing-links.

pdpinch:
Perfect, thank you!

blarghmatey (approved):

Issue #1968, opened 2023-12-07
Provide Superset as a service for data exploration


blarghmatey:
# User Story

- As a data platform owner, I want to deploy an installation of Superset, so I can provide data consumers with a portal for exploration and visualization

# Description/Context

Deploy a customized installation of Superset on our own AWS infrastructure

# Acceptance Criteria

- [x] Superset is deployed
- [x] Superset is integrated with Keycloak for Authn/Authz
- [x] Superset is connected to Trino for querying the data lake
- [ ] dbt mart models are published to Superset as data sets

# Plan/Design

- [x] Create a Pulumi project for managing the AWS infrastructure to support Superset
- [x] Create a customized Docker image for our Superset installation
  - [x] Custom Python settings module and plugins
  - [x] Specify additional Python dependencies and versions
- [ ] Integrate dbt with Superset following examples from https://engineering.hometogo.com/how-hometogo-connected-dbt-and-superset-to-make-metadata-more-accessible-and-reduce-analytical-2223af539cc6

PR #1971, opened 2023-12-08
Deploy superset

blarghmatey:
#1968 

- Add Pulumi definitions for AWS infrastructure
- Add a customized docker image
- Add an AMI build for Superset
- Add a Concourse pipeline to manage deployment

# How can this be tested?
Visit https://bi-qa.ol.mit.edu and verify that the application is available


blarghmatey:
pre-commit.ci run

src/bilder/images/superset/superset.pkr.hcl
```diff
@@ -0,0 +1,93 @@
...
+variable "node_type" {
+  type    = string
+  default = "server"
+}
+
+source "amazon-ebs" "superset" {
+  ami_description         = "Deployment image for Superset server generated at ${local.timestamp}"
+  ami_name                = "superset-${var.node_type}-${local.timestamp}"
+  ami_virtualization_type = "hvm"
+  instance_type           = "t3a.large"
```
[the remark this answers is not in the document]
blarghmatey: So, this instance type only applies to the AMI build, but yes, we will want to change the instance types for the ASG once we start relying on this.

src/bilder/images/superset/files/docker-compose.yaml
```diff
@@ -0,0 +1,84 @@
...
+    - 8088:8088
+    depends_on:
+      superset-init:
+        condition: service_completed_successfully
+    volumes: *superset-volumes
+    labels:
+    - "traefik.enable=true"
+    - "traefik.http.routers.superset.tls=true"
+    - "traefik.http.routers.superset.entrypoints=https"
+    - "traefik.http.routers.superset.rule=(Host(`${DOMAIN:-bi-qa.ol.mit.edu}`) ||\
```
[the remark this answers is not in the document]
blarghmatey: Probably not? I'll remove it.

src/ol_infrastructure/applications/edxapp/__main__.py
```diff
@@ -1018,7 +1018,11 @@
 
 edxapp_acm_cert_validation_records = (
     edxapp_web_acm_cert.domain_validation_options.apply(
-        partial(acm_certificate_validation_records, zone_id=edxapp_zone_id)
+        partial(
```
shaidar, line 1021: I know this is just a formatting change, but do we want this to be in the PR?

blarghmatey: We do, because I actually changed the signature of the method being used, which is what caused the formatting to get bumped to multiple lines.

src/ol_superset/requirements.txt
```diff
@@ -0,0 +1,468 @@
+#
```
shaidar, line 1: We can't use pyproject.toml instead?

blarghmatey: Not really, because we need to generate a requirements file that is included as part of the image build for Superset. I started by including it as part of pyproject.toml, but it caused downgrades of packages that we rely on for other things. This seemed like the best option to have the requirements that we need for Superset, while also having a lock-file, and being able to optionally install them in a local venv for purposes of auto-complete during development.

src/ol_infrastructure/applications/superset/__main__.py
```diff
@@ -0,0 +1,742 @@
...
+    listener_cert_domain=superset_domain,
+    listener_cert_arn=superset_web_acm_cert.arn,
+)
+
+superset_tg_config = OLTargetGroupConfig(
+    vpc_id=data_vpc["id"],
+    health_check_interval=60,
+    health_check_matcher="200-399",
+    health_check_path="/health",
+    health_check_unhealthy_threshold=3,  # give extra time for airbyte to start up
```
Ardiea, line 499: 🍝

src/ol_infrastructure/applications/superset/__main__.py
```diff
@@ -0,0 +1,742 @@
...
+                            GRAFANA_CLOUD_PROMETHEUS_API_USER={grafana_credentials['prometheus_user_id']}
+                            GRAFANA_CLOUD_LOKI_API_USER={grafana_credentials['loki_user_id']}
+                            """
+                                ),
+                                "owner": "root:root",
+                            },
+                            {
+                                "path": "/etc/default/superset",
+                                "content": f"DOMAIN={superset_domain}",
+                            },
```
Ardiea, line 571: I don't see where this is ever used?

src/ol_infrastructure/applications/superset/__main__.py
```diff
@@ -0,0 +1,742 @@
...
+                            GRAFANA_CLOUD_PROMETHEUS_API_USER={grafana_credentials['prometheus_user_id']}
+                            GRAFANA_CLOUD_LOKI_API_USER={grafana_credentials['loki_user_id']}
+                            """
+                                ),
+                                "owner": "root:root",
+                            },
+                            {
+                                "path": "/etc/default/superset",
+                                "content": f"DOMAIN={superset_config.get('domain')}",
+                            },
```
Ardiea, line 687: Same here, I don't think this is ever used, is it?

src/ol_infrastructure/applications/superset/__main__.py
```diff
@@ -0,0 +1,742 @@
...
+                                "content": "COMPOSE_PROFILES=worker",
+                            },
+                            {
+                                "path": "/etc/profile",
+                                "content": "export COMPOSE_PROFILES=worker",
+                                "append": True,
+                            },
+                            {
+                                "path": "/etc/docker/compose/.env",
+                                "content": f"DOMAIN={superset_domain}\nVAULT_ADDR=https://vault-{stack_info.env_suffix}.odl.mit.edu\n",
```
Ardiea, line 699: I believe it is this reference to DOMAIN that gets interpolated into the traefik labels in the docker-compose.yaml file.

Ardiea: Also, do the workers need this at all?

src/ol_superset/docker-compose.yaml
Ardiea, line 1: What is this for? Local dev via `docker compose build .` ?

src/ol_superset/pythonpath/auth/keycloak.py
```diff
@@ -0,0 +1,73 @@
...
+
+
+class OIDCSecurityManager(SupersetSecurityManager):
+    def __init__(self, appbuilder):
+        super().__init__(appbuilder)
+        if self.auth_type == AUTH_OID:
+            self.oid = OpenIDConnect(self.appbuilder.get_app)
+        self.authoidview = AuthOIDCView
+
+
```
Ardiea, line 22:
```suggestion
# Reference: https://superset.apache.org/docs/security/
```

src/ol_superset/pythonpath/superset_config.py
```diff
@@ -0,0 +1,223 @@
...
+}
+
+# Set this API key to enable Mapbox visualizations
+MAPBOX_API_KEY = os.environ.get("MAPBOX_API_KEY", "")
+
+
+class CeleryConfig:  # pylint: disable=too-few-public-methods
+    broker_url = "rediss://superset-redis.service.consul:6379/1"
+    imports = ("superset.sql_lab", "superset.tasks.scheduler")
+    result_backend = "rediss://superset-redis.service.consul:6379/2"
```
Ardiea, line 180: Should we provide a way to override this with an env var as we do with `VAULT_ADDR` at the top of the file? There are a few more URLs scattered in this file that maybe could use that consideration as well.

src/ol_superset/pythonpath/vault/aws_auth.py
```diff
@@ -0,0 +1,185 @@
+import logging.handlers
```
Ardiea, line 1:
```suggestion
# Reference: https://hvac.readthedocs.io/en/stable/usage/auth_methods/aws.html#ec2-authentication

import logging.handlers
```

Ardiea (approved):

src/ol_infrastructure/applications/superset/__main__.py
blarghmatey: Yes, I believe the workers need to know the domain to ensure that e.g. emails have proper values.

src/ol_superset/docker-compose.yaml
blarghmatey: In theory, yes. It was copy :spaghetti: to start with and I was debating whether to keep it.

src/ol_infrastructure/applications/superset/__main__.py
blarghmatey: It's not. Instead it's adding the domain to the .env file directly. I'll remove this to avoid confusion.

src/ol_superset/pythonpath/superset_config.py
blarghmatey: We will likely want to have a means of toggling between prod-level and local URLs to support non-prod environments for testing, but given that this is a Consul service address it shouldn't need to be mutable.

PR #1975, opened 2023-12-11
Managing the tika/access-token value in vault

Ardiea:
Addresses an issue reported in slack with MITOpen

This will manage the tika access-token in vault automatically. Previously it was manually configured.

### DIFF
--- a/src/ol_infrastructure/applications/tika/__main__.py
+++ b/src/ol_infrastructure/applications/tika/__main__.py
@@ -173,6 +173,13 @@ x_access_token = read_yaml_secrets(Path(f"tika/tika.{stack_info.env_suffix}.yaml
     "x_access_token"
 ]
 
+# Store the access token in vault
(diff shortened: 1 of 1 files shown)


src/ol_infrastructure/applications/tika/__main__.py
```diff
@@ -173,6 +173,13 @@
     "x_access_token"
 ]
 
+# Store the access token in vault
+vault.generic.Secret(
+    "tika-server-x-access-token-vault-secret",
+    path="secret-operations/tika/access-token",
+    data_json=f'{{"value": "{x_access_token}"}}',
```
blarghmatey, line 180: Why not do this as `json.dumps({"value": x_access_token})`?

Ardiea: 🤷

blarghmatey (approved):

PR #2009, opened 2023-12-19
Change Keycloak domains to be rooted at ol.mit.edu

blarghmatey:
N/A

The odl subdomain doesn't have any real bearing on the current organization. The ol subdomain is what we want to use for exposing services to end users.


Ardiea (approved):

shaidar:
A few other things need to be updated in addition to this PR:
- This will break the Touchstone integration until IST updates their end.
- Updates needed here - https://github.com/mitodl/salt-ops/blob/main/pillar/heroku/mitopen.sls
- A few docs that reference the URL

Will start working on those issues.

Issue #1979, opened 2023-12-14
Tag openedx AMIs with code information


Ardiea:
# Description/Context
As a devops engineer, I want to easily determine what code is running in any given edxapp deployment.

Populate tags on the AMI for the following

- [ ] upstream edxapp repo address
- [ ] upstream edxapp repo branch
- [ ] upstream edxapp commit hash
- [ ] upstream theme repo address
- [ ] upstream theme repo branch
- [ ] upstream theme commit hash
- [ ] docker image repo 
- [ ] docker image tag (?)
- [ ] docker image hash

# Plan/Design

Issue #2033, opened 2024-01-02
Tika version hard-coded


Ardiea:
## Expected Behavior
Actually use the tika version information from the pipeline triggers/resources when building a new tika image. 

## Current Behavior
The tika pipeline is configured to build a new version of the AMI when a new Tika container is release by apache foundation: https://github.com/mitodl/ol-infrastructure/blob/main/src/ol_concourse/pipelines/infrastructure/tika/packer_pulumi_pipeline.py#L11-L14

However that version information is never sent to / utilized by the packer build step when making the AMI and setting up `.env` or `docker-compose.yaml`. The version of the tika container is hard coded here: https://github.com/mitodl/ol-infrastructure/blob/main/src/bilder/images/tika/files/docker-compose.yaml#L22

## Possible Solution

It should be similar to edxapp where we pass an env var through the pipelines to the py infra code for `DOCKER_IMAGE_DIGEST` and `DOCKER_IMAGE_NAME` but maybe don't need to split it up and we would use the tag rather than the digest.

feoh:
Are there any developer / user ramifications to this?

Like, I get the impression that some semantics can change between Tika versions.

Ardiea:
I think maybe between major versions, yes? But I think within a major version there shouldn't be ramifications.

PR #2041, opened 2024-01-04
Software Version metadata tagging for AMIs

Ardiea:
Closes https://github.com/mitodl/ol-infrastructure/issues/1979

This PR adds and documents a mechanism for tagging AMI images with metadata about the software installed on them, both 3rd party software as well as our applications. It sets up this tagging for two of our images, `edxapp` and `tika` and provides documentation/pattern/blueprint for completing the rest of our images. 

image
image

# How can this be tested?
Tika:
```
pr packer build -only amazon-ebs.tika -var node_type=server .
```
edxapp is more complicated because you need to provide a docker image digest. I just pulled one from concourse:
```
DOCKER_IMAGE_DIGEST="sha256:c1bcbebe8738f2ac3cc58451f0a0dcbd71ab903178465fa36f8422d621e7418d" pr packer build -only amazon-ebs.edxapp -var openedx_release=master -var node_type=web -var-file=packer_vars/mitxonline.pkrvars.hcl -var-file=packer_vars/master.pkrvars.hcl  .
```

### DIFF
--- a/src/bilder/images/tika/deploy.py
+++ b/src/bilder/images/tika/deploy.py
@@ -1,4 +1,6 @@
+import json
 import os
+from io import StringIO
 from pathlib import Path
 
 from bridge.lib.magic_numbers import VAULT_HTTP_PORT
@@ -40,6 +42,7 @@ from bilder.components.vector.steps import (
     vector_service,
 )
 from bilder.facts.has_systemd import HasSystemd
+from bilder.lib.ami_helpers import build_tags_document
 from bilder.lib.linux_helpers import DOCKER_COMPOSE_DIRECTORY
 
 VERSIONS = {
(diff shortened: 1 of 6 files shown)


src/bilder/images/tika/deploy.py
```diff
@@ -145,6 +149,24 @@
 for product in hashicorp_products:
     configure_hashicorp_product(product)
 
+# Place the tags document
+tags_json = json.dumps(
+    build_tags_document(
+        source_tags={
+            "consul_version": VERSIONS["consul"],
+            "vault_version": VERSIONS["vault"],
+            "traefik_version": VERSIONS["traefik"],
```
blarghmatey, line 158: Should this also include the Tika version?

blarghmatey (approved):

src/bilder/images/tika/deploy.py
Ardiea: So, that is in an interesting question/point and comes up in an issue I opened yesterday as I was working on this and doing my experimentation with Tika. https://github.com/mitodl/ol-infrastructure/issues/2033 The tika version we are using is actually hard coded at the moment, so while I could certainly include that hard coded value here I figured it made more sense to address it outright in its own issue and fix the greater problem.

blarghmatey: Makes sense, I'd rather get this merged and address in future work.

PR #2047, opened 2024-01-05
Create new Keycloak realm

shaidar:

Part of https://github.com/mitodl/ol-infrastructure/issues/1968

Creates a new Keycloak realm to be used for BI apps for inside group and potentially outsiders.

### DIFF
--- a/src/ol_infrastructure/substructure/keycloak/__main__.py
+++ b/src/ol_infrastructure/substructure/keycloak/__main__.py
@@ -274,6 +274,103 @@ ol_platform_engineering_rsa_key = keycloak.RealmKeystoreRsa(
 )
 """
 
+# Create OL Data Platform Realm
+ol_data_platform_realm = keycloak.Realm(
+    "ol-data-platform",
+    access_code_lifespan="30m",
+    access_code_lifespan_user_action="15m",
+    attributes={
+        "business_unit": f"operations-{env_name}",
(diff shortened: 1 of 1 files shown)


src/ol_infrastructure/substructure/keycloak/__main__.py
```diff
@@ -274,6 +274,103 @@
...
+    duplicate_emails_allowed=False,
+    otp_policy=keycloak.RealmOtpPolicyArgs(
+        algorithm="HmacSHA256",
+        digits=6,
+        initial_counter=2,
+        look_ahead_window=1,
+        period=30,
+        type="totp",
+    ),
+    realm="ol-data-platform",
```
blarghmatey, line 298: The only thing I want to think/discuss more about is whether to explicitly tie this to the data platform, or consider it to be a bit broader in scope/purpose.

shaidar: I thought about that and felt that it's important with the data involved to be very explicit about this particular realm and have it be very clear.

blarghmatey: I agree and that makes sense. I just don't want to do unnecessary proliferation of realms if we can avoid it. It definitely makes sense to keep this one narrowly scoped.

PR #2073, opened 2024-01-08
config: disable new problem editor for MITx Online QA

arslanashraf7:
https://github.com/mitodl/hq/issues/3299 more precisely https://github.com/mitodl/hq/issues/3299#issuecomment-1879262462

- Disables the Authoring MFE for problem editors as per https://github.com/mitodl/hq/issues/3299#issuecomment-1879262462

### How can this be tested?
- We should not see https://github.com/mitodl/hq/issues/3299 -- Editing or creating a problem on https://studio-qa.mitxonline.mit.edu/ should work fine.


arslanashraf7:
> This will keep it from being re-enabled but someone will need to manually remove the flag in the UI once this is applied. 👍

I've already disabled it there manually.

PR #2130, opened 2024-01-30
Md/gh ratelimit fixes

Ardiea:
https://github.com/mitodl/ol-concourse-github-issues/issues/62

 Updated the gh issues resource defaults to use app auth rather than token. Added app auth config to vault mount points for concourse.


src/ol_concourse/lib/resources.py
```diff
@@ -91,14 +91,18 @@ def github_issues(  # noqa: PLR0913
     name: Identifier,
     repository: str,
     issue_prefix: str,
+    auth_method: Literal["token", "app"] = "token",
     access_token: str = "((github.issues_resource_access_token))",  # noqa: S107
+    app_id: str = "((github.issues_resource_app_id))",
+    app_installation_id: str = "((github.issues_resource_app_installation_id))",
+    private_ssh_key: str = "((github.issues_resource_private_ssh_key))",
```
blarghmatey, line 98: Where is this value coming from?

src/ol_concourse/lib/resources.py
```diff
@@ -91,14 +91,18 @@ def github_issues(  # noqa: PLR0913
     name: Identifier,
     repository: str,
     issue_prefix: str,
+    auth_method: Literal["token", "app"] = "token",
```
blarghmatey, line 94: Should we just default this to "app" rather than having to override it explicitly in other places?

feoh (approved):

src/ol_concourse/lib/resources.py
Ardiea: Like where did the key come from? Github generates it downloads it when you setup an application. It is being stored into vault else where in this PR.

src/ol_concourse/lib/resources.py
Ardiea: Maybe? I was operating on the "don't break what is currently running" mindset but I'm not sure that applies here.

PR #2137, opened 2024-02-09
feat: add learning base url

asadali145:

Fixes https://github.com/mitodl/hq/issues/3475

- Adds a new config `LEARNING_BASE_URL`.
- Adds it for all the apps and all environments.

### How can this be tested?

Add it to the .env.development for the authoring MFE and test it.


src/ol_concourse/pipelines/open_edx/mfe/pipeline.py
```diff
@@ -84,6 +85,7 @@ def mfe_params(
             f"{open_edx.environment}-open-edx-language-preference"
         ),
         "LMS_BASE_URL": f"https://{open_edx.lms_domain}",
+        "LEARNING_BASE_URL": f"https://{open_edx.learning_base_url}",
```
arslanashraf7, line 88: I think we can make this change more robust like. This way we would not need to add this to `src/ol_concourse/pipelines/open_edx/mfe/values.py`. So, the only change would be this one line.

```suggestion
        "LEARNING_BASE_URL": f"https://{open_edx.learning_base_url}/{mfe.application.path}",
```
OR

```suggestion
        "LEARNING_BASE_URL": f"https://{open_edx.learning_base_url}/learn",
```

arslanashraf7 (changes requested):
I requested some changes, But it would be best to get opinion from @mitodl/devops as well.

src/ol_concourse/pipelines/open_edx/mfe/pipeline.py
asadali145: Yeah, this 2nd suggestion seems best. I am updating it.

asadali145: I am not sure about a way to get the learning MFE path dynamically here. @mitodl/devops Could suggest one maybe?

blarghmatey: I just pushed a commit that will automatically pull in the proper value. I started to write out how to do it, but it was easier to just show you with code :)

asadali145: Thanks, This makes much more sense now.

arslanashraf7 (approved):

PR #2162, opened 2024-02-12
Concourse publishing pipeline for open-api-clients

rhysyngsun:
Closes https://github.com/mitodl/mit-open/issues/269

This implements a Concourse pipeline that generates new versions of API clients of MIT Open and publishes them.

- Triggers only when a change occurs:
  - In the OpenAPI specs in the mit-open repo's `release` (production) branch.
  - In the open-api-clients repo itself.
- The version gets bumped automatically using [bumpver](https://pypi.org/project/bumpver/) and a date-based versioning scheme.
- We're only currently generating a release to npm

I'm less than sure about how I setup the credentials, but it's an educated attempt based on what else I see in the repo. I'm also unsure if there's other work that has to be done to get values into concourse itself.

Screenshot 2024-02-12 at 16-01-32 api-clients-release - Concourse
(ignore that these are all failing at the time of screenshot)

### How can this be tested?

You can run this locally, although I haven't tested this personally with the credentials parts as I'm unsure how to set that up locally.

### DIFF
--- a/src/ol_concourse/pipelines/libraries/open_api_clients.py
+++ b/src/ol_concourse/pipelines/libraries/open_api_clients.py
@@ -1,0 +1,232 @@
+import sys
+from pathlib import Path
+
+from ol_concourse.lib.models.pipeline import (
+    Command,
+    GetStep,
+    Identifier,
+    InParallelStep,
+    Input,
+    Job,
+    Output,
+    Pipeline,
+    PutStep,
+    Resource,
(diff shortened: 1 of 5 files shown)


rhysyngsun:
@feoh I think this should be all set now, at least it validates, generates the definition file, and concourse is happy enough to load the file. Since it's using Python code i moved the small scripts into actual shell script files so that shellcheck can check those too and read those in.

PR #2191, opened 2024-02-21
Resize residential worker nodes

Ardiea:
I did this on the console yesterday. Codifying now. 

https://github.mit.edu/mitx/residential-mitx/issues/44
https://github.com/mitodl/hq/issues/3636

Upgrading mitx-production-worker nodes to m7a.xlarge to address crashing job issues.

### DIFF
--- a/src/ol_infrastructure/applications/edxapp/Pulumi.applications.edxapp.mitx.Production.yaml
+++ b/src/ol_infrastructure/applications/edxapp/Pulumi.applications.edxapp.mitx.Production.yaml
@@ -53,7 +53,7 @@ config:
   edxapp:web_node_capacity: "9"
   edxapp:worker_disk_size: "50"
   edxapp:worker_node_capacity: "5"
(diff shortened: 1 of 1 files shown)


pdpinch:
Do we have a common size for worker nodes across xpro, residential and mixonline?

Ardiea:
> Do we have a common size for worker nodes across xpro, residential and mixonline?

Sort of / not really.

Res prod: m7a.xlarge
xpro prod: m7a.medium
mitxonline prod: c6a.large
staging prod: t3a.medium

everything else: t3a.medium 

From looking at the charts and stuff, c6a.large was fine for residential prod 99% of the time except when this canvas export ran which tipped over the memory (c6a.large is only 4GB memory), which then threw the node into a death spiral and it would eventually crash. 

We could probably create a swap file during the build process to give the node the virtual memory it would need for these rare edge cases and save some money $$$ but it would be an investigation / effort.

feoh (approved):

PR #2247, opened 2024-03-06
change default mtixonline email addres to no-reply

pdpinch:

follow up to https://github.com/mitodl/hq/issues/669#issuecomment-1980593538

Changes the default "from" address to no-reply on QA so we can test the impact of this change. 

### How can this be tested?

Generate some emails from open edX using the instructor dashboard.

### DIFF
--- a/src/ol_infrastructure/applications/edxapp/__main__.py
+++ b/src/ol_infrastructure/applications/edxapp/__main__.py
@@ -918,7 +918,8 @@ consul_kv_data = {
         else "false"
     ),
     "bulk-email-default-from-email": (
-        edxapp_config.get("bulk_email_default_from_email") or edxapp_config.require("sender_email_address")
+        edxapp_config.get("bulk_email_default_from_email")
+        or edxapp_config.require("sender_email_address")
     ),
     "ses-configuration-set": f"edxapp-{env_name}",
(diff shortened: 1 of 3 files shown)


Ardiea (approved):

src/ol_infrastructure/applications/edxapp/__main__.py
```diff
@@ -917,6 +917,10 @@
         if edxapp_config.get_bool("email_use_course_id_from_for_bulk")
         else "false"
     ),
+    "bulk-email-default-from-email": (
+        edxapp_config.get("bulk_email_default_from_email")
+        or edxapp_config.require("sender_email_address")
+    ),
```
mudassir-hafeez, line 923: It should use the configured default bulk email address if available.
For QA instances, use the no-reply email address as the default i.e., edxapp:bulk_email_default_from_email: no-reply@edxapp-mail-qa.mitxonline.mit.edu).
Other environments should retain the old configured email address for bulk emails until the new configuration is thoroughly tested and set.

mudassir-hafeez:
@pdpinch @Ardiea would you please review the change?

**Note:** Reviewer has to make sure it should be enabled for mitxonline QA instance, other environments should retain the old configured email address for bulk emails until the new configuration is thoroughly tested and set.

PR #2269, opened 2024-03-13
Adding a custom macro to get the current user email.

quazi-h:
Unblocks https://github.com/mitodl/ol-infrastructure/issues/1968

Defined a new macro to return user email from the flask app in our superset config.

Superset's function hasn't been released yet and so isn't available in our instance.
See Superset's own code here:
[Jinja_context](https://github.com/apache/superset/blob/master/superset/jinja_context.py#L136-L148)
[Superset Utils](https://github.com/Vitor-Avila/superset/blob/40b0c78dd1e0b003e34361ec6920bdc36c279d5b/superset/utils/core.py#L1387-L1397)
[Superset Docs on Jinja Templating and adding custom macros](https://superset.apache.org/docs/installation/sql-templating/)

### How can this be tested?
Try running `SELECT {{ current_user_email() }}` in the SQL Lab on our Superset instance.

### DIFF
--- a/src/ol_superset/pythonpath/superset_config.py
+++ b/src/ol_superset/pythonpath/superset_config.py
@@ -263,9 +263,9 @@ SLACK_API_TOKEN = vault_client.secrets.kv.v2.read_secret(
     path="app-config", mount_point="secret-superset"
 )["data"]["data"]["slack_token"]
 
-#############################
-# Custom Template Processors #
-############################
+#######################
+# Custom Jinja Macros #
+#######################
 # Enable usage of current_user_email() macro until Superset releases that feature
 
 
--- a/src/ol_superset/pythonpath/superset_config.py
+++ b/src/ol_superset/pythonpath/superset_config.py
@@ -266,9 +266,8 @@ SLACK_API_TOKEN = vault_client.secrets.kv.v2.read_secret(
 #######################
 # Custom Jinja Macros #
 #######################
-# Enable usage of current_user_email() macro until Superset releases that feature
-
-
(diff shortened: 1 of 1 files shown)


src/ol_superset/pythonpath/superset_config.py
```diff
@@ -260,3 +262,24 @@ class CeleryConfig:  # pylint: disable=too-few-public-methods
...
+    :returns: The email
+    """
+
+    try:
+        return g.user.email
+    except Exception:  # pylint: disable=broad-except
+        return None
+
+
+JINJA_CONTEXT_ADDONS = {"current_user_email": current_user_email()}
```
blarghmatey, line 285:
```suggestion
JINJA_CONTEXT_ADDONS = {"current_user_email": current_user_email}
```
We want to set the function object in the context. Adding the parentheses will set the result of the function call.

blarghmatey (changes requested):

src/ol_superset/pythonpath/superset_config.py
```diff
@@ -260,3 +262,26 @@ class CeleryConfig:  # pylint: disable=too-few-public-methods
...
+def current_user_email() -> Optional[str]:
+    """
+    Get the email (if defined) associated with the current user.
+
+    :returns: The email
+    """
+
+    try:
+        return g.user.email
+    except Exception:  # pylint: disable=broad-except
```
blarghmatey, line 281:
```suggestion
    except Exception:  # noqa: BLE001 
```

blarghmatey (approved):

PR #2282, opened 2024-03-19
Allow use of locally cloned edX and theme repos in Concourse

blarghmatey:
N/A

When trying to pin a commit for an edx-platfom build it is ignored because of cloning from the target branch as part of the Earthly steps. This allows for passing the locally cloned repo so that pinned commits in the pipeline are properly handled.

### How can this be tested?
Try running the earthly build using the newly added args and see if the cloned repos are used

### Additional Context
This is to address a bug introduced in the master branch of edx-platform that is currently blocking testing of MITx online

### DIFF
--- a/src/ol_concourse/pipelines/open_edx/edx_platform_v2/earthly_packer_pulumi_pipeline.py
+++ b/src/ol_concourse/pipelines/open_edx/edx_platform_v2/earthly_packer_pulumi_pipeline.py
@@ -157,7 +157,11 @@ def build_edx_pipeline(release_names: list[str]) -> Pipeline:  # noqa: ARG001
                             # Use some cleverness with path to mount resources within
                             # the earthly git resource so code is where the Earthfile
                             # expects
-                            inputs=[Input(name=earthly_git_resource.name)],
+                            inputs=[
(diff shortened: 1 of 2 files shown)


src/ol_concourse/pipelines/open_edx/edx_platform_v2/earthly_packer_pulumi_pipeline.py
```diff
@@ -157,7 +157,11 @@ def build_edx_pipeline(release_names: list[str]) -> Pipeline:  # noqa: ARG001
                             # Use some cleverness with path to mount resources within
                             # the earthly git resource so code is where the Earthfile
                             # expects
-                            inputs=[Input(name=earthly_git_resource.name)],
+                            inputs=[
+                                Input(name=earthly_git_resource.name),
+                                Input(edx_platform_git_resource.name),
+                                Input(theme_git_resource.name),
```
Ardiea, line 163:
```suggestion
                                Input(name=edx_platform_git_resource.name),
                                Input(name=theme_git_resource.name),
```

Ardiea (approved):

PR #2303, opened 2024-03-26
Add npm-package resource to Pydantic ol_concourse Python wrapper / models and openapi_clients pipeline.

### LINKED ISSUE #2190: Pipelines should be able to publish nodejs packages to NPM
### Description/Context

Right now we have no way of allowing pipelines to publish to NPM.

This isn't great, and forces devs to build and publish manually on their own machines.

### Plan/Design

We could use this [concourse resource](https://github.com/timotto/concourse-npm-resource) to accomplish the task.

We'd just need to plumb the appropriate token through Vault.

Completion of this task should include a sample pipeline , or maybe even just finish off the NPM publishing piece of https://github.com/mitodl/ol-infrastructure/blob/main/src/ol_concourse/pipelines/libraries/open_api_clients.py

feoh:
Fixes #2190

This modifies @rhysyngsun's open-apilcinets axios publishing pipeline to be more Concourse like and to actually publish to the npm registry.

NOTE: Along the way we created python wrappers for the concourse-npm-resource and semver packages, and we're leaving those in even though we're not using them at this time. No point in undoing all that work and we may choose to use them in future.

### How can this be tested?

Add the npm package resource to the publish step of any pipeline, being sure to specify the registry token. An appropriate token lives in Vault production.


src/ol_concourse/lib/resources.py
```diff
@@ -257,3 +257,19 @@ def slack_notification(name: Identifier, url: str) -> Resource:
...
+    return Resource(
+        name=package,
+        type="npm-package",
+        icon="language-javascript",
+        source={
+            "package": package,
+            "scope": scope,
+            "registry": {
+                "uri": "https://registry.npmjs.org",
+                "token": "{{npmjs.concourse_token}}",
```
Ardiea, line 272: I don't think this is right. Refer to lines 214 + 215 above for how it is done with pypi. https://github.com/mitodl/ol-infrastructure/pull/2303/files#diff-ab9c861e659afe118ba16db129458952fb2ddbf8f5c3104c58864b89b1733c1cR215-R216. 

I also don't see the secret(s) in vault production?  You can add it to vault automatically by putting it in the sops file and then this block in the concourse pulumi stack will take care of it: https://github.com/mitodl/ol-infrastructure/pull/2303/files#diff-ab9c861e659afe118ba16db129458952fb2ddbf8f5c3104c58864b89b1733c1cR215-R216

It would be something like:

```yaml
pipelines:
    <team name that needs the token>/npmjs:
        token: <your super secret token> 
```

And then this line would be:

```
"token": "((npmjs.token))",
```

It will make more sense once you decrypt that file. You will need to do an entry for each team that needs it. _Most_ of our stuff is in the `infrastructure` team.

Ardiea (changes requested):

src/ol_concourse/pipelines/libraries/open_api_clients.py
```diff
@@ -66,9 +69,9 @@ def _read_script(script_name: str) -> str:
 )
 mit_open_api_clients_repository = ssh_git_repo(
     name=Identifier("mit-open-api-clients"),
-    uri="git@github.com/mitodl/open-api-clients.git",
+    uri="git@github.com:mitodl/open-api-clients.git",
     branch="main",
-    private_key="((git-private-key))",
+    private_key="((open_api_clients.odlbot_private_ssh_key))",
```
blarghmatey, line 74: We should probably make this more general since we can make use of that key outside of just this pipeline.

src/ol_concourse/pipelines/libraries/open_api_clients.py
```diff
@@ -147,74 +117,55 @@ def _read_script(script_name: str) -> str:
...
-                    Input(name=mit_open_api_clients_repository.name),
-                ],
-                outputs=[Output(name=mit_open_api_clients_repository.name)],
-                run=Command(
-                    path="/bin/bash",
+                    dir="mit-open-api-clients",
                     args=[
                         "-exc",
-                        _read_script("open-api-clients-tag-release.sh"),
+                        ("scripts/open-api-clients-bumpver.sh"),
```
rhysyngsun, line 123: This is expecting the file to exist [here](https://github.com/mitodl/open-api-clients/tree/main/scripts), which it doesn't, it's a file local to this repository, which is why I had written that little `_read_script` function to inline it into command similar to how `bash -exc "$(cat scripts/open-api-clients-bumpver.sh)"` would work.

rhysyngsun: If it's easier to add this script to that repo, feel free to do so

src/ol_concourse/pipelines/libraries/scripts/open-api-clients-publish-node.sh
```diff
@@ -2,4 +2,4 @@
 # publish the npm package to npmjs.org
 
 yarn build
```
rhysyngsun, line 4: To fix `/bin/sh: 1: tsc: not found`:

```suggestion
yarn install --immutable
yarn build
```

src/ol_concourse/pipelines/libraries/open_api_clients.py
feoh: Agreed. Where should we move it?

feoh: Gonna say this is a problem for another day for now.

blarghmatey (approved):

PR #2318, opened 2024-04-01
Changed perms on Open bucket

shaidar:

This is a changed based on a [Slack discussion ](https://mitodl.slack.com/archives/C03K5HYGPT9/p1712002252226879). The bucket contains avatars and such that are loaded through Django admin and should be public visible.


blarghmatey (commented):
I don't think that this will work given the changes in permission definitions with the BucketV2 API updates.

shaidar:
That's still using v1 as I haven't changed the call to `s3.Bucket` and just added a policy to it.

blarghmatey (approved):

PR #2323, opened 2024-04-04
Configure olapps to use the ol email theme

rhysyngsun:

Related to https://github.com/mitodl/ol-keycloak/issues/48

A [recent PR](https://github.com/mitodl/ol-keycloak/pull/49) to `ol-keycloak` introduced a new email theme for keycloak emails to be used for the `olapps` realm. This PR adds the configuration to use it.

Docs: https://www.pulumi.com/registry/packages/keycloak/api-docs/realm/#email_theme_python


shaidar (approved):

collinpreston (commented):
Should we update the display name?  The referenced PR indicates using a different display name.

PR #2325, opened 2024-04-04
Superset role mapping

shaidar:

Part of https://github.com/mitodl/ol-infrastructure/issues/1968

This creates the two superset roles along with the mappings required to Moira groups. The Moira groups need to be created still. This should automatically map superset roles to Moira groups so that a person who is a member of the `ol-eng-data` group will automatically be a `superset_admin` when logged in through Touchstone via Keycloak.

### How can this be tested?

I ran it on CI and it appears to do what's expected.
### Additional Context

Not the idea way to code this, but went down the route of setting up a helper method and got somewhat lost in the details. Will need at some point to circle back and refactor some of the code, but this should do what is required for now.

### DIFF
--- a/src/ol_infrastructure/substructure/keycloak/Pulumi.substructure.keycloak.CI.yaml
+++ b/src/ol_infrastructure/substructure/keycloak/Pulumi.substructure.keycloak.CI.yaml
@@ -22,14 +22,14 @@ config:
   keycloak:okta_single_sign_on_service_url: https://dev-66940844.okta.com/app/dev-66940844_mitci_1/exkcshk0uw4gnJILX5d7/sso/saml
   keycloak:openid_clients:
   - client_info:
-      open-discusions: https://discussions-ci.odl.mit.edu
+      open-discusions: ["https://discussions-ci.odl.mit.edu"]
       open-local: "*"
     realm_name: olapps
   - client_info:
-      superset: https://bi-ci.ol.mit.edu
+      superset: ["https://bi-ci.ol.mit.edu", superset_admin, superset_alpha]
     realm_name: ol-data-platform
   - client_info:
-      leek: https://celery-monitoring-ci.odl.mit.edu
+      leek: ["https://celery-monitoring-ci.odl.mit.edu"]
     realm_name: ol-platform-engineering
   keycloak:url: https://sso-ci.ol.mit.edu
   vault:address: https://vault-ci.odl.mit.edu
--- a/src/ol_infrastructure/substructure/keycloak/Pulumi.substructure.keycloak.CI.yaml
+++ b/src/ol_infrastructure/substructure/keycloak/Pulumi.substructure.keycloak.CI.yaml
@@ -26,7 +26,7 @@ config:
       open-local: "*"
     realm_name: olapps
   - client_info:
-      superset: ["https://bi-ci.ol.mit.edu", superset_admin, superset_alpha]
+      superset: ["https://bi-ci.ol.mit.edu", "superset_admin", "superset_alpha"]
     realm_name: ol-data-platform
   - client_info:
       leek: ["https://celery-monitoring-ci.odl.mit.edu"]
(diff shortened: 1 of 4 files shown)


src/ol_infrastructure/substructure/keycloak/__main__.py
```diff
@@ -455,21 +455,51 @@
...
         openid_client = keycloak.openid.Client(
             f"{realm_name}-{client_name}-client",
+            name=f"ol-{client_name}-client",
             realm_id=realm_name,
             client_id=f"ol-{client_name}-client",
             enabled=True,
             access_type="CONFIDENTIAL",
             standard_flow_enabled=True,
-            valid_redirect_uris=[f"{client_url}/*"],
+            valid_redirect_uris=[f"{url}/*"],
```
blarghmatey, line 502: Should we make the `/*` part of the value passed in to allow for restricting to specific paths?

src/ol_infrastructure/substructure/keycloak/__main__.py
```diff
@@ -455,21 +455,51 @@
...
+    ),
+    # Map Moira group membership to superset role
+    # ol-eng-finance -> superset_alpha
+    keycloak.AttributeToRoleIdentityMapper(
+        "ol-data-saml-superset-alpha-ol-eng-finance",
+        realm=ol_data_platform_realm.id,
+        attribute_friendly_name="mitMoiraMemberOf",
+        identity_provider_alias=ol_data_platform_touchstone_saml_identity_provider.alias,
+        attribute_value="ol-eng-finance",
+        role="ol-superset-client.superset_alpha",
```
blarghmatey, line 480: Is this assuming that the client role already exists? Should we add code for managing creation of that state?

src/ol_infrastructure/substructure/keycloak/__main__.py
```diff
@@ -489,6 +519,15 @@
                 realm_name=realm_name,
             ).apply(json.dumps),
         )
+        if len(client_detail) > 1:
+            for role in client_detail[1:]:
```
blarghmatey, line 523: You can leave out the conditional, because if the list is only length 1 then `[1:]` will yield an empty list, short-circuiting the `for` loop.

src/ol_infrastructure/substructure/keycloak/Pulumi.substructure.keycloak.CI.yaml
```diff
@@ -22,14 +22,14 @@ config:
   keycloak:okta_single_sign_on_service_url: https://dev-66940844.okta.com/app/dev-66940844_mitci_1/exkcshk0uw4gnJILX5d7/sso/saml
   keycloak:openid_clients:
   - client_info:
-      open-discusions: https://discussions-ci.odl.mit.edu
+      open-discusions: ["https://discussions-ci.odl.mit.edu"]
       open-local: "*"
     realm_name: olapps
   - client_info:
-      superset: https://bi-ci.ol.mit.edu
+      superset: ["https://bi-ci.ol.mit.edu", "superset_admin", "superset_alpha"]
```
blarghmatey, line 29: Rather than making this a mixed-use list, should it be a dictionary object? That seems like a more extensible and maintainable means of managing the information associated with clients. That could also allow for specifying additional client roles to be created, etc.

src/ol_infrastructure/substructure/keycloak/__main__.py
blarghmatey: Did it cause the role to be created?

shaidar: Yeah the roles were created and the mapping seemed to have the correct client role associated with it

blarghmatey: Great! 👍

src/ol_infrastructure/substructure/keycloak/Pulumi.substructure.keycloak.CI.yaml
```diff
@@ -22,14 +22,14 @@ config:
   keycloak:okta_single_sign_on_service_url: https://dev-66940844.okta.com/app/dev-66940844_mitci_1/exkcshk0uw4gnJILX5d7/sso/saml
   keycloak:openid_clients:
   - client_info:
-      open-discusions: ["https://discussions-ci.odl.mit.edu"]
+      open-discusions: ["https://discussions-ci.odl.mit.edu/*"]
```
blarghmatey, line 25: We'll just need to make these changes in the QA and production stacks as well.

PR #2342, opened 2024-04-10
Updating the Superset icon to an OL branded logo

quazi-h:
Adding a local static image with the OL brand icon.
Updated superset_config to point to new app logo.

https://github.com/mitodl/ol-infrastructure/issues/1968

Updates the app icon in the Superset UI.


src/ol_superset/pythonpath/superset_config.py
```diff
@@ -35,7 +35,7 @@
 # --------------------------------
 APP_NAME = "MIT OL Business Intelligence"
 # Specify the App icon. Useful for white-labeling
-APP_ICON = "/static/assets/images/superset-logo-horiz.png"
+APP_ICON = "../static/assets/images/OL_sub-brand_lockup_one-line_rgb_black.png"
```
pdpinch, line 38: We'll get a version of this that says "Data Platform" above "Open Learning" 

@quazi-h does it have to be a .png, or would a .svg work?

quazi-h: Yes, I believe an .svg is an accepted icon format here

src/ol_superset/static/assets/images/OL_sub-brand_lockup_one-line_rgb_black.png
blarghmatey, line 1: Can you rename this to something a bit less verbose? Maybe just `ol-data-platform-logo.png`

src/ol_superset/pythonpath/superset_config.py
blarghmatey, line 38: I think this gets rendered as the path in the web UI for loading the image, so it shouldn't have the leading `../`. You'll also need to add a `COPY` command in the Dockerfile to place the logo into the final image at the proper location.

blarghmatey (changes requested):

pdpinch:
@quazi-h please use the attached svg logo:

DataPlatform-OL_sub-brand

src/ol_superset/Dockerfile
```diff
@@ -1,6 +1,7 @@
 FROM apachesuperset.docker.scarf.sh/apache/superset:${SUPERSET_TAG:-latest}
 COPY --link requirements.in requirements.txt
 COPY --link pythonpath /app/pythonpath
+COPY --link static /app/static
```
blarghmatey, line 4: This completely overrides the static directory in the docker image, which isn't what we want. Instead we just want to copy the file into the target directory.
```suggestion
COPY static/assets/images/ol-data-platform-logo.svg /app/static/assets/images/ol-data-platform-logo.svg
```

blarghmatey (approved):

PR #2363, opened 2024-04-19
standardize MITx Online support email address

pdpinch:

fixes https://github.com/mitodl/hq/issues/4051

mitxonline-support@mit.edu is a email support address that works. It forwards to ZenDesk.

AFAIK, there is no mail server at mitxonline.mit.edu, so support@mitxonline.mit.edu cannot work. 

### How can this be tested?

~~Apply for financial aid on a course on mitxonline and inspect the status email that is returned. ~~

Even easier: enroll in any course for free, e.g. https://mitxonline.mit.edu/courses/course-v1:MITxT+8.02.3x/ 

The return address on the enrollment email is currently support@mitxonline.mit.edu. After this change it should be mitxonline-support@mit.edu

### DIFF
--- a/src/ol_infrastructure/applications/mitxonline/__main__.py
+++ b/src/ol_infrastructure/applications/mitxonline/__main__.py
@@ -197,9 +197,9 @@ heroku_vars = {
     "MITX_ONLINE_ADMIN_EMAIL": "cuddle-bunnies@mit.edu",
     "MITX_ONLINE_DB_CONN_MAX_AGE": "0",
     "MITX_ONLINE_DB_DISABLE_SSL": "True",  # pgbouncer buildpack uses stunnel to handle encryption"
-    "MITX_ONLINE_FROM_EMAIL": "MITx Online <support@mitxonline.mit.edu>",
+    "MITX_ONLINE_FROM_EMAIL": "MITx Online <mitxonline-support@mit.edu>",
     "MITX_ONLINE_OAUTH_PROVIDER": "mitxonline-oauth2",
-    "MITX_ONLINE_REPLY_TO_ADDRESS": "MITx Online <support@mitxonline.mit.edu>",
+    "MITX_ONLINE_REPLY_TO_ADDRESS": "MITx Online <mitxonline-support@mit.edu>",
     "MITX_ONLINE_SECURE_SSL_REDIRECT": "True",
     "MITX_ONLINE_USE_S3": "True",
     "NODE_MODULES_CACHE": "False",


shaidar (commented):
Received two email responses when I emailed the address. One referencing MITx Online and the other mentioning MicroMasters

pdpinch:
> Received two email responses when I emailed the address. One referencing MITx Online and the other mentioning MicroMasters

I got that too. I think it's a configuration issue in ZenDesk. I don't think that should block this change though?

shaidar:
No, not a blocker, but I would fix the Zendesk side and verify it prior to merging this.

src/ol_infrastructure/applications/mitxonline/__main__.py
```diff
@@ -188,9 +188,9 @@
     "MITX_ONLINE_ADMIN_EMAIL": "cuddle-bunnies@mit.edu",
     "MITX_ONLINE_DB_CONN_MAX_AGE": "0",
     "MITX_ONLINE_DB_DISABLE_SSL": "True",  # pgbouncer buildpack uses stunnel to handle encryption"
-    "MITX_ONLINE_FROM_EMAIL": "MITx Online <support@mitxonline.mit.edu>",
+    "MITX_ONLINE_FROM_EMAIL": "MITx Online <mitxonline-support@mit.edu>",
```
jkachel, line 191: It looks like this isn't the setting that needs to change. Did some digging locally and there's one called `MAILGUN_FROM_EMAIL` that appears to be the one we use - for some reason I don't see this at all in the config here. (It does exist in Heroku, though.) Changing that has the desired effect. 

On RC this is set to `MITx Online <no-reply@mitxonline-rc-mail.mitxonline.mit.edu>` and on prod this is `MITx Online <no-reply@mail.mitxonline.mit.edu>`. 

(The reply-to setting is fine and is arguably more important.)

jkachel (changes requested):
See note. Since this is just the from address I can approve this and we can then fix the config settings for the from address in a separate PR if you'd like - email programs should respect the Reply-To setting so the From address is important but less so.

pdpinch:
@jkachel I don't understand. Should we add another setting, for `MAILGUN_FROM_EMAIL` or is there some other change that you're recommending?

I don't suppose we could make a change to code and get rid of some of these (apparently) duplicative settings?

jkachel:
> @jkachel I don't understand. Should we add another setting, for `MAILGUN_FROM_EMAIL` or is there some other change that you're recommending?
> 
> I don't suppose we could make a change to code and get rid of some of these (apparently) duplicative settings?

I think we have two potential paths here:
- Add `MAILGUN_FROM_EMAIL` here and set it appropriately; this will fix the issue immediately, and then we can remove the `MITX_ONLINE_FROM_EMAIL` setting from the MITx Online app.
- Continue with the change as-is (setting`MITX_ONLINE_FROM_EMAIL` to the proper email address), then update the app to use `MITX_ONLINE_FROM_EMAIL` over `MAILGUN_FROM_EMAIL`.

In my mind, the second option is the proper option - having there be extra settings is confusing and the `MITX_ONLINE_FROM_EMAIL` setting matches the other app-specific email settings. But the first route will fix the issue at hand faster.

src/ol_infrastructure/applications/mitxonline/__main__.py
pdpinch: Should we remove `MITX_ONLINE_FROM_EMAIL` then, and add `MAILGUN_FROM_EMAIL` ?

pdpinch:
Discovered this is still hanging out here. Seems like it should be merged. @jkachel what do you think?

jkachel (approved):
I think this is fine to merge as-is. 👍

jkachel:
FYI: I have https://github.com/mitodl/mitxonline/pull/2605 open to clean up the email settings in the app a bit. This is good to go even without that change, though. 

@shaidar can you take a quick look (or pass it on)? I rebased it so it should be fine but since I made changes to it someone else should take a look at it too.

blarghmatey (approved):

Issue #2164, opened 2024-02-13
Set up automatic cycling of OCW Concourse workers


gumaerc:
### Description/Context
Occasionally, OCW site pipelines will fail with messages similar to `run check: find or create container on worker ip-10-0-1-142: volume mounted to container is missing 'fa54e9cb-7178-4181-555d-8a372ce4e8ae' from worker 'ip-10-0-1-142'` or other similar error messages about workers not responding. This is generally solved by cycling out the worker instances. It would be nice to do this on a regular basis for OCW workers to prevent them getting "stale."

feoh:
This isn't resolved.

In order to add the 3AM start time @gumaerc signposted we need in order to avoid disruptng OCW production.

I thought this was an easy fix so I dived in but it's taken most of the day and right now my changes from the latest PR are failing (no definition for build_schedule) when I try to update the pipeline.

So I'm going to revert my changes to main. My work on this up to this point can be seen in https://github.com/mitodl/ol-infrastructure/pull/2166

Whenever someone picks this up, somehow we need additional code here to actually include the definition for the build_schedule resource, because right now when I try to apply this I get:
```

pipeline name: instance-refresh-concourse

apply configuration? [yN]: y
error: invalid pipeline config:
invalid jobs:
        jobs.ci-worker-infra-instance-refresh.plan.do[0].get(build-schedule): unknown resource 'build-schedule'
        jobs.ci-worker-ocw-instance-refresh.plan.do[0].get(build-schedule): unknown resource 'build-schedule'
        jobs.ci-worker-generic-instance-refresh.plan.do[0].get(build-schedule): unknown resource 'build-schedule'
        jobs.ci-web-instance-refresh.plan.do[0].get(build-schedule): unknown resource 'build-schedule'
        jobs.qa-worker-infra-instance-refresh.plan.do[0].get(build-schedule): unknown resource 'build-schedule'
        jobs.qa-worker-ocw-instance-refresh.plan.do[0].get(build-schedule): unknown resource 'build-schedule'
        jobs.qa-worker-generic-instance-refresh.plan.do[0].get(build-schedule): unknown resource 'build-schedule'
        jobs.qa-web-instance-refresh.plan.do[0].get(build-schedule): unknown resource 'build-schedule'
        jobs.production-worker-infra-instance-refresh.plan.do[0].get(build-schedule): unknown resource 'build-schedule'
[...]

PR #2382, opened 2024-04-29
Add start and stop times to the scheduled build task

feoh:
#2164 

Right now our Concourse Python model's `schedule` task only supports regularly recurring "every" runs.

This change adds the ability to support discrete start and end times - e.g. run at 3AM and be sure you're done by 11AM etc.

### How can this be tested?
I used this change to rebuild the [Concourse instance refresh](https://cicd.odl.mit.edu/teams/infrastructure/pipelines/instance-refresh-concourse) pipeline.


src/ol_concourse/pipelines/infrastructure/concourse/instance_refresh.py
```diff
@@ -10,6 +11,10 @@
 environments = ["ci", "qa", "production"]
 node_classes = ["worker-infra", "worker-ocw", "worker-generic", "web"]
 
+build_schedule = schedule(
+    Identifier("build-schedule"), interval="24h", start="08:00", stop="11:00"
```
Ardiea, line 15: Presumably this is in UTC? So between 3AM and 6AM daily?

Ardiea (approved):

src/ol_concourse/pipelines/infrastructure/concourse/instance_refresh.py
feoh: Yup that was Carey's request in the issue. That window is timed around critical OCW work that can't be interrupted.

Thanks for checking!

PR #2383, opened 2024-04-29
Removing no longer needed migration plugin

shaidar:

Remove the Keycloak user migration plugin that we were testing with and potentially were gonna use for migrating users from Discussions. However, appears that we won't be going that route and thus the plugin is no longer necessary.


blarghmatey (approved):

collinpreston (commented):
Should we also remove `keycloak:user_migration_open_discussions_bearer_token` and wherever the associated secret is held (I think in Vault)?

I think we should also remove the token from within Open Discussions in QA.

collinpreston:
I deleted the user associated with this token in Discussions in RC.

collinpreston (approved):

PR #2414, opened 2024-05-09
Feat: Enable OpenEdX Redwood for mitx/staging in CI

feoh:
Sets mitx/staging up for redwood in CI.

### How can this be tested?
Watch the edx-platform-global pipeline.

### DIFF
--- a/src/bridge/settings/openedx/version_matrix.py
+++ b/src/bridge/settings/openedx/version_matrix.py
@@ -169,19 +169,19 @@ ReleaseMap: dict[
                 application="communications",
                 application_type="MFE",
                 release="redwood",
-                branding_overrides=pinned_branding_overrides,
+                branding_overrides=default_branding_overrides,
             ),
             OpenEdxApplicationVersion(
                 application="course-authoring",
                 application_type="MFE",
                 release="redwood",
-                branding_overrides=pinned_branding_overrides,
+                branding_overrides=default_branding_overrides,
             ),
             OpenEdxApplicationVersion(
                 application="discussions",
                 application_type="MFE",
                 release="redwood",
-                branding_overrides=pinned_branding_overrides,
+                branding_overrides=default_branding_overrides,
             ),
             OpenEdxApplicationVersion(
                 application="edx-platform",
--- a/src/bridge/settings/openedx/version_matrix.py
+++ b/src/bridge/settings/openedx/version_matrix.py
@@ -13,7 +13,7 @@ class OpenLearningOpenEdxDeployment(Enum):
     mitx = DeploymentEnvRelease(
         deployment_name="mitx",
         env_release_map=[
-            EnvRelease("CI", OpenEdxSupportedRelease["quince"]),
+            EnvRelease("CI", OpenEdxSupportedRelease["redwood"]),
             EnvRelease("QA", OpenEdxSupportedRelease["quince"]),
             EnvRelease("Production", OpenEdxSupportedRelease["quince"]),
         ],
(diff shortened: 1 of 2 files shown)


src/bridge/settings/openedx/version_matrix.py
```diff
@@ -80,6 +80,174 @@ def get_item(cls, key: OpenEdxDeploymentName) -> DeploymentEnvRelease:
     OpenEdxSupportedRelease,
     dict[OpenEdxDeploymentName, list[OpenEdxApplicationVersion]],
 ] = {
+    "redwood": {
+        "mitx": [
+            OpenEdxApplicationVersion(
+                application="communications",
+                application_type="MFE",
+                release="redwood",
+                branding_overrides=pinned_branding_overrides,
```
blarghmatey, line 89: I think we'll want to use the default_branding_overrides for MFEs in Redwood.

src/bridge/settings/openedx/version_matrix.py
```diff
@@ -80,6 +80,174 @@ def get_item(cls, key: OpenEdxDeploymentName) -> DeploymentEnvRelease:
...
+                application_type="IDA",
+                release="redwood",
+                branch_override="mitx/redwood",
+                origin_override="https://github.com/mitodl/edx-platform",
+            ),
+            OpenEdxApplicationVersion(
+                application="edxapp_theme",
+                application_type="IDA",
+                release="redwood",
+                branch_override="redwood",
```
blarghmatey, line 114: We'll want to either create this branch ourselves, or ask the Arbisoft folks to help with that before we merge this change.

src/bridge/settings/openedx/version_matrix.py
```diff
@@ -13,15 +13,15 @@ class OpenLearningOpenEdxDeployment(Enum):
     mitx = DeploymentEnvRelease(
         deployment_name="mitx",
         env_release_map=[
-            EnvRelease("CI", OpenEdxSupportedRelease["quince"]),
+            EnvRelease("CI", OpenEdxSupportedRelease["redwood"]),
```
blarghmatey, line 16: You'll also need to add a definition for `redwood` to the `OpenEdxSupportedReleases` Enum in the `types.py` neighboring file.

blarghmatey (changes requested):

src/bridge/settings/openedx/version_matrix.py
arslanashraf7: I have cut a `redwood` branch in mitx-theme https://github.com/mitodl/mitx-theme/tree/redwood. I've tested the theme with latest master of the edx-platform and the theme seems to work just fine which means that we can expect it to work with open-release/redwood.master as well.

In addition: 

1. We'll need to cut a `mitx/redwood` branch in https://github.com/mitodl/edx-platform. So I created https://github.com/mitodl/edx-platform/tree/mitx/redwood in our fork of the edx-platform. 
    - **NOTE**: This branch is based on https://github.com/openedx/edx-platform/tree/open-release/redwood.master and not the upstream `master`
2. We'll need to cherry-pick some commit(s) for our customization for MITx Residential. I didn't do that in https://github.com/openedx/edx-platform/tree/open-release/redwood.master because I wasn't sure if we usually cut our named branches i.e. `mitx/quince or mitx/redwood` from named release branches of upstream or the master.

@blarghmatey You might want to take a look at the above comment as well. Also, I see a cherry-picked commit on the master branch of our edx-platform fork. Is that intentional? My assumption was that we would have done that in a named branch and not master.

blarghmatey: Thanks for all of this @arslanashraf7! This is great.

To your question of the cherry-pick on master, that was to address a bug for our MITx Online deployment, which has since landed on upstream. I'll update our fork from upstream to remove that confusion.

For our named releases, we typically branch off of the named release branch and then apply cherry-picks. At this point we are down to only 1 custom commit, and that's related to our integration with Canvas. I'll bring that in to the branch that you made.

blarghmatey (approved):

PR #2434, opened 2024-05-17
Let MIT Open ingest OLL course data from S3

blarghmatey:
N/A

The course bundles have been loaded to S3 for ingest by MIT Open, this grants permissions on the Vault role definition and sets the values for the application to use for loading that data.


mbertrand (approved):
👍

PR #2441, opened 2024-05-22
MIT Open Application routing paths

Ardiea:
https://github.com/mitodl/mit-open/issues/629

Adding the rest of the paths to route to the application rather than to s3 based on feedback from Chris and Shankar. 

This should recreate the [nginx regex](https://github.com/mitodl/mit-open/blob/main/config/nginx.conf#L35C18-L35C210) from the MITOpen repo.


blarghmatey (approved):

blarghmatey:
That's a lot of paths...

PR #2444, opened 2024-05-23
refactor: Use DockerHub image as upstream for Superset

blarghmatey:
N/A

Superset is now on verion 4.x, and has started publishing Dcker images to DockerHub. This fixes our upstream target, as well as using the GitHub release tags for identifying which image tag to use for our builds. This should also help to resolve the errors that we were experiencing with invalid database migrations during deployments.


quazi-h (approved):
These changes to start pulling the official superset release instead of the upstream image makes sense to me. I don't understand how this resolves the database migrations issues because shouldn't the latest image they release that we were using be the same as the official GitHub releases? Or is the lack of using the GitHub_release resource the part that's causing the db migrations issue?

blarghmatey:
The problem was that they were no longer publishing images to the scarf.sh repository, and instead are using DockerHub, which we weren't using for our base image.

PR #2457, opened 2024-05-31
config: disable Discussions MFE as a default for discussions

arslanashraf7:
https://github.com/mitodl/hq/issues/4161#issuecomment-2141466986

This flag enables the new Discussions MFE for all the courses with or without legacy discussions. We don't want that. We want to load new Discussions MFE for only those courses where we explicitly want to.

I've already disabled it from the QA environment. More details on https://github.com/mitodl/hq/issues/4161#issuecomment-2141466986

### DIFF
--- a/src/ol_infrastructure/applications/edxapp/Pulumi.applications.edxapp.mitxonline.QA.yaml
+++ b/src/ol_infrastructure/applications/edxapp/Pulumi.applications.edxapp.mitxonline.QA.yaml
@@ -12,7 +12,6 @@ config:
   - ["course_experience.relative_dates", "--create", "--superusers", "--everyone",
     "--staff", "--authenticated"]
   - ["discussions.pages_and_resources_mfe", "--create", "--superusers", "--staff"]
-  - ["discussions.enable_discussions_mfe", "--create", "--everyone"]
   - ["grades.bulk_management", "--create", "--superusers", "--everyone", "--staff",
     "--authenticated"]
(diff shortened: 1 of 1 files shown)


src/ol_infrastructure/applications/edxapp/Pulumi.applications.edxapp.mitxonline.QA.yaml
```diff
@@ -12,7 +12,6 @@ config:
   - ["course_experience.relative_dates", "--create", "--superusers", "--everyone",
     "--staff", "--authenticated"]
   - ["discussions.pages_and_resources_mfe", "--create", "--superusers", "--staff"]
-  - ["discussions.enable_discussions_mfe", "--create", "--everyone"]
```
blarghmatey, line 15: We can delete this, and manually disable the flag in the admin, or we can change the `--create` to `--disable` if we want to ensure that the flag is not enabled.

blarghmatey (approved):

src/ol_infrastructure/applications/edxapp/Pulumi.applications.edxapp.mitxonline.QA.yaml
arslanashraf7: I've already disabled it through Django Admin https://courses-qa.mitxonline.mit.edu/admin/waffle/flag/. Also checked on Prod and the flag doesn't exist there https://courses.mitxonline.mit.edu/admin/waffle/flag/. So we should be good.

PR #2461, opened 2024-05-31
Explicitly reduce the TTL for instructor uploaded files

blarghmatey:
https://github.com/mitodl/hq/issues/2466

Files uploaded by instructors in edX do not get explicitly expired from the cache when they are updated. This explicitly limits the TTL for those files to 30 seconds to avoid problems with stale content getting served when a file is updated.

### How can this be tested?

This has been applied to the QA environment of MITx Online. Upload a file to a course, load that file, then upload a different file with the same name and verify that the new file is displayed after no more than 30 seconds.

### DIFF
--- a/src/ol_infrastructure/applications/edxapp/__main__.py
+++ b/src/ol_infrastructure/applications/edxapp/__main__.py
@@ -1536,6 +1536,18 @@ edxapp_fastly_service = fastly.ServiceVcl(
         ),
         fastly.ServiceVclSnippetArgs(
             content=textwrap.dedent(
+                """\
+                if (req.url.path ~ "^/asset-v1:") {
+                  set beresp.ttl = 30;
+                  return (deliver)
+                }
+                """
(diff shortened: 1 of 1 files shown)


pdpinch:
I think this is the best, reasonable solution, but I'm curious about a couple of things. 

1. Does Fastly update the file immediately when the TTL is passed, or does the update need to be triggered by a browser request?
2. When Fastly updates the file, does it make a If-Modified-Since request to the origin server, or is it just a regular GET.

Regardless of the answers to these questions, I think we should go ahead with this change.

pdpinch (approved):
Will we be able to test this on QA before releasing to production?

blarghmatey:
We will be able to test in QA. The ttl is set for the cached file as it exists in Fastly from the time that the file is requested, so the longest delay it will ever have is 30 seconds, and then it will request from origin again.

PR #2472, opened 2024-06-05
feat: Add configs for purging Fastly cache in GH Actions

blarghmatey:
Related to https://github.com/mitodl/mit-open/pull/1021

Adds configs for being able to purge the Fastly cache

### DIFF
--- a/src/ol_infrastructure/applications/mitopen/__main__.py
+++ b/src/ol_infrastructure/applications/mitopen/__main__.py
@@ -794,6 +794,20 @@ gh_workflow_embedlykey_env_secret = github.ActionsSecret(
     ),
     opts=ResourceOptions(provider=github_provider),
 )
+gh_workflow_fastly_api_key_env_secret = github.ActionsSecret(
+    f"ol_mitopen_gh_workflow_fastly_api_key_env_secret-{stack_info.env_suffix}",
+    repository=gh_repo.name,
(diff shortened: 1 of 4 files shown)


src/ol_infrastructure/applications/mitopen/__main__.py
```diff
@@ -794,6 +794,20 @@
     ),
     opts=ResourceOptions(provider=github_provider),
 )
+gh_workflow_fastly_api_key_env_secret = github.ActionsSecret(
+    f"ol_mitopen_gh_workflow_fastly_api_key_env_secret-{stack_info.env_suffix}",
+    repository=gh_repo.name,
+    secret_name="FASTLY_API_KEY",  # noqa: S106 # pragma: allowlist secret
```
Ardiea, line 800: These need to have `env_var_suffix` appended to the end like the secrets above otherwise both stacks are going to smash on the same secret name.

Ardiea (changes requested):

src/ol_infrastructure/applications/mitopen/__main__.py
blarghmatey: I generated the key to work for both RC and production, but I updated the stack id to include the suffix.

Ardiea (approved):

Issue #2454, opened 2024-05-29
CLI for managing secrets for software developer consumption


feoh:
### Description/Context

Right now the management of secrets for devs is entirely ad-hoc. This lends itself to unfortunate incidents where monstrously huge .env files contain all the secrets and linger on filesystems just waiting to be unintentionally posted somewhere public.

We should have a CLI that every developer can access as a standard part of their environment which would either write temporary environment files to be deleted on process exit, or even better just use the process environment manager so they're never written at all.

PR #2481, opened 2024-06-11
Reformat some Open key names

shaidar:

Related to work on https://github.com/mitodl/ol-infrastructure/issues/2454

Some of the key names have dashes instead of underscores and when trying to use the vault agent templating, that was throwing errors. It's a known [issue](https://support.hashicorp.com/hc/en-us/articles/15667022913299-Vault-Agent-throws-bad-character-U-002D) and the recommendation is to use underscores, otherwise we would need to do a few workarounds which would be confusing down the line, so better to fix it now.


src/bridge/secrets/mitopen/secrets.production.yaml
blarghmatey, line 1: Unfortunately these keys need to be edited in the context of SOPS so that it can re-write the mac for the file. If you edit directly then it invalidates the file and SOPS can't read it any more.

Ardiea (approved):

PR #2513, opened 2024-06-25
Added keycloakify starter theme

shaidar:

This add the [Keycloakify](https://docs.keycloakify.dev) [starter template](https://github.com/keycloakify/keycloakify-starter)

### DIFF
--- a/src/ol_concourse/pipelines/infrastructure/keycloak/docker_packer_pulumi_pipeline.py
+++ b/src/ol_concourse/pipelines/infrastructure/keycloak/docker_packer_pulumi_pipeline.py
@@ -125,6 +125,14 @@ def build_keycloak_infrastructure_pipeline() -> PipelineFragment:
         repository="keycloak-scim",
     )
 
+    # Repo: https://github.com/keycloakify/keycloakify-starter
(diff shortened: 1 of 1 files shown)


src/ol_concourse/pipelines/infrastructure/keycloak/docker_packer_pulumi_pipeline.py
```diff
@@ -125,6 +125,14 @@ def build_keycloak_infrastructure_pipeline() -> PipelineFragment:
         repository="keycloak-scim",
     )
 
+    # Repo: https://github.com/keycloakify/keycloakify-starter
+    # Use: Keycloak Starter Theme
+    keycloakify_spi = github_release(
+        name=Identifier("keycloakify-spi"),
+        owner="keycloakify",
```
blarghmatey, line 132: I think that the intended method of using the starter theme is to fork it so that it can be used as the basis for the customized theme.

blarghmatey (commented):
Once we have our own fork of the theme starter then this looks good. Thanks!

blarghmatey (approved):

PR #2530, opened 2024-06-28
Add Open redirect URL

shaidar:

Add a URL to the Keycloak Open config so that when a user logs out, it redirect back to the homepage


Ardiea (approved):

Ardiea:
pre-commit.ci run

Issue #2556, opened 2024-07-17
Review vault role / user creation 


Ardiea:
### Description/Context

```
4:04 PM sar  Yeah think as long as we don't have the grant definitions, then we should be good as those are the statements that are causing the error
4:06 PM tmacey Looking at it more, it seems like the static role definition is for when you have already defined the role and just want to rotate credentials. There's no parameter for defining the creation statements to allow for specifying the grants.
4:21 PM sar Can it be broken down into multiple roles?
4:22 PM tmacey Yeah, we would need to have one role for admins, one for app users, and one for readonly
4:23 PM sar We kinda have that though. I was thinking more in terms of the app role or whatever role that's throwing the error. Can that be broken down further
4:25 PM tmacey So, the issue is that the app user explicitly grants permissions every time it gets created. We need to move those grants to the role definition, and then the user creation permissions are just "grant <role> to <user>;"
4:28 PM sar But if the problem is the overloaded grants, then wouldn't the error be thrown when trying to create that role with all the grant statements
4:32 PM tmacey The problem is that the grants are getting run too many times (I think)
4:32 PM Also, I think the revoke statements aren't all working right, which is leaving lots of excess user objects laying around
4:34 PM Basically, we need to do a thorough and holistic review of our Postgres permissioning. At this point it's mostly a cobbled together set of "I got it to work?" fixes that have accrued over time.
4:34 PM sar I'm not sure about that, cause i think when i ran into this, I tried running a script that would run the grants sequentially and that seemed to have worked, but u can't do that in the vault creation statement
4:34 PM tmacey I think you can.
4:34 PM If you look at the UI, it accepts a list of statements
4:34 PM We just have it all as a single blob because of historical reasons.
4:35 PM I think we can turn our "one big string" into an array
4:35 PM sar Oh great, then maybe that would be the way to solve it. But yeah I agree, we need to look over the entire approach
```

https://github.com/mitodl/ol-infrastructure/blob/main/src/ol_infrastructure/lib/vault.py#L19-L98

PR #2582, opened 2024-08-01
Vault DB user provisioning revisit - #2556

Ardiea:
#2556

1. Refactors the OLVault*DatabaseConfig component resource to take in a list of templates rather than just one template. 
2. Updates all the existing long-string database role statements into multiple discrete statements that will exeucte sequentially. This should help with issue in postgres regarding row size and too large privs.
3. Updated the user provisioning in postgres to use a PLPSQL function to look for the existing role and create it as needed and do nothing if it isn't, rather than depending on the `approle` hack. 
4. Removed the `approle` hack. 
5. Refactored the `readonly` users to use a role the same as the `app` user. It is basically exactly the same except it only gets `SELECT` rather than `ALL`. 
6. Refactored the custom database roles + schema creation for MITOpen to follow the same pattern as items 3+5. The only reason MITOpen is special is because of the `external` schema and the `reverse_etl` role that is not allowed to interact with the `public` schema. 
7. Made the strings that show up in the vault configuration look nice.
8. Fixed revocation for postgres so it actually works now and the user goes bye-bye when the lease expires / is revoked. 

### Additional Context
The SQL statements should be given close attention.

### DIFF
--- a/src/ol_infrastructure/applications/mitopen/__main__.py
+++ b/src/ol_infrastructure/applications/mitopen/__main__.py
@@ -337,69 +337,243 @@ mitopen_db = OLAmazonDB(mitopen_db_config)
 mitopen_role_statements = postgres_role_statements.copy()
 mitopen_role_statements.pop("app")
 mitopen_role_statements["app"] = {
-    "create": Template(
-        """CREATE USER "{{name}}" WITH PASSWORD '{{password}}'
-            VALID UNTIL '{{expiration}}' IN ROLE "mitopen" INHERIT;
-          GRANT CREATE ON SCHEMA public TO mitopen WITH GRANT OPTION;
-          GRANT ALL PRIVILEGES ON ALL TABLES IN SCHEMA public TO "mitopen"
+    "create": [
+        # Check if the mitopen role exists and create it if not
+        Template(
+            """
+            DO
+            $$do$$
+            BEGIN
+               IF EXISTS (
+                  SELECT FROM pg_catalog.pg_roles
+                  WHERE  rolname = 'mitopen' THEN
+               ELSE
+                  BEGIN   -- nested block
+                     CREATE ROLE mitopen;
+                  EXCEPTION
+                     WHEN duplicate_object THEN
+                        RAISE NOTICE 'Role "mitopen" was just created by a concurrent transaction. Skipping.';
+                  END;
+               END IF;
+            END
+            $$do$$;
+            """
+        ),
+        # Create the external schema if it doesn't exist already
+        Template("""CREATE SCHEMA IF NOT EXISTS external;"""),
+        # Do grants on to the mitopen in both schemas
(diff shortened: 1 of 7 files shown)


src/ol_infrastructure/applications/mitopen/__main__.py
```diff
@@ -337,67 +337,207 @@
...
+    "create": [
+        # Check if the reverse_etl role exists and create it if not
+        Template(
+            """
+            DO
+            $$do$$
+            BEGIN
+               IF EXISTS (
+                  SELECT FROM pg_catalog.pg_roles
+                  WHERE  rolname = 'reverse_etl' THEN
```
blarghmatey, line 465: Is this supposed to have a dangling `THEN`?

src/ol_infrastructure/lib/vault.py
```diff
@@ -17,134 +18,292 @@
...
+                WITH GRANT OPTION;
+                """
+            ),
+        ],
+        "revoke": [
+            # Remove the user from the pre-existing rds_superuser role
+            Template("""REVOKE "rds_superuser" FROM "{{name}}";"""),
+            # Ensure that the application role can do creations
+            Template(
+                """GRANT CREATE ON SCHEMA public TO ${app_name} WITH GRANT OPTION;"""
```
blarghmatey, line 48: Is this copy 🍝 from the app role? I wouldn't imagine that we want to assign admin objects to the app role.

src/ol_infrastructure/lib/vault.py
```diff
@@ -17,134 +18,292 @@
...
+                   END IF;
+                END
+                $$do$$;
+                """
+            ),
+            # Set/refresh the default privileges for the new role
+            Template(
+                """GRANT CREATE ON SCHEMA public TO ${app_name} WITH GRANT OPTION;"""
+            ),
+            Template("""
```
blarghmatey, line 97: Do we still want to apply all of these grants during the user creation step? I'm just curious about the benefit of not having the approle definition for assigning/updating the role permissions and then just having the app/readonly/etc. just get assigned the appropriate role?

Ardiea: So I think the primary benifit here is that the role gets 'fixed' every time a new user is created and assigned too it. If there were any out-of-band changes that took place to the role, at least this minimum set of privs will be restored. It won't address any _additional_ permissions that may have been given but it will at least restore the minimum set.

We can move this into the fancy conditional structure when the role is now 'created' or its existence is checked but it will have the same effect. If we put these steps up there, each line will have to exist twice in both steps of the conditional. 

I believe moving these grants / default privs into their own statements like this will address error message we get about block size on new account provisions.

src/ol_infrastructure/lib/vault.py
Ardiea: I'm not entirely sure where this came from. I went through the admin revoke and tried to make it make sense so it could probably use another close review with the next commit.

blarghmatey (approved):

Issue #2589, opened 2024-08-05
Deploy OpenMetaData On Our Shiny New Data Platfork EKS Cluster


feoh:
### Description/Context
We want to add [OpenMetaData](https://open-metadata.org/) to our data platform so we can have a single source of truth index for all our data.

We also want to standardize on Kubernetes / EKS for our future deployments.

### Plan/Design
- Spin up a throw-away K3s cluster for testing this
- Write Pulumi code to build an OMD specific Opensearch cluster
- Write Pulumi code to create an OMD specific RDS Postgres instance
- Deploy OpenMetaData to the K3s cluster using its Helm chart.

feoh:
Override the storage class to use EFS:
```
helm upgrade openmetadata-dependencies open-metadata/openmetadata-dependencies -n omd  --set "airflow.dags.persistence.storageClass=efs-sc" --set "airflow.logs.persistence.storageClass=efs-sc"
```

feoh:
Mike has broken the back of this task. See https://github.com/mitodl/hq/discussions/5463

PR #2593, opened 2024-08-06
Pulumi RDS, Consul, k8s and supporting infra for OpenMetaData CI

feoh:

Part of #2589 

Create an opensearch cluster for the OpenMetadata application on EKS.


src/ol_infrastructure/applications/open_metadata/__main__.py
```diff
@@ -0,0 +1,180 @@
...
+                "Principal": {"Service": "ec2.amazonaws.com"},
+            },
+        }
+    ),
+    name=f"open-metadata-instance-role-{stack_info.env_suffix}",
+    path="/ol-infrastructure/open-metadata-role/",
+    tags=aws_config.tags,
+)
+
+open_metadata_server_security_group = ec2.SecurityGroup(
```
Ardiea, line 63: We don't need this because there will not be an open metadata server.

src/ol_infrastructure/applications/open_metadata/__main__.py
```diff
@@ -0,0 +1,180 @@
...
+            to_port=DEFAULT_POSTGRES_PORT,
+            description="Access to postgres from open metadata servers.",
+        ),
+        ec2.SecurityGroupIngressArgs(
+            security_groups=[],
+            protocol="tcp",
+            from_port=DEFAULT_POSTGRES_PORT,
+            to_port=DEFAULT_POSTGRES_PORT,
+            cidr_blocks=["73.218.126.92/32", "98.110.175.231/32"],
+            description="Allow mike and matt to connect to postgres.",
```
Ardiea, line 103: You must have taken this from the semantic chart? You'll want to change the CIDR block(s) to be the EKS pod address space, which should be an output of the network stack  like `infrastructure.aws.network.CI` or something.

src/ol_infrastructure/applications/open_metadata/__main__.py
```diff
@@ -0,0 +1,180 @@
...
+    ],
+    opts=consul_provider,
+)
+
+# Create a vault policy to allow dagster to get to the secrets it needs
+open_metadata_server_vault_policy = vault.Policy(
+    "open-metadata-server-vault-policy",
+    name="open-metadata-server",
+    policy=Path(__file__)
+    .parent.joinpath("open_metadata_server_policy.hcl")
```
Ardiea, line 176: This file doesn't exist.

But also we don't need _this_ specifically but we will need something like it to give the app / vault-secrets-operator the permissions it needs to generate db creds.

feoh: I just forgot to add the file which I'd written locally. I put it in, even if we need something like it we can alter the contents once we figure them out :)

src/ol_infrastructure/applications/open_metadata/__main__.py
```diff
@@ -0,0 +1,209 @@
...
+)
+
+open_metadata_db_config = OLPostgresDBConfig(
+    instance_name=f"open-metadata-db-{stack_info.env_suffix}",
+    password=open_metadata_config.get("db_password"),
+    subnet_group_name=apps_vpc["rds_subnet"],
+    security_groups=[open_metadata_database_security_group],
+    storage=open_metadata_config.get("db_capacity")
+    or str(AWS_RDS_DEFAULT_DATABASE_CAPACITY),
+    engine_major_version="15",
```
blarghmatey, line 94: Why not 16? That'ş the current latest release, with 17 on the way.

src/ol_infrastructure/applications/open_metadata/__main__.py
```diff
@@ -0,0 +1,209 @@
...
+    kubeconfig=kubeconfig,
+)
+
+open_metadata_application = kubernetes.helm.v3.Release(
+    f"{cluster_name}-open-metadata-application-helm-release",
+    kubernetes.helm.v3.ReleaseArgs(
+        name="open-metadata-application",
+        chart="open-metadata-application",
+        version=VERSIONS["OPEN_METADATA"],
+        namespace="applications",
```
blarghmatey, line 171: Shouldn't this be `open-metadata`?

src/ol_infrastructure/infrastructure/aws/eks/__main__.py
```diff
@@ -53,6 +53,7 @@
     "EBS_CSI_DRIVER": "v1.33.0-eksbuild.1",
     "EFS_CSI_DRIVER": "v2.0.7-eksbuild.1",
     "VAULT_SECRETS_OPERATOR": "0.8.1",
+    "OPEN_METADATA": "1.5.5",
```
blarghmatey, line 56: This doesn't seem like it should go here?

src/ol_infrastructure/applications/open_metadata/__main__.py
```diff
@@ -0,0 +1,209 @@
...
+                    "database": {
+                        "host": open_metadata_db.db_instance.address,
+                        "port": open_metadata_db_config.port,
+                        "databaseName": open_metadata_db_config.db_name,
+                        "auth": {
+                            "username": open_metadata_db_config.username,
+                            "password": {
+                                "secretRef": open_metadata_db_config.password_secret_name,
+                                "secretKey": open_metadata_db_config.password_secret_key,
+                            },
```
blarghmatey, line 201: These values should either be coming from Vault, or we might take a look at using IAM auth to RDS since OpenMetadata supports that out of the box.

Ardiea: I will figure out the vault dynamic secret resources and add it to this PR. It will be a good example. At the moment the only thing I've used on the EKS stacks is static secrets.

src/ol_infrastructure/infrastructure/aws/eks/__main__.py
Ardiea: This should probably go [here](https://github.com/mitodl/ol-infrastructure/blob/main/src/bridge/lib/versions.py) (maybe this entire dict should be moved into `versions.py`?) and eventually we should make it/them envrionment var loads so that the stacks can be updated to latest versions automagically by the pipeline (eventually).

For now, probably just move this to `versions.py` so that we aren't creating an unusual relationship between this application stack and the infrastructure stack.

src/ol_infrastructure/applications/open_metadata/__main__.py
feoh: Mike implemented a check_cluster_namespace method to help with this.

src/ol_infrastructure/infrastructure/aws/eks/__main__.py
feoh: Looks like mike moved this to VERSIONS. Thanks Mike!

src/ol_infrastructure/infrastructure/aws/eks/__main__.py
```diff
@@ -62,6 +62,8 @@
     "EXTERNAL_DNS_CHART": "1.15.0",
     "TRAEFIK_CHART": "v31.0.0",
     "VAULT_SECRETS_OPERATOR_CHART": "0.8.1",
+    "VAULT_SECRETS_OPERATOR": "0.8.1",
+    "OPEN_METADATA": "1.5.5",
```
blarghmatey, line 66: Should this be coming from `versions.py`?

Ardiea: mmm, slipped in when I was rebasing, thanks!

Ardiea (approved):

blarghmatey (approved):

PR #2598, opened 2024-08-09
Mailgun MITLearn setup

shaidar:

Configure MITLearn Mailgun domain for future use with app.


src/ol_infrastructure/applications/mailgun/__main__.py
blarghmatey, line 1: This seems like it would be worth either turning into a component resource, or a loop to be able to manage all of our Mailgun domains. I think the determination of which path to take is on how much customization we need across the different domains?

shaidar: Yeah, I just kept it basic for now to get something in place and down the line when we decide to move all the existing stuff under pulumi management, we can do some refactoring.

blarghmatey (approved):

PR #2612, opened 2024-08-16
adding email notification backend

shanbady:
Sets the "MITOL_NOTIFICATION_EMAIL_BACKEND" env variable so we can use mailgun

### How can this be tested?
when i send emails directly using
```
from profiles.utils import send_email
send_email(['ambady@mit.edu'], 'test email', 'test', False)
```
it seems to work regardless of this setting. I noticed however, subscription emails would not send. to do that I have a script which manually sets the created date for new learning resources to today and sends subscription updates (which i have not done on prod). You should be able to test this by unsetting this variable (on rc heroku) and then running the following and see that subscription emails dont get sent unless this setting is in place

```
from learning_resources_search.tasks import send_subscription_emails
send_subscription_emails("channel_subscription_type")
```

### DIFF
--- a/src/ol_infrastructure/applications/mitopen/Pulumi.applications.mitlearn.QA.yaml
+++ b/src/ol_infrastructure/applications/mitopen/Pulumi.applications.mitlearn.QA.yaml
@@ -41,7 +41,6 @@ config:
     POSTHOG_API_HOST: "https://app.posthog.com"
     POSTHOG_TIMEOUT_MS: 1000
     POSTHOG_PROJECT_ID: "63497"
-    MITOL_NOTIFICATION_EMAIL_BACKEND: "anymail.backends.mailgun.EmailBackend"
   heroku_app:interpolation_vars:
     auth_allowed_redirect_hosts:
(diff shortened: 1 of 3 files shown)


src/ol_infrastructure/applications/mitopen/Pulumi.applications.mitlearn.QA.yaml
```diff
@@ -41,6 +41,7 @@ config:
     POSTHOG_API_HOST: "https://app.posthog.com"
     POSTHOG_TIMEOUT_MS: 1000
     POSTHOG_PROJECT_ID: "63497"
+    MITOL_NOTIFICATION_EMAIL_BACKEND: "anymail.backends.mailgun.EmailBackend"
```
blarghmatey, line 44: This can just be added in `__main__.py` since it's the same regardless of environment.

shanbady: moved it to main

Ardiea (approved):

PR #2618, opened 2024-08-19
renaming mitopen->mitlearn rds instance

Ardiea:
renaming mitopen->mitlearn rds instance

@blarghmatey Can you review the change set for this? As far as I can tell it is okay. It wants to recreate almost everything that RDS related EXCEPT the rds instance itself. I believe this okay?


blarghmatey:
I just ran the preview to determine what changes it wants to execute and they all look safe and sane. 👍

blarghmatey (approved):

PR #2628, opened 2024-08-21
feat: remove certificate link from resedential course authoring mfe

Anas12091101:
Note: This PR should be merged after the merge of https://github.com/openedx/frontend-app-course-authoring/pull/1223

https://github.com/mitodl/hq/issues/4658

This PR removes the certificate link from course authoring header settiings tab.

Before
Screenshot 2024-08-21 at 4 00 37 PM

After
Screenshot 2024-08-21 at 3 48 30 PM

### How can this be tested?
Note: Make sure you enable the contentstore.new_studio_mfe.use_new_course_outline_page waffle flag so that the authoring MFE will be used for displaying the course outline.

- In Studio, visit any course outline and click the Settings tab. You should see a certificate link there.
- In the .env.development file, set ENABLE_CERTIFICATE_PAGE=false.
- Restart the MFE container. Now, repeat the first step. There should be no certificate link.

### DIFF
--- a/src/ol_concourse/pipelines/open_edx/mfe/pipeline.py
+++ b/src/ol_concourse/pipelines/open_edx/mfe/pipeline.py
@@ -40,6 +40,7 @@ class OpenEdxVars(BaseModel):
     accessibility_url: Optional[str] = None
     account_settings_url: Optional[str] = None
     contact_url: Optional[str] = None
+    certificate: Optional[str] = None
     deployment_name: OpenEdxDeploymentName
     display_feedback_widget: Optional[str] = None
     environment: str
@@ -82,6 +83,7 @@ def mfe_params(
         "CONTACT_URL": open_edx.contact_url,
         "CSRF_TOKEN_API_PATH": "/csrf/api/v1/token",
         "DISPLAY_FEEDBACK_WIDGET": open_edx.display_feedback_widget,
+        "ENABLE_CERTIFICATE_PAGE":open_edx.certificate,
         "DISCUSSIONS_MFE_BASE_URL": f"https://{open_edx.lms_domain}/{discussion_mfe_path}",
         "FAVICON_URL": open_edx.favicon_url,
         "HONOR_CODE_URL": open_edx.honor_code_url,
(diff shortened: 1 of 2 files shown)


src/ol_concourse/pipelines/open_edx/mfe/pipeline.py
```diff
@@ -40,6 +40,7 @@ class OpenEdxVars(BaseModel):
     accessibility_url: Optional[str] = None
     account_settings_url: Optional[str] = None
     contact_url: Optional[str] = None
+    certificate: Optional[str] = None
```
blarghmatey, line 43: Should this be a boolean that we coalesce to a string in the build, or at least a `Literal['true', 'false']`? Setting it as a `str` makes it hard to know what the valid values are.

blarghmatey: I would also suggest that the variable be renamed to something like `enable_certificate_page` so that it's clearer when only seeing the variable what its purpose is.

src/ol_concourse/pipelines/open_edx/mfe/pipeline.py
```diff
@@ -82,6 +83,7 @@ def mfe_params(
         "CONTACT_URL": open_edx.contact_url,
         "CSRF_TOKEN_API_PATH": "/csrf/api/v1/token",
         "DISPLAY_FEEDBACK_WIDGET": open_edx.display_feedback_widget,
+        "ENABLE_CERTIFICATE_PAGE": open_edx.enable_certificate_page,
```
blarghmatey, line 86: This will result in `"False"` or `"True"` being passed as the parameter. As long as that is acceptable to the MFE build that will work, but if the intent is to have lowercased values then this would need to be stringified and lowercased.
```suggestion
        "ENABLE_CERTIFICATE_PAGE": str(open_edx.enable_certificate_page).lower(),
```

Anas12091101: I am not sure, it's a boolean value for `ENABLE_CERTIFICATE_PAGE` which has been set in mfe [here](https://github.com/openedx/frontend-app-course-authoring/pull/1223/files#diff-e9cbb0224c4a3d23a6019ba557e0cd568c1ad5e1582ff1e335fb7d99b7a1055dR37)

> This will result in "False" or "True" being passed as the parameter. 
Why would it be converted to string? I have set it to boolean type..

blarghmatey (approved):

PR #2658, opened 2024-09-09
Enable storage autoscaling up to 1TB by default for RDS

blarghmatey:

N/A

The current default configuration is that storage autoscaling of our RDS instances is disabled. This will lead to failures if the storage capacity is exhausted. This changes that to a default of up to 1TB autoscaled storage capacity in the event that the starting allocation starts to fill up.

### How can this be tested?

Run a `pulumi up` on a stack that uses an RDS instance and verify that the `maxAllocatedCapacity` will be modified if it is not otherwise set explicitly.

### Additional Context

This can lead to increased AWS costs, but it raises the threshold under which we are likely to experience errors due to storage capacity being exhausted. Currently, the largest database that we have in operation is ~400GB so 1TB is a substantial ceiling that we are unlikely to hit.


shaidar:
I think we should add some kinda alerting/notification to this when it's triggered

PR #2664, opened 2024-09-13
Md/k8s cleanup

Ardiea:
https://github.com/mitodl/hq/discussions/5463
https://github.com/mitodl/ol-infrastructure/issues/2589

Creates an operations-ci EKS cluster. Adds storage provisioner addons, the vault secrets operator, and the aws load balancer controller into that cluster. 

### How can this be tested?
Refer to [KUBE_CONFIG_SETUP.md](https://github.com/mitodl/ol-infrastructure/compare/main...md/k8s_cleanup?expand=1#diff-e8e590e9bb8000713936872eb060587f811306330064b24d9aacdb790419ca51) for how to connect to the cluster with kubectl + helm. 

stack name is infrastructure.aws.eks.operations.CI 

Also added a very simplistic IAM stack that creates a few user groups for us. Ended up not being able to really use this IAM framework beyond the global var `EKS_ADMIN_USERNAMES` but it still may come in handing for something some day or can be expanded upon for other ideas we have about giving ppl access to servers / etc.


src/ol_infrastructure/infrastructure/aws/eks/__main__.py
```diff
@@ -104,38 +152,543 @@
...
+        addon_version=VERSIONS["EFS_CSI_DRIVER"],
+        service_account_role_arn=efs_csi_driver_role.arn,
+        opts=ResourceOptions(
+            parent=cluster,
+            # Addons won't install properly if there are not nodes to schedule them on
+            depends_on=[cluster, node_groups[0]],
+        ),
+    )
+
+# Configure and install any extra addons
```
feoh, line 494: What does this mean? Would extra eks addons be addons configured elsewhere, like static configuration files or such? Just wondering about the use case.

feoh (approved):

src/ol_infrastructure/infrastructure/aws/eks/__main__.py
Ardiea: I think I originally imagined allowing us to define any "extra" addons beyond the storage providers as config yaml and then just creating the resources on fly using this loop. Turns out that we probably don't need to do that and there are not really any more interesting addons at the moment beyond the two we're doing explicitly for storage. So I will get rid of this.

src/ol_infrastructure/infrastructure/aws/eks/__main__.py
```diff
@@ -104,38 +152,543 @@
...
     subnet_ids=pod_subnet_ids,
+    use_default_vpc_cni=False,
+    # Ref: https://docs.aws.amazon.com/eks/latest/userguide/security-groups-pods-deployment.html
+    # Ref: https://docs.aws.amazon.com/eks/latest/userguide/sg-pods-example-deployment.html
+    vpc_cni_options=eks.cluster.VpcCniOptionsArgs(
+        enable_pod_eni=eks_config.get_bool("pod_security_groups"),
+        disable_tcp_early_demux=eks_config.get_bool("pod_security_groups"),
+        log_level="INFO",
+        #        warm_eni_target=2,
+        #        warm_ip_target=2,
```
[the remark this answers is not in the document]
Ardiea: Yes, I've been revisiting this today actually. The fancy networking with security groups and pods that I can't really get working.

src/ol_infrastructure/infrastructure/aws/eks/files/omd_values.yaml
blarghmatey, line 1: Should this be moved into an `application` project that handles creation of the namespace and applying of the helm chart? Along with any other resource management such as the RDS, OpenSearch, etc.

blarghmatey (approved):

src/ol_infrastructure/infrastructure/aws/eks/files/omd_values.yaml
Ardiea: We talked about this in real life. These are just notes / things I felt were important to keep during my experimentation. Ideally we won't be managing too much direct yaml for our apps, I think. 👍

PR #2669, opened 2024-09-16
Adding a missing export to the DNS stack

Ardiea:
Adding a normalized export for the odl zone in the dns stack.


blarghmatey:
Unfortunately, that won't actually work directly as is. For whatever reason, once you have "consumed" a given `Output` with an export, any subsequent `export` of that `Output` yields a null. For this to work, you'll need to refactor any usage of the `odl_zone_id` export with the new export that you created.

PR #2676, opened 2024-09-23
feat: Turn on max instance lifetime for our ec2 instances

### LINKED ISSUE #2652: Set an automatic lifetime limit for instances on our ASGs
### Description/Context
Right now we keep having problems with stale instances. We have a number of out of band pipelines to do refreshes, but that's not ideal.

### Plan/Design
Use https://www.pulumi.com/registry/packages/aws/api-docs/autoscaling/group/#max_instance_lifetime_python to modify our auto scaling groups to automatically refresh stale instances.

feoh:
This is set to 30 days as expressed in seconds.

Fixes #2652

This change sets the [max instance lifetime attribute](https://docs.aws.amazon.com/autoscaling/ec2/userguide/asg-max-instance-lifetime.html) on all of our AWS auto scaling groups to 30 days. It is currently unset.

This would mean that instances would refresh automatically every 30 days, likely curtailing many of the stale creds issues we now have.
 
### How can this be tested?

Run `pulumi up` on an application in CI. Note that the resource changes as expected. Monitor application health to ensure this isn't a destabilizing change.

### DIFF
--- a/src/ol_infrastructure/components/aws/auto_scale_group.py
+++ b/src/ol_infrastructure/components/aws/auto_scale_group.py
@@ -83,6 +83,7 @@ class OLTargetGroupConfig(AWSBase):
     health_check_protocol: str = "HTTPS"
     health_check_timeout: PositiveInt = PositiveInt(5)
     health_check_unhealthy_threshold: PositiveInt = PositiveInt(3)
+    max_instance_lifetime: Optional[PositiveInt] = 2592000  # 30 days
     model_config = ConfigDict(arbitrary_types_allowed=True)
 
     @field_validator("stickiness")


src/ol_infrastructure/components/aws/auto_scale_group.py
```diff
@@ -83,6 +83,7 @@ class OLTargetGroupConfig(AWSBase):
     health_check_protocol: str = "HTTPS"
     health_check_timeout: PositiveInt = PositiveInt(5)
     health_check_unhealthy_threshold: PositiveInt = PositiveInt(3)
+    max_instance_lifetime: Optional[PositiveInt] = 2592000  # 30 days
```
blarghmatey, line 86: This should include a unit in the variable name to avoid confusion. Also, this attribute will need to be passed to the actual ASG resource in the body of the component resource to have an effect.

blarghmatey (approved):

PR #2683, opened 2024-09-25
Adds iam tag to vault users created in AWS

shaidar:

Noticed that Vault created users in AWS IAM are not tagged. This should add tags to newly created accounts by vault based on [vault documentation](https://developer.hashicorp.com/vault/api-docs/secret/aws#parameters-3)

### DIFF
--- a/src/ol_infrastructure/components/services/vault.py
+++ b/src/ol_infrastructure/components/services/vault.py
@@ -235,7 +235,7 @@ class OLVaultAWSSecretsEngineConfig(BaseModel):
     vault_backend_path: str
     policy_documents: dict[str, str]
     credential_type: str = "iam_user"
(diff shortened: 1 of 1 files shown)


Ardiea (approved):

src/ol_infrastructure/components/services/vault.py
```diff
@@ -235,6 +235,7 @@ class OLVaultAWSSecretsEngineConfig(BaseModel):
     vault_backend_path: str
     policy_documents: dict[str, str]
     credential_type: str = "iam_user"
+    iam_tags: list[str] = ["OU=operations"]
```
blarghmatey, line 238: Probably also worth adding something like vault_managed=true

blarghmatey (approved):

PR #2685, opened 2024-09-26
Added definitions for olapps realm user profile

rhysyngsun:

Part of https://github.com/mitodl/hq/issues/5559

This adds the `fullName` attribute to the user profile. It also defines the other profile attributes as I'm assuming if I omit them here they'll be removed and we'll lose data.


shaidar (approved):

rhysyngsun:
Holding on merging this until the other pieces are ready to merge too.

PR #2686, opened 2024-09-26
set CERTIFICATES_HTML_VIEW to false in mitxonline

pdpinch:
Setting this to false disables access to the certificate settings in Studio, preventing course authors from accidentally turning on open edX certificates. Our certificates from mitxonline.mit.edu instead.

- recently came up in https://github.com/mitodl/hq/issues/5618 
- some unfinished work from https://github.com/mitodl/hq/issues/1630

Turns off the certificates UI in studio and hides certificate links from the progress page.


pdpinch:
@arslanashraf7 I think you've been bugging me about reverting this setting for some time. Let me know if you think I've done what's needed in https://github.com/mitodl/hq/issues/5618#issuecomment-2377878962

feoh (approved):

pdpinch:
I just confirmed that certificate settings are still available on, for example, https://studio-qa.mitxonline.mit.edu/course/course-v1:MITxT+0.69.2S+3T2023

When I merge this, I expect that will go away.

Similarly, the settings should be unavailable in production when this is deployed, probably on 10/2.

PR #2692, opened 2024-09-30
Md/k8s improvements 2

Ardiea:
- Recreate the data-ci EKS cluster
- Reorganizes a bunch of the `infrastructure.aws.eks` python code to be more readable
  - Moves several reasources out to a `substructure.aws.eks` stack due to dependency issues between pulumi resources. 
  - Added a lot of comments explaining what is happening where and why because it isn't entirely intuitive. 
- Fully implements the k8s gateway api for managing ingress into applications running in the cluster using traefik.
  - Disables the legacy ingress resources.
  - Removes the AWS load-balancer controller entirely 
- Implements cert-manager for automatically requesting and installing certificates from letsencrypt production + staging environments. 
  - This integrates with the k8s gateway api via an annotation on your gateway resources.  
- Implements external-dns for automatic management of DNS records in route 53 
  - This integrates with the k8s gateway api automatically w/o any annotations.
- Adds automatic installation of a *.odl.mit.edu wildcard certificate from vault into the operations namespace that can be used by any application that needs it.


src/ol_infrastructure/substructure/aws/eks/__main__.py
```diff
@@ -0,0 +1,291 @@
...
+        ),
+        opts=ResourceOptions(
+            provider=k8s_provider,
+            parent=vault_traefik_service_account,
+            delete_before_replace=True,
+        ),
+    )
+)
+
+# Create a VaultAuth resource and a VaultStaticSecret resource
```
feoh, line 127: Does this mean that we'll need a VaultStaticSecret per namespace per secret?

feoh (approved):

src/ol_infrastructure/substructure/aws/eks/__main__.py
Ardiea: Per Namespace / application / logical use case :
1 VaultConnection Resource (the helm chart makes the one we're using here)
1 ServiceAccount Resource
1 ClusterRoleBinding Resource
1 VaultAuth Resource

Per Secret:
X VaultDynamicSecret resources and/or Y VaultStaticSecret resources based on requirements, each of which maps to a k8s Secret Resource.

blarghmatey (approved):

PR #2713, opened 2024-10-08
Add email optin attribute and protocol mappers to ol apps keycloak config

rhysyngsun:

Closes https://github.com/mitodl/hq/issues/5559#issuecomment-2392330230

This adds the email opt in field to the user profile attributes and also adds protocol mappers for this and the last name field so that those can come back on the `openid-connect/userinfo` API response.

### How can this be tested?

I'm not aware of how to test this locally, I think we need to test it in a deployed test environment first.

### DIFF
--- a/src/ol_infrastructure/substructure/keycloak/__main__.py
+++ b/src/ol_infrastructure/substructure/keycloak/__main__.py
@@ -356,6 +356,14 @@ ol_apps_user_profile = keycloak.RealmUserProfile(
                 views=["admin", "user"], edits=["admin", "user"]
             ),
         ),
+        keycloak.RealmUserProfileAttributeArgs(
+            name="emailOptIn",
+            display_name="${emailOptIn}",
+            required_for_roles=[],
+            permissions=keycloak.RealmUserProfileAttributePermissionsArgs(
+                views=["admin", "user"], edits=["admin", "user"]
+            ),
+        ),
     ],
     groups=[
         keycloak.RealmUserProfileGroupArgs(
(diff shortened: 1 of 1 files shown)


shaidar:
Removed `multivalued` as it's not a valid key according to the [docs](https://www.pulumi.com/registry/packages/keycloak/api-docs/realmuserprofile/) and when I tried running pulumi, it threw an error stating that. Changed `config` as it also threw an error when trying to run pulumi to verify things:
```
Argument of type "dict[str, list[str]]" cannot be assigned to parameter "config" of type "Input[Mapping[str, Input[str]]] | None" in function "__init__"
  Type "dict[str, list[str]]" is not assignable to type "Input[Mapping[str, Input[str]]] | None"
```

rhysyngsun:
@shaidar when you get a chance to re-review, it'd be good to get this out today. It relates to changes included in https://github.com/mitodl/mit-learn/pull/1727 (although things won't break).

src/ol_infrastructure/substructure/keycloak/__main__.py
```diff
@@ -356,6 +356,20 @@
                 views=["admin", "user"], edits=["admin", "user"]
             ),
         ),
+        keycloak.RealmUserProfileAttributeArgs(
+            name="emailOptIn",
+            display_name="${emailOptIn}",
+            validators=[
+                keycloak.RealmUserProfileAttributeValidatorArgs(
+                    name="integer-bool",
+                    config={"integer": {"min": 0, "max": 1}},
```
shaidar, line 365: This is throwing the following error:
```
error: keycloak:index/realmUserProfile:RealmUserProfile resource 'olapps-user-profile' has a problem: '' expected type 'string', got unconvertible type 'map[string]interface {}', value: 'map[max:1 min:0]'. Examine values at 'olapps-user-profile.attributes[5].validators'.
```

rhysyngsun: Hopefully I fixed that - keycloak exports min/max as strings (although for some reason it's not consistent because `max-length` is an integer), if that still doesn't work (the error makes me think it might not), I'll just remove the validation entirely since it's not directly user input.

shaidar (approved):

PR #2721, opened 2024-10-15
Minor tweaks to CustomUserFederation

shaidar:

Nathan noticed that syncing using SCIM wasn't running and after looking at the UI, the sync periods weren't set. Looking at pulumi docs, moved the sync_periods from the config block to the defined keys.

- [ ] Desktop screenshots

### How can this be tested?

Tried running pulumi up on the QA stack without any syntax error.


rhysyngsun (approved):
LGTM

PR #2724, opened 2024-10-15
Initial commit of the k8s work for airbyte.

Ardiea:
Deployed an airbyte CI environment running in CI.

### Checklist:
- [ ] Need to update the pipeline.
- [ ] Need to figure out what migrating the existing environments looks like. 
- [ ] Needs testing from someone who know what airbyte is and what it does.
- [ ] May be an issue or two with IAM role / powers. The lakehouse? Sounds like a nice place.


src/ol_infrastructure/applications/airbyte/__main__.py
```diff
@@ -294,12 +301,6 @@
     role=airbyte_server_instance_role.name,
 )
 
-airbyte_server_instance_profile = iam.InstanceProfile(
-    f"airbyte-server-instance-profile-{env_name}",
-    role=airbyte_server_instance_role.name,
-    path="/ol-infrastructure/airbyte-server/profile/",
-)
-
```
blarghmatey, line 302: Where does the IAM policy get bound to the service account?

Ardiea: Haven't done that bit yet.

blarghmatey:
It looks like the helm chart definition is duplicated. Can you take a look over to see if this is ready for review?

src/ol_infrastructure/applications/airbyte/__main__.py
```diff
@@ -530,157 +558,564 @@
...
+        "memory": "1Gi",
+    },
+}
+
+airbyte_helm_release = kubernetes.helm.v3.Release(
+    "airbyte-helm-release",
+    kubernetes.helm.v3.ReleaseArgs(
+        name="airbyte",
+        chart="airbyte",
+        version="1.1.0",
```
blarghmatey, line 659: We should be pulling this from an environment variable and passing that in the pipeline, falling back to what is set in the version constants lib module.

blarghmatey (approved):

PR #2748, opened 2024-10-23
Small EKS improvements

Ardiea:
- Added env var lookups for most core version numbers referenced with an EKS cluster build.
- Added an AWS API call to lookup supported K8S versions and automatically keep clusters up to date to avoid $$$ charges for falling behind. 
- Added missing requests and limits to core EKS components. 
- Removed an errant backup file.


src/ol_infrastructure/infrastructure/aws/eks/__main__.py
```diff
@@ -56,14 +58,17 @@
...
-    "VAULT_SECRETS_OPERATOR_CHART": "0.8.1",
+    "CERT_MANAGER_CHART": os.environ.get("CERT_MANAGER_CHART", "v1.16.0-beta.0"),
+    "EBS_CSI_DRIVER": os.environ.get("EBS_CSI_DRIVER", "v1.33.0-eksbuild.1"),
+    "EFS_CSI_DRIVER": os.environ.get("EFS_CSI_DRIVER", "v2.0.7-eksbuild.1"),
+    "GATEWAY_API": os.environ.get("GATEWAY_API", "v1.1.0"),
+    "EXTERNAL_DNS_CHART": os.environ.get("EXTERNAL_DNS_CHART", "1.15.0"),
+    "TRAEFIK_CHART": os.environ.get("TRAEFIK_CHART", "v31.0.0"),
+    "VAULT_SECRETS_OPERATOR_CHART": os.environ.get(
+        "VAULT_SECRETS_OPERATOR_CHART", "0.8.1"
+    ),
```
blarghmatey, line 69: Should we put those default versions into the `bridge.versions` file?

src/ol_infrastructure/infrastructure/aws/eks/__main__.py
```diff
@@ -56,14 +58,17 @@
...
+    # K8S version is special, retrieve it from the AWS APIs
+    "KUBERNETES": os.environ.get("KUBERNETES", eks_versions()[0]),
```
blarghmatey, line 71: Is there a consistent ordering of the versions returned, or should we add some sorting to the response?

src/ol_infrastructure/infrastructure/aws/eks/__main__.py
Ardiea: Probably.

src/ol_infrastructure/infrastructure/aws/eks/__main__.py
Ardiea: It seems to be sorted. I couldn't find it written down anywhere that that is the case but this is the raw response from the CLI client. 

```
{
    "addons": [
        {
            "addonName": "aws-ebs-csi-driver",
            "type": "storage",
            "addonVersions": [
                {
                    "addonVersion": "v1.36.0-eksbuild.1",
                    "architecture": [
                        "amd64",
                        "arm64"
                    ],
                    "compatibilities": [
                        {
                            "clusterVersion": "1.31",
                            "platformVersions": [
                                "*"
                            ],
                            "defaultVersion": true
                        },
                        {
                            "clusterVersion": "1.30",
                            "platformVersions": [
                                "*"
                            ],
                            "defaultVersion": true
                        },
                        {
                            "clusterVersion": "1.29",
                            "platformVersions": [
                                "*"
                            ],
                            "defaultVersion": true
                        },
                        ...
```

blarghmatey (approved):

PR #2757, opened 2024-10-28
Added configuration for keycloak client scopes

rhysyngsun:

Part of https://github.com/mitodl/hq/issues/5559

The configuration that's currently in the code ends up causing Pulumi to create a secondary "profile" scope with a generated name (e.g. `profile-abf8e`) rather than modifying the current one. These changes:

- Give the scope a well-known name that doesn't collide with `profile`.
- Add this scope to the defaults for clients in the `olapps` realm.

### How can this be tested?

I think we can test this in CI and verify on manual inspection the configuration comes out correct.

The `ol-profile` scope should show up for the app clients as a default scope.

It should look like this (we're only concerned about the `ol-profile` scope, namely that it is default and has the two attribute mappers:

Screenshot 2024-10-28 at 13-54-26 Keycloak Administration UI

Screenshot 2024-10-28 at 13-54-44 Keycloak Administration UI

### DIFF
--- a/src/ol_infrastructure/substructure/keycloak/Pulumi.substructure.keycloak.QA.yaml
+++ b/src/ol_infrastructure/substructure/keycloak/Pulumi.substructure.keycloak.QA.yaml
@@ -37,6 +37,7 @@ config:
       mitlearn: ["https://api.rc.learn.mit.edu/*", "https://rc.learn.mit.edu/*"]
       open: ["https://api.mitopen-rc.odl.mit.edu/*", "https://mitopen-rc.odl.mit.edu/*"]
       open-discussions: ["https://discussions-rc.odl.mit.edu/*"]
+    extra_default_scopes: ["ol-profile"]
     realm_name: olapps
   - client_info:
       airbyte: ["https://airbyte-qa.odl.mit.edu/*"]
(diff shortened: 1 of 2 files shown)


src/ol_infrastructure/substructure/keycloak/Pulumi.substructure.keycloak.QA.yaml
```diff
@@ -37,6 +37,7 @@ config:
       mitlearn: ["https://api.rc.learn.mit.edu/*", "https://rc.learn.mit.edu/*"]
       open: ["https://api.mitopen-rc.odl.mit.edu/*", "https://mitopen-rc.odl.mit.edu/*"]
       open-discussions: ["https://discussions-rc.odl.mit.edu/*"]
+    extra_default_scopes: ["ol-profile"]
```
rhysyngsun, line 40: @shaidar I feel like this would be better done as a property under `client_info`, but that'd involve refactoring that to be an object w/ `urls` which is probably more invasive here and better of as a separate refactor.

src/ol_infrastructure/substructure/keycloak/__main__.py
```diff
@@ -380,7 +380,7 @@
 )
 
 ol_apps_profile_client_scope = keycloak.openid.ClientScope(
-    "profile", realm_id=ol_apps_realm.id
+    "ol-profile", realm_id=ol_apps_realm.id
 )
```
shaidar, line 384: It would be good to add a `name="ol-profile"` so that it does not add the random suffix

shaidar (approved):

PR #2759, opened 2024-10-29
Md/gateway improvements

Ardiea:
Refactors the OLEKSGateway resources to be a little bit more flexible and not as abstract. Updates open-metadata to use this new version.

### DIFF
--- a/src/ol_infrastructure/components/aws/eks.py
+++ b/src/ol_infrastructure/components/aws/eks.py
@@ -26,6 +26,12 @@ class OLEKSGatewayRouteConfig(BaseModel):
     name: str
     port: int
 
+    @field_validator("hostnames")
+    @classmethod
+    def de_dupe_hostnames(cls, hostnames: list[str]) -> list[str]:
+        # Just to be safe
+        return list(set(hostnames))
+
 
 class OLEKSGatewayListenerConfig(BaseModel):
     name: str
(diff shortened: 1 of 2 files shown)


src/ol_infrastructure/components/aws/eks.py
```diff
@@ -19,13 +19,22 @@ class OLEKSGatewayRouteConfig(BaseModel):
...
+
+
+class OLEKSGatewayListenerConfig(BaseModel):
     name: str
+    hostname: str
     port: int
-    protocol: str
+    protocol: Literal["HTTPS", "HTTP"] = "HTTPS"
     tls_mode: Literal["Passthrough", "Terminate"] = "Terminate"
+    certificate_secret_name: Optional[str]
```
feoh, line 36: The comment above to the effect that cert-manager will generate this was super helpful. Could we maybe name this field in a way that signposts that?

generated_certificate_secret_name comes to mind off the top but I'm sure folks could come up with something better :)

feoh (approved):

src/ol_infrastructure/components/aws/eks.py
Ardiea: Yah I was actually thinking about that too and forgot to come back to it. The (little) complication is that you don't HAVE to use cert-manager. Let me see how I can clarify it.

blarghmatey (approved):

feoh (approved):

PR #2786, opened 2024-11-08
 Store public Keycloak cert in vault - part 2

### LINKED ISSUE #2740: Store public Keycloak cert in vault
### Expected Behavior

A successful Superset deployment should not rely on the presence of Keycloak and its reachability.
### Current Behavior

If Keycloak is unreachable, a Superset deployment fails due to the following config:
https://github.com/mitodl/ol-infrastructure/blob/main/src/ol_superset/pythonpath/superset_config.py#L84

### Possible Solution

Store Keycloak public cert along with other required config variables in Vault so that Keycloak reachability does not hinder Superset deployment.

Ardiea:
Part two of  / Closes #2740 

Do not merge until https://github.com/mitodl/ol-infrastructure/pull/2785 is fully deployed to all envs. 

Switch superset config to load the OIDC public key from vault rather than from key cloak. Allows this to run even if keycloak is unavailable.

### DIFF
--- a/src/ol_superset/pythonpath/superset_config.py
+++ b/src/ol_superset/pythonpath/superset_config.py
@@ -1,7 +1,5 @@
-import json
 import logging
 import os
-import urllib.request
 from ssl import CERT_OPTIONAL
 from typing import Optional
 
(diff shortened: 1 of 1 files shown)


src/ol_superset/pythonpath/superset_config.py
```diff
@@ -1,7 +1,5 @@
-import json
```
shaidar, line 1: Think there's a call to `json` in `CustomSsoSecurityManager` class

Ardiea: That is a call to the `json()` method off of ? Something? `oauth_remotes()`? There were no pre-commit grievances about removing the whole lib. And actually an insistence that it be removed.

PR #2826, opened 2024-12-05
Resize edxapp mitxonline prod workers

shaidar:

MITx Online edxapp worker prod instance seem to constantly reach their memory limits. Currently they're configured as `c6a.large` compute optimized instances which have 2 vCPUs and 4GB of RAM. This would switch them to memory optimized instances with 2 vCPUs and 8GB of RAM

### DIFF
--- a/src/ol_infrastructure/applications/edxapp/Pulumi.applications.edxapp.mitxonline.Production.yaml
+++ b/src/ol_infrastructure/applications/edxapp/Pulumi.applications.edxapp.mitxonline.Production.yaml
@@ -55,7 +55,7 @@ config:
   edxapp:framework: "earthly"
   edxapp:web_node_capacity: "9"
   edxapp:worker_node_capacity: "5"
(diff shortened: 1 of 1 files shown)


src/ol_infrastructure/applications/edxapp/Pulumi.applications.edxapp.mitxonline.Production.yaml
```diff
@@ -55,7 +55,7 @@ config:
   edxapp:framework: "earthly"
   edxapp:web_node_capacity: "9"
   edxapp:worker_node_capacity: "5"
-  edxapp:worker_instance_type: "c6a.large"
+  edxapp:worker_instance_type: "r8g.large"
```
blarghmatey, line 58: These are graviton instances so they'll be running ARM chips. I'm not sure if that will work for edX. Typically we go with the `a` class instances, such as `r7a.large`.

blarghmatey (approved):

PR #2828, opened 2024-12-05
Md/unifed ecomm

Ardiea:
https://github.com/mitodl/hq/issues/6201

- Creates / fixes applications-ci cluster
- Creates a build + publish pipeline off the main branch of the unified-ecommerce repo for the docker image needed. 
- Expands the build + publish pipeline to also deploy pulumi stacks based on things doof does with the `release` and `release-candidate` branches. 
- Fixes a small issue with James' ecommerce nodejs package publish pipeline. 
- Pulumi stack for unified-ecommerce
  - create s3 bucket for frontend deployment/hosting
  - create gh env secrets to allow gh actions to access s3 bucket
  - create fastly distrubtion pointing to s3 bucket
  - create route53 entries pointing to fastly distribution.
  - create K8S deployment, service, secrets, configmaps, ~gateway configuration~, and APISIX resources  for running the backend.
    - Automatically makes ~`api.pay-ci.ol.mit.edu`~ `api-pay-ci.ol.mit.edu` route53 entry. 
      - Need to use a wildcard because rather than `cert-manager` because `cert-manager` doesn't have full support for apisix custom resources. It is always somethin'.  

### TODO after this is merged
- Some kind of autoscaling on the k8s deployment resource.


src/ol_concourse/pipelines/libraries/unified_ecommerce_api_clients.py
```diff
@@ -65,7 +65,7 @@ def _read_script(script_name: str) -> str:
 unified_ecommerce_repository = git_repo(
     name=Identifier("unified-ecommerce"),
     uri=unified_ecommerce_repository_uri,
-    branch="release",
+    branch="main",
```
jkachel, line 68: Question about this - changing this will mean that the API clients rebuild whenever we push to `main`? I'm not sure we want to do that since `main` won't necessarily be production code.

jkachel (approved):
Looks OK to me. I have a question noted about the changes to the API client build pipeline, and one more:

For clarity - we'll have a separate branch for production deploy that we'll coordinate releases to via doof?

src/ol_concourse/pipelines/libraries/unified_ecommerce_api_clients.py
Ardiea: This is mostly because the `release` branch wasn't there when I started looking at this early last week. Seems like it should be changed back now that it is available.

PR #2851, opened 2024-12-13
UE API Client pipeline: install yarn explicitly

jkachel:

https://github.com/mitodl/hq/issues/5723

We're getting some errors in the publish step of the pipeline, mostly relating to the yarn version that comes with Node. Node ships with yarn 1.2 (yarn classic) and the client specifies yarn 4.2 (berry). So, this updates the publish step script to enable Corepack, which then enables using the yarn version specified in the `package.json`. 

This also updates the publish command since it's changed in yarn berry.

### How can this be tested?

dunno but this seemed to work just running it manually outside the pipeline (other than publishing, because I didn't want to publish it for real). 

### Additional Context

`yarn npm publish` doesn't give you a way to specify the version, so I think it comes out of the package manifest. It will also default to setting the tag to `latest`.


jkachel:
Probably worth noting: the `open-api-clients` pipeline has been failing for the same reason. Since the UE client build was a cut-and-paste, it suffers from the same issue. Once this is worked out, then these changes specifically should also be applied over there.

feoh (approved):

PR #2854, opened 2024-12-16
switch to large text embedding model

shanbady:

Related to https://github.com/mitodl/hq/issues/6271

This PR switches us to using open ai's large embedding model (currently we are using small)

### Additional Context
When this is deployed I will need to regenerate the embeddings.


shanbady:
will merge and coordinate on the release of this once https://github.com/mitodl/mit-learn/pull/1907 lands

PR #2876, opened 2025-01-03
config, fix: Add COURSE_LIVE_HELP_URL in mitx/cms_only.yml.tmpl

Anas12091101:
https://github.com/mitodl/hq/issues/2490

- [ ] Desktop screenshots


pdpinch:
I'm confused. Do you think this will actually help, or do you need some input from @blarghmatey as you requested in https://github.com/mitodl/hq/issues/2490#issuecomment-2569047798

Anas12091101:
This PR fixes the AttributeError for COURSE_LIVE_HELP_URL. We handled similar errors in the daily digest issue the same way. The problem seems to be that the edX app isn’t picking up default values from `common.py`. I’ve asked Tobias if there’s any particular reason why the values aren’t being read from that file. There’s also a chance the same error could happen for other attributes defined in `common.py` but not in our ol-infra files.

feoh (approved):

PR #2863, opened 2024-12-19
config: enable new studio mfe in mitx all envs

Anas12091101:
### What are the relevant tickets?
https://github.com/mitodl/hq/issues/6260

### Description (What does it do?)
This PR enables new studio MFE in all mitx envs

### Screenshots (if appropriate):

- [ ] Desktop screenshots
- [ ] Mobile width screenshots

### How can this be tested?
I manually set the waffle flags on mitx QA. Verify the working of new studio mfe there.

PR #2877, opened 2025-01-03
revert new Studio MFE waffle flags in production

pdpinch:

partially reverts https://github.com/mitodl/ol-infrastructure/pull/2863
closes https://github.com/mitodl/hq/issues/6387 

We released the new Studio MFE without testing to production. I already turned off the flags manually. This PR turns them off in the repo so it doesn't go out to production again by mistake. 

### How can this be tested?

I would appreciate it if you could double-check my work to see that this turns off all and only the waffle flags that were turned on in https://github.com/mitodl/ol-infrastructure/pull/2863


blarghmatey (approved):

Anas12091101 (approved):
LGTM 👍

PR #2914, opened 2025-01-21
fix: Split tracking log locations for LMS and CMS

blarghmatey:
https://github.com/mitodl/hq/issue/6049

Because there are both LMS and CMS containers running via docker compose, they end up competing for the ability to write to the on-disk location that Vector reads from. This splits their usage of the physical disk to be namespaced by the service type. This also updates the Vector config to watch _all_ of the tracking log files since some logs are being written to a `.1` location simultaneously with logs being written to the primary file.

### How can this be tested?

Build an edxapp instance and deploy to CI to verify that:
- Vector is reading all of the tracking log files
- The tracking log files are being written to both the lms and cms variant directories
- The tracking logs are all making it to S3

### DIFF
--- a/src/bilder/images/edxapp_v2/files/docker-compose.yaml.tmpl
+++ b/src/bilder/images/edxapp_v2/files/docker-compose.yaml.tmpl
@@ -12,7 +12,7 @@ services:
     command: ["1000", "/openedx/data", "/openedx/media"]
     restart: on-failure
     volumes:
-    - /opt/data:/openedx/data
+    - /opt/data/lms:/openedx/data
     - /opt/data/media:/openedx/media
     - /opt/staticfiles-${ENVIRONMENT_TIER}:/openedx/staticfiles:ro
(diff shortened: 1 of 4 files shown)


Ardiea (approved):

src/bilder/images/edxapp_v2/files/docker-compose.yaml.tmpl
```diff
@@ -12,7 +12,7 @@ services:
     command: ["1000", "/openedx/data", "/openedx/media"]
     restart: on-failure
     volumes:
-    - /opt/data:/openedx/data
```
quazi-h, line 15: Is there anything else under the /opt/data directory besides the logs that the lms/cms would need access to?

quazi-h (approved):
Code changes look good to me, just left a question about the volumes mounted for lms and cms-permissions since this change is removing the /opt/data directory for each to be replaced with the opt/data/lms and opt/data/cms specific directories.

PR #2924, opened 2025-01-29
Adding health checks to the Dagster docker image

quazi-h:
https://github.com/mitodl/hq/issues/6223

Added a new health-check service to list of containers for the dagster image template.
Referenced the latest ol-infra-health-checks image and added ports, volumes, and traefik labels.
Also defined explicit priority values for the two Traefik routers (will review with Mike).

### How can this be tested?
Check with Mike on the best method to test these kinds of changes.


src/bilder/images/dagster/templates/docker-compose.yaml.j2
```diff
@@ -72,3 +73,23 @@ services:
...
       - "traefik.http.middlewares.traefik-forward-auth.forwardauth.authResponseHeaders=X-Forwarded-User"
       - "traefik.http.services.traefik-forward-auth.loadbalancer.server.port=4181"
+  # per node healthcheck container
+  health-check:
+    image: 610119931565.dkr.ecr.us-east-1.amazonaws.com/dockerhub/mitodl/ol-infra-health-checks:latest
+    container_name: health-check
+    ports:
+    - "8097:8907"
+    # volumes:
+    # do I need this? Didn't need to specify when testing locally with ol-data-platform
```
Ardiea, line 83: No you probably don't need these from looking at https://github.com/mitodl/ol-infra-health-checks

src/bilder/images/dagster/templates/docker-compose.yaml.j2
```diff
@@ -72,3 +73,23 @@ services:
...
+    - /var/run/docker.sock:/var/run/docker.sock
+    - /etc/docker/compose:/etc/docker/compose
+    labels:
+      # Explicitly tell Traefik to expose this container
+      - "traefik.enable=true"
+      # Send all health_check host traffic to a router
+      - "traefik.http.routers.healthcheck.rule=Host(`0.0.0.0`)"
+      - "traefik.http.routers.healthcheck.priority=2"
+      # Specify the authentication middleware for the healthcheck router
+      - "traefik.http.routers.healthcheck.middlewares=traefik-forward-auth"
```
blarghmatey, line 93: We don't want to gate the healthcheck behind any auth because that will block our ability to use Pingdom.

blarghmatey, line 93: We don't want to gate the healthcheck behind any auth because that will block our ability to use Pingdom.

blarghmatey, line 93: We don't want to gate the healthcheck behind any auth because that will block our ability to use Pingdom.

src/bilder/images/dagster/templates/docker-compose.yaml.j2
```diff
@@ -72,3 +73,23 @@ services:
...
+    - "8097:8907"
+    # volumes:
+    # do I need this? Didn't need to specify when testing locally with ol-data-platform
+    - /var/run/docker.sock:/var/run/docker.sock
+    - /etc/docker/compose:/etc/docker/compose
+    labels:
+      # Explicitly tell Traefik to expose this container
+      - "traefik.enable=true"
+      # Send all health_check host traffic to a router
+      - "traefik.http.routers.healthcheck.rule=Host(`0.0.0.0`)"
```
blarghmatey, line 90: The routing rule that we want to use is a path prefix of `/healthcheck` so that only requests to that path and paths under it get routed to this container.

src/bilder/images/dagster/templates/docker-compose.yaml.j2
```diff
@@ -20,6 +20,7 @@ services:
     labels:
       - "traefik.enable=true"
       - "traefik.http.routers.dagit.rule=Host(`${DAGSTER_HOSTNAME}`)"
+      - "traefik.http.routers.dagit.priority=1"
```
blarghmatey, line 23: This shouldn't be necessary if using the path prefix routing for the healthcheck container.

src/bilder/images/dagster/templates/docker-compose.yaml.j2
Ardiea, line 90: Probably looking for something like 

```
Host(`${DAGSTER_HOSTNAME}`) && PathPrefix(`/healthcheck`))
```

And I think you can omit the two priority lines entirely because those are for situations when it is possible that multiple routes will match an incoming request and you need to explicitly state that one is more important than the other. In this case, The addition of `PathPrefix` will always put the healthcheck step ahead of a the earlier step. If the request doesn't have that prefix, it goes to the earlier step that is just `Host`.

blarghmatey (approved):

PR #2926, opened 2025-01-30
Md/psg attempt 2

Ardiea:

- Restructures VPC layout to make K8S worker nodes on private subnets that route through one or more nat gateways based on config.
- Changes worker nodes to use the NodeGroupV2 pulumi resource rather than the ManagedNodeGroup resource
- Configured unified ecommerce to use pod security groups.
- Opens port 53:udp:tcp to all base worker nodes for DNS traffic.


src/ol_infrastructure/lib/aws/eks_helper.py
```diff
@@ -3,10 +3,51 @@
...
+def get_default_psg_ingress_args(
+    k8s_pod_subnet_cidrs: list[str],
+    k8s_service_subnet_cidr: str,
+) -> list[ec2.SecurityGroupIngressArgs]:
+    return [
+        ec2.SecurityGroupIngressArgs(
+            protocol="-1",
+            from_port=0,
+            to_port=0,
+            cidr_blocks=["0.0.0.0/0"],
```
blarghmatey, line 32: This looks like it allows all traffic from anywhere to the pod?

blarghmatey (approved):
Aside from the one question about default ingress args this looks great!

blarghmatey (approved):
Aside from the one question about default ingress args this looks great!

blarghmatey (approved):
Aside from the one question about default ingress args this looks great!

blarghmatey (approved):
Aside from the one question about default ingress args this looks great!

Issue #2938, opened 2025-02-03
Modify IAM Configurations to Allow learn-ai Application to Use AWS Resources


feoh:
### Description/Context
1. Create an IAM role + policies + attachments
2. Create a k8s service account
3. Annotate that service account with the IAM role
4. Tell the ‘deployment’ / pods to use the service account when they run

### Plan/Design

Prior Art:

https://github.com/mitodl/ol-infrastructure/blob/main/src/ol_infrastructure/applications/airbyte/__main__.py#L584-L614

The only other thing you’ll need is to make the service account because in that example we’re relying on one that comes with the helm chart. The pulumi stuff to make one is here https://www.pulumi.com/registry/packages/kubernetes/api-docs/core/v1/serviceaccount/

there are special things about the role in the IAM config that the TrustRole does for you. AssumeRole stuff

blarghmatey:
Beyond the core mechanics of the IAM policy, this also needs to include provisions for interacting with AWS Bedrock functionality. Additionally, this should create a Vault role for generating IAM credentials with the same IAM policy bound to it that can be generated for use in developer environments.

feoh:
Decided to branch off his branch rather than PR and merge before I understood what I was merging :)

https://github.com/mitodl/ol-infrastructure/tree/cpatti_learn_ai

feoh:
I think this PR is in good shape now. Mike will review in the AM.

PR #2943, opened 2025-02-06
Md/learn ai

Ardiea:
#2938

Implements the learn-ai application in QA on K8S.

- Creates RDS and Redis Infrastructure.
- Pipeline
  - Builds and publishes a docker image for `main` and `release-candidate` branches.
  - Deploys pulumi stacks
- Defines pulumi stack and place holder configurations for a QA stack.
- Defines all pulumi code needed
  - IAM (excluding bedrock pod privs) 
  - RDS
  - Redis
  - K8S Resources
  - Vault resources
  - DNS resources
  - Fastly configuration (excluding TLS which is click-ops)
- Keycloak substructure stack changes
- Few small fixes to the unified ecommerce   

Missing Items:
- Stack secrets + configurations for CI and production


src/ol_concourse/pipelines/infrastructure/learn_ai/docker_pulumi.py
```diff
@@ -0,0 +1,320 @@
...
+    )
+
+    pulumi_resource_type = pulumi_provisioner_resource()
+    pulumi_resource = pulumi_provisioner(
+        name=Identifier("pulumi-ol-infrastructure-learn-ai-application"),
+        project_name="ol-infrastructure-learn-ai-application",
+        project_path=f"{ol_infra_repo.name}/src/ol_infrastructure/learn-ai",
+    )
+
+    learn_ai_ci_deployment_job = Job(
```
blarghmatey, line 144: Is there something in the `pulumi_job` and/or `pulumi_jobs_chain` that is incompatible with this pipeline?

src/ol_infrastructure/applications/learn_ai/Pulumi.applications.learn_ai.QA.yaml
```diff
@@ -0,0 +1,28 @@
...
+  learn_ai:backend_domain: "api-learn-ai-qa.ol.mit.edu"
+  learn_ai:db_password:
+    secure: v1:moKYAbHsHOnTUrEk:4OgRBNFXT0WKrz1KaKZDhQxT09dA36jN4+q/SO0IIAemTSiwGqrXbXn+r5dFlGrr24S+VGtiyNXSVKts+j/TgKS33FTo86OX2saBDrL4dFo=
+  learn_ai:env_vars:
+    CELERY_TASK_ALWAYS_EAGER: "false"
+    CSRF_ALLOWED_ORIGINS: '["https://learn-ai-qa.ol.mit.edu", "https://api-learn-ai-qa.ol.mit.edu"]'
+    CSRF_TRUSTED_ORIGINS: '["https://learn-ai-qa.ol.mit.edu", "https://api-learn-ai-qa.ol.mit.edu"]'
+    DEBUG: "true"
+    DEV_ENV: "true"
+    ENVIRONMENT: "dev"
```
blarghmatey, line 15: Do we want these to be true for QA? Wouldn't that be more suited for CI?

src/ol_infrastructure/applications/learn_ai/__main__.py
```diff
@@ -0,0 +1,1236 @@
...
+github_provider = github.Provider(
+    "github_provider",
+    owner=read_yaml_secrets(Path("pulumi/github_provider.yaml"))["owner"],
+    token=read_yaml_secrets(Path("pulumi/github_provider.yaml"))["token"],
+)
+
+k8s_global_labels = {
+    "ol.mit.edu/stack": stack_info.full_name,
+    "ol.mit.edu/service": "learn-ai",
+}
```
blarghmatey, line 92: This is something we should probably turn into a Pydantic object similar to `AWSBase` so that we can have consistent labels/tags especially for cost allocation via Vantage. (in a follow-up PR, not a blocker for this work)

blarghmatey (approved):

src/ol_concourse/pipelines/infrastructure/learn_ai/docker_pulumi.py
```diff
@@ -0,0 +1,320 @@
...
+                build_parameters={
+                    "CONTEXT": learn_ai_release_candidate_repo.name,
+                },
+                build_args=[],
+            ),
+            PutStep(
+                put=learn_ai_registry_rc_image.name,
+                params={
+                    "image": "image/image.tar",
+                    "additional_tags": f"./{learn_ai_release_candidate_repo.name}/.git/ref",  # Should contain a tag if doof is doing his job
```
blarghmatey, line 120: I think this wants to be `.git/tag`

Ardiea: https://github.com/concourse/git-resource?tab=readme-ov-file#additional-files-populated

blarghmatey: 👍

src/ol_concourse/pipelines/infrastructure/learn_ai/docker_pulumi.py
Ardiea: Yeah, it has been a while because I first did this in December with UE so I'm not sure the technical reasons at a glance but the gist comes down to these pipelines being more closely integrated with the application / code repos than they are with ol-inf. We are triggering docker build for `main` and deploying to CI automatically but `release-candidate` and `release` are totally separate from that with their own docker build (not build**s**!) 

Rather than trying to jam these changes into `pulumi_jobs_chain`, or even using `pulumi_job` (which I don't think was meant to be used independently of `pulumi_jobs_chain`?) I decided to just recreate here. 

What we should do is come back to it and make a library function to do this just like `pulumi_jobs_chain` that follows this adapted pattern because I think it is something we're going to need a lot more of as we move away from heroku.

Ardiea: https://github.com/mitodl/ol-infrastructure/issues/2945

src/ol_infrastructure/applications/learn_ai/__main__.py
Ardiea: https://github.com/mitodl/ol-infrastructure/issues/2946

src/ol_infrastructure/applications/learn_ai/Pulumi.applications.learn_ai.QA.yaml
Ardiea: This is a good question that I just don't know the answer to? Sure?

src/ol_concourse/pipelines/infrastructure/learn_ai/docker_pulumi.py
blarghmatey: The `pulumi_job` function was actually intended for independent use when necessary. We can easily re-use it for the single case and then construct a shorter `chain` using the RC and Production stages.

blarghmatey: But I agree that we should probably build a helper to simplify the replication of this pattern as we migrate more workloads into this approach.

blarghmatey (approved):

PR #2983, opened 2025-02-21
Use a different suffix for logout url when apisix is enabled on mit-learn

mbertrand:
Related to https://github.com/mitodl/mit-learn/pull/2061

Adds ` MITOL_API_LOGOUT_SUFFIX: "logout/oidc"` so the frontend will send the user there (apisix) first to log out, which should then redirect the user back to the url specified by `APISIX_LOGOUT_URL`.

Assumption is that APISIX will be configured with the [same login/logout urls as defined in this file](https://github.com/mitodl/mit-learn/blob/3bb5d4ffe8dfea45eb80678158b7049f5997356b/config/apisix/apisix.yaml) - is that correct, @blarghmatey ?

### How can this be tested?
N/A

### DIFF
--- a/src/ol_infrastructure/applications/mitlearn/Pulumi.applications.mitlearn.Production.yaml
+++ b/src/ol_infrastructure/applications/mitlearn/Pulumi.applications.mitlearn.Production.yaml
@@ -40,6 +40,7 @@ config:
     INDEXING_API_USERNAME: "od_mm_prod_api"
     KEYCLOAK_BASE_URL: "https://sso.ol.mit.edu"
     KEYCLOAK_REALM_NAME: "olapps"
+    MITOL_API_LOGOUT_SUFFIX: "logout"
     MITOL_API_URL: "https://api.learn.mit.edu"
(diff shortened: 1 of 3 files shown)


Ardiea (approved):

ChristopherChudzicki (commented):
@mbertrand can you add
```
gh_workflow_learn_ai_syllabus_endpoint_env_secret = github.ActionsSecret(
    f"ol_mitopen_gh_workflow_mitol_api_logout_suffix-{stack_info.env_suffix}",
    repository=gh_repo.name,
    secret_name=f"MITOL_API_LOGOUT_SUFFIX_{env_var_suffix}",
    plaintext_value=f"{heroku_vars['MITOL_API_LOGOUT_SUFFIX']}",
    opts=ResourceOptions(provider=github_provider, delete_before_replace=True),
)
```
to the end of `src/ol_infrastructure/applications/mitlearn/__main__.py` to make the vars available in github actions?

@blarghmatey or @Ardiea can you double check ☝️

blarghmatey (approved):

PR #2993, opened 2025-02-26
Upgrade Hugo, Go, Node for ocw-course-publisher

ibrahimjaved12:
https://github.com/mitodl/hq/issues/6718

Upgrades `Hugo`, `Go`, and `Node` to `0.142.0`, `0.123.0` (Go version as mentioned in Hugo's current documentation), and `22` respectively.
Also adds Renovate annotations in Dockerfile and updates `README.txt` to `README.md` with correct pipeline name.


blarghmatey (approved):
We can also add annotations in this to allow Renovate to manage these versions similar to what was done in https://github.com/mitodl/ol-infrastructure/blob/main/src/bridge/lib/versions.py and https://github.com/mitodl/ol-infrastructure/blob/main/renovate.json5#L17-L27

blarghmatey (approved):

blarghmatey (approved):

Issue #2933, opened 2025-01-31
Onboarding/Offboarding users to data platform realm


shaidar:
### User Story

- As a user, I want access to apps in the data platform realm.
- As an admin, I want to invite someone to the data platform realm.
- As an admin, I want to ensure that only users no longer have access to the data platform realm once they leave the institute and/or no longer require access to any of the apps in that realm.
### Description/Context

Currently we have no process in place for inviting users or allowing a user to request access to some of the apps under the data platform realm. Additionally, there is no clear method of off boarding users who no longer need or are no longer affiliated with the Institute.
### Acceptance Criteria

- [ ] A clear process (ideally automated, but can be manual for now) to grant someone access to apps in the realm
  - [ ] MIT users
  - [ ] Non-MIT users
- [ ] A clear process (ideally automated, but can be manual for now) to revoke a users access to apps in the realm
  - [ ] MIT users
  - [ ] Non-MIT users

### Limitations/Considerations
- Keycloak does not have an email invite feature out of the box
- SAML claims don't include all the groups that an MIT user belongs to and the only ones being sent as part of a successful auth are names with `ol-` prefix
- There are different clients in the realm and we need to ensure that granting a user access to an app in the realm, does not necessarily give them access to the other apps within that realm
- A user won't have any roles mapped to them and thus access would be limited. Once an invite is accepted and a user account is created, that user can be manually assigned roles depending on the requirements.

### Plan/Design

 
- Touchstone integration offers us the following:
  - User familiarity with authentication flow
  - User account management

However, those advantages, at this point don't seem to outweigh the limitations/issues we're running into trying to integrate it with how we would like to manage users in the data platform realm.

I propose the following:
- Remove Touchstone Integration
  - If down the line, Keycloak makes changes that remove some of the stated blockers, we should be able to re-enable the Touchstone integration and link existing MIT accounts with the IdP.
- Enable organizations in the data-platform realm and initially setup two orgs (MIT, non-MIT)
  - At this stage, the org feature is mainly used to handle email invites (no other Keycloak built-in feature for that) and group users for potentially later filtering/targeting
- Pre-create user accounts through the Pulumi Keycloak provider
  - We would need to maintain a private username list that Pulumi can operate on and a commit to that list would trigger a concourse pipeline to add/remove user.
[...]

PR #2994, opened 2025-02-26
Split Superset CI build

shaidar:

https://github.com/mitodl/ol-infrastructure/issues/2933#issuecomment-2669260679

This should split the superset CI pipeline into its own build so that we can iterate on the working to configure authz between Keycloak and Superset without disrupting QA/Prod.

### DIFF
--- a/src/bilder/images/superset/deploy.py
+++ b/src/bilder/images/superset/deploy.py
@@ -1,6 +1,8 @@
 import io
+import json
 import os
 from pathlib import Path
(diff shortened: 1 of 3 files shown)


blarghmatey:
I think you'll also need to add logic in the PyInfra code and/or docker compose file for the AMI build to be able to use the CI image in the runtime. Otherwise that looks good.

blarghmatey (approved):

PR #2995, opened 2025-02-27
add OCW_STUDIO_DELETABLE_CONTENT_TYPES

pdpinch:
This env setting controls which types of objects can be deleted

N/A, but related to https://github.com/mitodl/ocw-studio/pull/2384


pdpinch:
I turned on the posthog feature flag for deleting items in Studio so I could delete a page in production, but it wasn't working -- even though it worked fine on RC. @pt2302 identified that this environment variable was set on RC in Heroku but not in production. I set it manually, but it should go into here for persistence.

umar8hassan (approved):
Looks good 👍

PR #3032, opened 2025-03-12
Different session cookie names for learn, learn-ai, xpro, mitxonline

mbertrand:
Closes https://github.com/mitodl/hq/issues/6917

Sets a unique value per app/environment for `SESSION_COOKIE_NAME` in the following apps:
- learn-ai
- mit-learn
- xpro
- mitxonline

### Additional Context
I left micromasters and ODL video service alone as well as the edx apps, not sure if this change is needed there.


blarghmatey:
This all looks good. I'm just wondering if we want to pull in xPRO yet? As far at the edx instances I think we're already customizing the cookie names.

mbertrand:
I can remove it for xpro, and anything else that doesn't need the change soon.  Maybe just keep the changes for mit-learn and learn-ai?

mbertrand:
I removed the xpro changes; if preferable I can remove the mitxonline changes too.

blarghmatey (approved):

PR #3039, opened 2025-03-14
Fix mediaconvert component resource for use with OVS and OCW Studio

blarghmatey:
N/A

The initial pass of the mediaconvert component resource threw errors and wouldn't execute. This addresses those failures.

- [ ] Desktop screenshots

### How can this be tested?

The changes have been applied manually in OVS CI and mediaconvert functionality can be tested there.

### Additional Context

We will want to be careful applying this to OCW Studio because it will delete and re-create the mediaconvert resources. Presumably that won't cause any errors, but I'm not certain of how that service works and whether continuity of the resource is required.


umar8hassan (approved):
Approved with separate resources for now to finish up OVS related stuff.

umar8hassan:
@blarghmatey It is a nitpick, but please update the docstring for the `get_standard_policy_statements` function that still lists `account_id` in args

Ardiea (approved):

PR #3079, opened 2025-03-31
set Posthog env vars for published OCW sites

gumaerc:
Prerequisite for https://github.com/mitodl/hq/issues/7017

This PR sets some env vars to have a separate Posthog project for the published OCW site vs `ocw-studio` itself. 

### How can this be tested?
- Make sure the env vars are properly set
- Not sure how else to test, probably has to get into RC / QA first

### DIFF
--- a/src/ol_infrastructure/applications/ocw_studio/__main__.py
+++ b/src/ol_infrastructure/applications/ocw_studio/__main__.py
@@ -378,6 +378,13 @@ heroku_vars = {
     "VIDEO_TRANSCODE_QUEUE": ocw_studio_mediaconvert.queue.name,
 }
 
+if stack_info.env_suffix.lower() != "production":
+    heroku_vars["PUBLISH_POSTHOG_ENABLED"] = "True"
+    heroku_vars["PUBLISH_POSTHOG_API_HOST"] = "https://ph.rc.learn.mit.edu"
+    heroku_vars["PUBLISH_POSTHOG_PROJECT_API_KEY"] = (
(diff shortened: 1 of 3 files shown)


src/ol_infrastructure/applications/ocw_studio/__main__.py
```diff
@@ -378,6 +378,13 @@
     "VIDEO_TRANSCODE_QUEUE": ocw_studio_mediaconvert.queue.name,
 }
 
+if stack_info.env_suffix.lower() != "production":
```
blarghmatey, line 381: What's the reasoning for not running against production? Is it because of the proxy?

gumaerc: The reasoning here is cost. With the amount of traffic OCW gets, having it on the production site is incurring far more events than we originally thought. I think that we will put up a PR to disable analytics in the client though instead: https://posthog.com/docs/libraries/js#opt-out-of-data-capture. The question right now is surrounding the question of "is the cost still going to be too much with production OCW asking if a feature flag is enabled on every page load?"

blarghmatey: Rather than putting these in as a conditional, it would be cleaner to just add them to the stack configs. That way it's easier to manage separate values per environment.

gumaerc:
@blarghmatey The decision was made to leave Posthog disabled in the OCW frontend on production for now, and I updated this PR based on your suggestion.

blarghmatey (approved):

PR #3089, opened 2025-04-04
Use posthog managed proxy

shaidar:

https://github.com/mitodl/hq/issues/6690

This change is to enable the use of [Posthog's managed proxy](https://posthog.com/docs/advanced/proxy/managed-reverse-proxy) for learn QA. 

### How can this be tested?

Once the change is deployed, verify that posthog data is still making it to its destination.


blarghmatey:
This should probably also be added to the learn-ai app

shaidar:
Agreed, but i wanted to initially test it out on one app and make sure it's working as expected and then we can create the other proxies.

blarghmatey (approved):

Issue #3073, opened 2025-03-28
move mitlearn to k8s


Ardiea:
### Description/Context
Need to move mitlearn out of heroku and into kubernetes. 

### Plan/Design
- Productionalize the docker file
- Create application configs. 
- Build a CI environment.
- Fgure out a migration path / design
-

PR #3094, opened 2025-04-07
Md/mitlearn ci env

Ardiea:
Starts https://github.com/mitodl/ol-infrastructure/issues/3073

A basic implementation of mitlearn-ci on K8S. Bunch of changes here:

- Creates an opensearch stack
- Refactor out some heroku stuff to make it more generically named. 
- Refactor how the secret names are passed with `OLApplicationK8sConfiguration`
- Split out K8SVaultSecret resources into their own function / file to make working with this stack a little easier. (Gemini both populated all these secrets for me and also split them out to their own file).
- Updated `OLApplicationK8sConfiguration` to let us pass in a uwsgi config if we need it in addition to an nginx config. 
- Rest of the configuration needed for mitlearn-ci and conditionals around what this env needs vs what another env needs (external service envs still in heroku).


src/ol_infrastructure/applications/mitlearn/Pulumi.applications.mitlearn.CI.yaml
```diff
@@ -0,0 +1,91 @@
...
+    MITOL_LOG_LEVEL: "INFO"
+    MITOL_NEW_USER_LOGIN_URL: "https://ci.learn.mit.edu/onboarding"
+    MITOL_NOINDEX: "true"
+    MITOL_SUPPORT_EMAIL: "odl-learn-ci-support@mit.edu" # Need to verify
+    MITPE_API_ENABLED: "true"
+    OCW_ITERATOR_CHUNK_SIZE: 300
+    OIDC_ENDPOINT: "https://sso-ci.ol.mit.edu/realms/olapps"
+    OPENSEARCH_INDEX: "mitlearn-ci"
+    OPENSEARCH_SHARD_COUNT: 2
+    OPENSEARCH_URL: "https://search-opensearch-mitlearn-ci-q7fuxrc2j26me44ymcftz3ahba.us-east-1.es.amazonaws.com"
```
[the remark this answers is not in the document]
Ardiea: It is actually correct that this no longer needs to be public but this change is out of scope for this PR. Making these private is a project in its own right. We can just leave it public for now because it is a pattern that works.

src/ol_infrastructure/applications/mitlearn/__main__.py
blarghmatey, line 1: To match the expectations of the k8s_apps pipeline we'll want to rename the directory for this project to `mit_learn` (I know, I advocated for a single word when we were renaming... 🤷 )

src/ol_infrastructure/applications/mitlearn/__main__.py
```diff
@@ -371,9 +380,13 @@
...
     options={"version": 2},
     description="Storage of configuration secrets used by MIT-Open",
-    opts=ResourceOptions(delete_before_replace=True),
+    opts=ResourceOptions(
+        delete_before_replace=True, depends_on=[mitopen_vault_iam_role]
+    ),
 )
 
+# There is a reason, I think, why these are still at `bridge/secrets/mitopen`
+# and not `bridge/secrets/mitlearn` -- Open Discussions
```
blarghmatey, line 389: Not a good reason. We should rename it. There are many other lingering renaming tasks that we should try to address. Anything that doesn't require downtime should be included in this work.

src/ol_infrastructure/applications/mitlearn/__main__.py
```diff
@@ -1329,60 +1010,49 @@
 }
 
 # Combine two var sources above with values explictly defined in pulumi configuration
-heroku_vars.update(**heroku_interpolated_vars)
-heroku_vars.update(**heroku_app_config.get_object("vars"))
+env_vars.update(**interpolated_vars)
+env_vars.update(**mitlearn_config.get_object("vars"))
 
 # Making these `get_secret_*()` calls children of the seemigly un-related vault mount `secret-mitopen/` tricks
```
blarghmatey, line 1016: As long is it's not a huge amount of work, we should replace the `secret-mitopen` mount with an equivalent `secret-mitlearn` mount

src/ol_infrastructure/applications/mitlearn/__main__.py
```diff
@@ -1584,10 +1246,611 @@
...
+        keys=[
+            consul.KeysKeyArgs(
+                path="edxapp/learn-api-domain",
+                delete=True,
+                value=mit_learn_api_domain,
+            )
+        ],
+        opts=xpro_consul_opts,
+    )
+
```
blarghmatey, line 1370: The xPRO consul keys block should move out of the `CI` conditional now that we've got a CI learn environment.

src/ol_infrastructure/components/services/k8s.py
```diff
@@ -132,6 +137,27 @@ def __init__(
             ol_app_k8s_config.application_lb_service_port_name
         )
 
+        # Determine the full name of the container image
+        if ol_app_k8s_config.application_image_repository_suffix:
+            app_image = f"{ol_app_k8s_config.application_image_repository}{ol_app_k8s_config.application_image_repository_suffix}:{ol_app_k8s_config.application_docker_tag}"
```
blarghmatey, line 142: This seems like a good location to add in the ECS repo prefix for image caching and reducing Dockerhub rate limit consumption.

Ardiea: Do you have an example of that?

src/ol_infrastructure/applications/mitlearn/__main__.py
Ardiea: what about the meta pipeline cares about it being `mitlearn` or mit_learn`?

src/ol_infrastructure/applications/mitlearn/__main__.py
Ardiea: I accomplished this via a symbolic link because I don't want to break anything I don't have to break and I also don't want to spend time digging through the odl-open-discussions stack and refactoring in there.

src/ol_infrastructure/components/services/k8s.py
blarghmatey: https://github.com/mitodl/ol-infrastructure/blob/main/src/bilder/images/dagster/templates/docker-compose.yaml.j2#L6

src/ol_infrastructure/applications/mitlearn/__main__.py
blarghmatey: It takes the "app_name" and uses it for resolving both the GH repo and the Pulumi path. We can tweak the pipeline to allow for more parameters, or conform the Pulumi to what it expects. e.g. https://github.com/mitodl/ol-infrastructure/blob/main/src/ol_concourse/pipelines/infrastructure/k8s_apps/docker_pulumi.py#L61 and https://github.com/mitodl/ol-infrastructure/blob/main/src/ol_concourse/pipelines/infrastructure/k8s_apps/docker_pulumi.py#L109

blarghmatey (approved):

PR #3103, opened 2025-04-10
Md/karpenter 2

Ardiea:
Implements karpenter on all OL k8s clusters. 

Will install Karpenter on all of our clusters and setup IAM permissions to allow it to do the needful from within the cluster via a standard OLTrustRole. Creates default EC2NodeClass and an default NodePool with reasonable spend limits to keep us from going bankrupt if it goes crazy.


src/ol_infrastructure/substructure/aws/eks/__main__.py
```diff
@@ -668,3 +679,350 @@
...
+    message_retention_seconds=300,
+    sqs_managed_sse_enabled=True,
+    tags=aws_config.merged_tags({"Name": cluster_name}),
+)
+
+karpenter_interruption_queue_policy = aws.sqs.QueuePolicy(
+    f"{cluster_name}-karpenter-interruption-queue-policy",
+    queue_url=karpenter_interruption_queue.id,
+    policy=Output.all(
+        queue_arn=karpenter_interruption_queue.arn, partition=aws.get_partition()
```
blarghmatey, line 701: Doesn't seem like the partition is getting used here?

src/ol_infrastructure/substructure/aws/eks/__main__.py
```diff
@@ -668,3 +679,350 @@
...
+)
+
+
+# Install Karpenter Helm Chart
+karpenter_release = kubernetes.helm.v3.Release(
+    f"{cluster_name}-karpenter-helm-release",
+    kubernetes.helm.v3.ReleaseArgs(
+        name="karpenter",
+        chart="oci://public.ecr.aws/karpenter/karpenter",
+        version="1.3.2",  # Specify a version for stability
```
blarghmatey, line 826: This should go in `versions.py` to be managed by Renovate

src/ol_infrastructure/substructure/aws/eks/__main__.py
```diff
@@ -668,3 +679,350 @@
         delete_before_replace=True,
     ),
 )
+
+
+############################################################
+# Install Karpenter for automatically growing and shrinking
+# the cluster
+############################################################
```
blarghmatey, line 687: This change seems substantial enough to merit starting to split into a module in this namespace for importing. That way changes for Karpenter can be scoped to just that module.

feoh (approved):

PR #3104, opened 2025-04-10
OCW Posthog QA env var name

shaidar:

Setup a Posthog managed proxy and updated the value to point to it.


gumaerc (changes requested):
Can we also set this proxy in `src/ol_infrastructure/applications/ocw_studio/__main__.py`?

blarghmatey (approved):

blarghmatey (approved):

PR #3106, opened 2025-04-10
Add HorizontalPodAutoscaling to our k8s, this time with feeling :)

feoh:
This is a continuation of https://github.com/mitodl/ol-infrastructure/pull/3065 - closed due to merge conflict pandemonium.

feat: Add a grotesquerie that will one day be a memory based HPA :\

Fixes https://github.com/mitodl/hq/issues/6979

Add a HorizontalPodAutoScaler to our k8s deployments with options for memory and CPU based metrics scaling.

DO NOT MERGE. This was me struggling with finding a way to create a complex object instantiation with two disparate paths and it can't stay this way.

### How can this be tested?

TBD. Probably involves load testing. Bees with Machine guns or some similar package?

### DIFF
--- a/src/ol_infrastructure/components/services/k8s.py
+++ b/src/ol_infrastructure/components/services/k8s.py
@@ -65,6 +65,8 @@ class OLApplicationK8sConfiguration(BaseModel):
     application_namespace: str
     application_lb_service_name: str
     application_lb_service_port_name: str
+    application_max_replicas: int = 10
+    application_min_replicas: int = 2
     k8s_global_labels: dict[str, str]
     env_from_secret_names: list[str]
     application_security_group_id: Output[str]
@@ -98,7 +100,6 @@ class OLApplicationK8sConfiguration(BaseModel):
             ),
         ),
     ]
-    target_cpu_utilization_percentage: int = 50
     import_nginx_config: bool = Field(default=True)
     import_uwsgi_config: bool = Field(default=False)
     resource_requests: dict[str, str] = Field(
@@ -420,8 +421,8 @@ class OLApplicationK8s(ComponentResource):
                         f"{ol_app_k8s_config.application_name}-app"
                     ),
                 ),
-                min_replicas=2,  # Minimum number of replicas
-                max_replicas=10,  # Maximum number of replicas
+                min_replicas=ol_app_k8s_config.application_max_replicas,  # Minimum number of replicas
+                max_replicas=ol_app_k8s_config.application_min_replicas,  # Minimum number of replicas
                 # Corrected parameter name from "metrics" to the proper name for the API
                 metrics=ol_app_k8s_config.hpa_scaling_metrics,
                 # Optional: behavior configuration for scaling
--- a/src/ol_infrastructure/components/services/k8s.py
+++ b/src/ol_infrastructure/components/services/k8s.py
@@ -4,7 +4,6 @@ This is a service components that replaces a number of "boilerplate" kubernetes
 calls we currently make into one convenient callable package.
 """
 
-from enum import Enum
 from pathlib import Path
 from typing import Any, Literal, Optional
 
@@ -53,15 +52,6 @@ class OLApplicationK8sCeleryWorkerConfig(BaseModel):
         default={"cpu": "500m", "memory": "600Mi"},
     )
     replicas: int = 1  # In lieu of a proper autoscaler 20250407
-
-
-class HPAScalerType(Enum):
-    """
-    Potential types of metric to autoscale on.
-    """
-
-    CPU = 1
-    MEMORY = 2
 
 
 # Refactor this to just be 'Config' rather than 'Configuration' - but not today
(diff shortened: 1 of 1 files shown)


src/ol_infrastructure/components/services/k8s.py
```diff
@@ -77,11 +81,33 @@ class OLApplicationK8sConfiguration(BaseModel):
...
-    hpa_scaler_type: HPAScalerType = Field(default=HPAScalerType.CPU)
+    # You get CPU and memory autoscaling by default
+    hpa_scaling_metrics: list[kubernetes.autoscaling.v2.MetricSpecArgs] = [
+        kubernetes.autoscaling.v2.MetricSpecArgs(
+            type="Resource",
+            resource=kubernetes.autoscaling.v2.ResourceMetricSourceArgs(
+                name="CPU",
+                target=kubernetes.autoscaling.v2.MetricTargetArgs(
+                    type="Utilization",
+                    average_utilization=80,  # Target memory utilization (80%)
```
blarghmatey, line 96:
```suggestion
                    average_utilization=80,  # Target CPU utilization (80%)
```
Assuming this is copy 🍝

src/ol_infrastructure/components/services/k8s.py
```diff
@@ -56,6 +56,10 @@ class OLApplicationK8sCeleryWorkerConfig(BaseModel):
 
 
 class HPAScalerType(Enum):
+    """
+    Potential types of metric to autoscale on.
+    """
+
     CPU = 1
     MEMORY = 2
```
blarghmatey, line 64: Do we still need this with the updated logic path?

blarghmatey (commented):
Looks great! I think that in any case where we have an HPA specified we need to remove any explicit replica counts.

src/ol_infrastructure/components/services/k8s.py
```diff
@@ -71,10 +71,34 @@ class OLApplicationK8sConfiguration(BaseModel):
...
+            resource=kubernetes.autoscaling.v2.ResourceMetricSourceArgs(
+                name="memory",  # Memory utilization as the scaling metric
+                target=kubernetes.autoscaling.v2.MetricTargetArgs(
+                    type="Utilization",
+                    average_utilization=80,  # Target memory utilization (80%)
+                ),
+            ),
+        ),
+    ]
+    target_cpu_utilization_percentage: int = 50
```
blarghmatey, line 101: This doesn't seem like it's having any impact any more

src/ol_infrastructure/components/services/k8s.py
```diff
@@ -377,6 +402,51 @@ def __init__(
...
+            "application-hpa",
+            spec=kubernetes.autoscaling.v2.HorizontalPodAutoscalerSpecArgs(
+                scale_target_ref=kubernetes.autoscaling.v2.CrossVersionObjectReferenceArgs(
+                    api_version="apps/v1",
+                    kind="Deployment",
+                    name=truncate_k8s_metanames(
+                        f"{ol_app_k8s_config.application_name}-app"
+                    ),
+                ),
+                min_replicas=2,  # Minimum number of replicas
```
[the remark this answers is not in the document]
feoh: How does this interact with your prior comment about EITHER replicas OR hpa? Should these all be 0 when HPA is in play?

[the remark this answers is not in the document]
blarghmatey: This is for the target replicas in the hpa. We don't want to scale down to 0 because of boot time delays

src/ol_infrastructure/components/services/k8s.py
```diff
@@ -412,8 +421,8 @@ def __init__(
                         f"{ol_app_k8s_config.application_name}-app"
                     ),
                 ),
-                min_replicas=2,  # Minimum number of replicas
-                max_replicas=10,  # Maximum number of replicas
+                min_replicas=ol_app_k8s_config.application_max_replicas,  # Minimum number of replicas
+                max_replicas=ol_app_k8s_config.application_min_replicas,  # Minimum number of replicas
```
blarghmatey, line 425: These values are flipped

src/ol_infrastructure/components/services/k8s.py
```diff
@@ -59,12 +59,14 @@ class OLApplicationK8sConfiguration(BaseModel):
     model_config = ConfigDict(arbitrary_types_allowed=True)
 
     project_root: Path
-    application_replicas: int = 1
+    application_replicas: int = 0
```
blarghmatey, line 62: I think we can probably get rid of this in favor of the `min_replicas` field, or vice versa. In the case where there are no scaling metrics passed, then we use the static replica count, otherwise we provision the HPA configs.

feoh: OK. I can't figure out how to implement the "either replicas OR ASG" check you mentioned previously given this, so I removed it.

PR #3120, opened 2025-04-16
refactor: slots config files and add footer slot

Anas12091101:
https://github.com/mitodl/hq/issues/5689

This PR adds the Footer Slot configs and refactors the existing slots configuration files for better consistency and maintainability across MFEs

#### Residential Header and Footer (Same across all MFEs)
Screenshot 2025-04-17 at 12 56 36 PM

#### MITxOnline Header and Footer (Same across all MFEs)
Screenshot 2025-04-17 at 12 46 29 PM

#### MITxPRO Header and Footer (Same across all MFEs)
Screenshot 2025-04-17 at 12 51 04 PM

#### Mobile View: Same for all MFEs
Image

### How can this be tested?
- Try to follow the steps in [pipeline.py](https://github.com/mitodl/ol-infrastructure/blob/9ac75365ccfdd4045f63d3741a1b58a49b580eaa/src/ol_concourse/pipelines/open_edx/mfe/pipeline.py) on your local setup
- Validate that the header and footer loads perfectly for all the MFEs for all the applications (xPro, MitxOnline, and MITx Residential)
- Validate that the logos and links are correct and working in the header and footer
- Validate that the remoteTutorDrawer is working fine for MITxPRO and MITxOnline learning MFE configurations.


src/ol_concourse/pipelines/open_edx/mfe/pipeline.py
```diff
@@ -163,34 +175,27 @@ def mfe_job(
...
+    mfe_plugin_types = [
+        "learning",
+        "discussions",
+        "ora-grading",
+        "communications",
+        "gradebook",
+        "learner-dashboard",
+        "authoring",
+    ]
+    if OpenEdxMicroFrontend[mfe_name].value in mfe_plugin_types:
```
blarghmatey, line 192: If we're using a single definition of slots for all MFEs now, then do we still need this conditional?

Anas12091101: I think this condition only adds the config files for the mfes listed in the list above. For any mfes not in that list, it doesn't copy the config files. A similar condition existed before this [change](https://github.com/mitodl/ol-infrastructure/blob/main/src/ol_concourse/pipelines/open_edx/mfe/pipeline.py#L187). I think we can remove this condition as [these are](https://github.com/mitodl/ol-infrastructure/blob/main/src/bridge/settings/openedx/types.py#L34) the only MFEs which we are using. I'll remove it in the next commit.

src/bridge/settings/openedx/mfe/slot_config/mitxonline/learning-mfe-config.env.jsx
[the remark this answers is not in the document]
Anas12091101: I believe @asadali145 would be better suited to answer this. From what I understand, the Tutor drawer is only required for the Learning MFE in MITxOnline and xPRO. That’s why we created a separate config file specifically for it.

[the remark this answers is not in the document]
asadali145: We do not install the Smoot Design in any other MFE. Import and initialization will fail if we add it to all MFEs. One option is to install Smoot design in all MFEs, but that's not a good idea.

src/bridge/settings/openedx/mfe/slot_config/mitx-staging/common-mfe-config.env.jsx
```diff
@@ -0,0 +1,175 @@
...
+const LearningHeaderUserMenu = (widget) => {
+  widget.content.items = [
+    {
+      href: `${configData.LMS_BASE_URL}/dashboard`,
+      message: 'Dashboard',
+    },
+    {
+      href: `${configData.LMS_BASE_URL}/logout`,
+      message: 'Sign Out',
+    },
```
asadali145, line 48: We should use userMenu here as we used it in DesktopHeaderUserMenu. This goes for the config files.

src/ol_concourse/pipelines/open_edx/mfe/pipeline.py
```diff
@@ -163,34 +175,27 @@ def mfe_job(
         tar -xvzf mitodl-smoot-design*.tgz
         mv package mitodl-smoot-design
         """
-        learning_mfe_slot_config_file = "learning-mfe-config"
+        slot_config_file = "learning-mfe-config"
+        copy_common_config = f"cp {mfe_configs.name}/src/bridge/settings/openedx/mfe/slot_config/{open_edx_deployment.deployment_name}/common-mfe-config.env.jsx {mfe_build_dir.name}/common-mfe-config.env.jsx"  # noqa: E501
     else:
         mfe_smoot_design_overrides = ""
```
asadali145, line 181: Can we move this with the other 2 variables? This will remove the else block.

asadali145 (commented):
Just a code review. I am going to test the config files now.

asadali145 (changes requested):
Also, we have branding overrides to install mitodl footer in the MFEs. default is defind in default_branding_overrides, we should remove it and the related code. @blarghmatey does that sound right?

src/bridge/settings/openedx/mfe/slot_config/Footer.jsx
```diff
@@ -0,0 +1,207 @@
...
+          <li className="mx-2">
+            <Hyperlink destination={`${marketingSiteUrl}/honor-code/`}>Honor Code</Hyperlink>
+          </li>
+        </>
+      )}
+      <li className="mx-2">
+        <Hyperlink destination={`${marketingSiteUrl}/${isMITX ? "tos/" : "terms-of-service/"}`}>
+          Terms of Service
+        </Hyperlink>
+      </li>
```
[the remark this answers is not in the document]
Anas12091101: The marketing site url for residential is defined [here](https://github.com/mitodl/ol-infrastructure/blob/main/src/ol_concourse/pipelines/open_edx/mfe/values.py#L44).

asadali145 (commented):
Testing looks good. I will approve after the requested changes.

I am assuming that the footer design and links are already approved by @pdpinch

asadali145 (approved):
LGTM!

Adding for reference:
- We have added the footer component similar to frontend-base, having all the subslots.
- We have overridden the slots in the config file.
- When frontend-base footer is ready, we can remove our footer component and the footer slot overrides and only override the frontend-base footer subslots. This will simplify our config files.

blarghmatey (approved):

PR #3128, opened 2025-04-18
feat: Add HPA configs to Celery worker deployments

blarghmatey:
N/A

Adds autoscaling to our Celery deployments in K8s

### How can this be tested?

Deploy it and then fire off a bunch of heavy tasks (or wait for it to happen) to see if the number of Celery pods goes up


feoh (approved):
LGTM but Copilot is right about the comment text mismatch :)

PR #3145, opened 2025-04-24
Fix learn-ai frontend query string and edX CORS

blarghmatey:
N/A

- Properly handles routing to the index.html when there is a query string present
- Adds CORS permission in edX for the learn-ai frontend

### How can this be tested?

Deploy to CI/QA and verify functionality


ChristopherChudzicki (approved):
👍 

PR all makes sense, except the question I asked in slack re "Where does the actual `learn-ai-qa.ol.mit.edu` value live?"

PR #3158, opened 2025-04-29
Add transcoding related values to consul tmplate.

feoh:

Related to mitodl/odl-video-service#1181

Adds transcoding related values to consul. Probably badly. Please check *carefully*.

### DIFF
--- a/src/bilder/images/odl_video_service/files/.env.tmpl
+++ b/src/bilder/images/odl_video_service/files/.env.tmpl
@@ -65,6 +65,14 @@ VIDEO_S3_SUBTITLE_BUCKET="{{ keyOrDefault "ovs/s3_subtitle_bucket_name" "missing
 VIDEO_S3_THUMBNAIL_BUCKET="{{ keyOrDefault "ovs/s3_thumbnail_bucket_name" "missing_value_s3_thumbnail_bucket_name" }}"
 VIDEO_S3_TRANSCODE_BUCKET="{{ keyOrDefault "ovs/s3_transcode_bucket_name" "missing_value_s3_transcode_bucket_name" }}"
 VIDEO_S3_WATCH_BUCKET="{{ keyOrDefault "ovs/s3_watch_bucket_name" "missing_value_s3_watch_bucket_name" }}"
+OVS_AWS_ROLE_NAME="{{ keyOrDefault "ovs/aws_role_name": ovs_mediaconvert.role.name }}"
+OVS_AWS_ACCOUNT_ID="{{ keyOrDefault "ovs/aws_account_id": aws_account.account_id}}"
+OVS_POST_TRANSCODE_ACTIONS="{{ keyOrDefault "ovs/post_transcode_actions": "cloudsync.api.process_transcode_results" }}"
+OVS_TRANSCODE_ENDPOINT="{{ "ovs/video_s3_transcode_endpoint": transcode_endpoint" }}"
+OVS_VIDEO_TRANSCODE_PREFIX="{{ "ovs/video_s3_upload_prefix": "" }}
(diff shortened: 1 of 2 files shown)


src/bilder/images/odl_video_service/files/.env.tmpl
```diff
@@ -65,6 +65,14 @@ VIDEO_S3_SUBTITLE_BUCKET="{{ keyOrDefault "ovs/s3_subtitle_bucket_name" "missing
...
 VIDEO_S3_TRANSCODE_BUCKET="{{ keyOrDefault "ovs/s3_transcode_bucket_name" "missing_value_s3_transcode_bucket_name" }}"
 VIDEO_S3_WATCH_BUCKET="{{ keyOrDefault "ovs/s3_watch_bucket_name" "missing_value_s3_watch_bucket_name" }}"
+OVS_AWS_ROLE_NAME="{{ keyOrDefault "ovs/aws_role_name": ovs_mediaconvert.role.name }}"
+OVS_AWS_ACCOUNT_ID="{{ keyOrDefault "ovs/aws_account_id": aws_account.account_id}}"
+OVS_POST_TRANSCODE_ACTIONS="{{ keyOrDefault "ovs/post_transcode_actions": "cloudsync.api.process_transcode_results" }}"
+OVS_TRANSCODE_ENDPOINT="{{ "ovs/video_s3_transcode_endpoint": transcode_endpoint" }}"
+OVS_VIDEO_TRANSCODE_PREFIX="{{ "ovs/video_s3_upload_prefix": "" }}
+OVS_VIDEO_S3_TRANSCODE_PREFIX="{{ "ovs/video_s3_transcode_prefix": "transcoded" }}"
+OVS_VIDEO_S3_THUMBNAIL_PREFIX="{{ "ovs/video_s3_thumbnail_prefix": "thumbnails" }}"
+OVS_VIDEO_TRANSCODE_QUEUE="{{ "ovs/video_transcode_queue": ovs_mediaconvert.queue.name }}"
```
Ardiea, line 75: So the format would be

```
VAR_NAME="{{ keyOrDefault "consul-key" "string literal default value" }}"
```
The quotes and spaces are important and you've got colons in all of these which won't work. The first three use `keyOrDefault` but the next five do not. I don't think those that omit it will work at all. They need some kind of function call, I believe. This is (essentially) go-templates. 👞 💥 🔫 

Lines 68, 69, and 71 seems like you're referencing python values from the code below in `__main__.py` rather than string literals defaults. 

Honestly, I like to use "missing_value_<env_var_in_lower_case>" as my string literal default so it is clear that consul-template has failed to find what it seeks. See line 65-67 as examples.

src/ol_infrastructure/applications/odl_video_service/__main__.py
```diff
@@ -707,6 +707,17 @@
...
+    "ovs/aws_account_id": aws_account.account_id,
+    "ovs/post_transcode_actions": "cloudsync.api.process_transcode_results",
+    "ovs/transcode_job_template": "./config/mediaconvert.json",
+    "ovs/video_s3_thumbnail_bucket": ovs_config.get("s3_thumbnail_bucket_name"),
+    "ovs/video_s3_transcode_bucket": ovs_config.get("s3_transcode_bucket_name"),
+    "ovs/video_s3_transcode_endpoint": secrets["transcode_endpoint"],
+    "ovs/video_s3_upload_prefix": "",
+    "ovs/video_s3_transcode_prefix": "transcoded",
+    "ovs/video_s3_thumbnail_prefix": "thumbnails",
+    "ovs/video_transcode_queue": ovs_mediaconvert.queue.name,
```
Ardiea, line 720: This ... looks good! As long as you feel good about all the values you're puking in here 👍

Ardiea (changes requested):

Ardiea (approved):

PR #3165, opened 2025-05-02
Frontend related changes for new learn-ai ui sandbox option

mbertrand:
Related to https://github.com/mitodl/learn-ai/pull/157 and https://github.com/mitodl/learn-ai/pull/172

Frontend variable setup for new learn-ai sandbox option

### How can this be tested?
N/A


src/ol_infrastructure/applications/learn_ai/__main__.py
```diff
@@ -1143,6 +1143,20 @@
     value=frontend_vars["OPENEDX_LOGIN_URL"],
     opts=ResourceOptions(provider=github_provider, delete_before_replace=True),
 )
+gh_workflow_api_base_env_var = github.ActionsVariable(
+    f"learn-ai-gh-workflow-ai-elasticsearch-url-variablet-{stack_info.env_suffix}",
```
shaidar, line 1147: Should this be `variable` instead of `variablet` or am I missing something?

mbertrand: I was wondering about that myself, but all the others above it are also `variablet`, so I decided to make these the same for consistency and also just in case it's supposed to be that way for some reason.

shaidar: It's just a name and doesn't really have any functional impact. So either way works

mbertrand: ok, I'll change them all then

shaidar (approved):

PR #3184, opened 2025-05-06
Update env.config.js to avoid double-bundling smoot-design

ChristopherChudzicki:
https://github.com/mitodl/hq/issues/7259

Loads the remote tutor drawer code from the static directory to avoid re-bundling (via webpack) the already-bundled code.

It works again!

Screenshot 2025-05-06 at 2 08 16 PM

### How can this be tested?

Follow the steps outlined in https://github.com/mitodl/open-edx-plugins/tree/cc/update-readme/src/ol_openedx_chat. In particular, steps 3 and 4 have changed slightly. You can see the diff here: https://github.com/mitodl/open-edx-plugins/pull/494

### DIFF
--- a/src/bridge/settings/openedx/mfe/slot_config/mitxonline/learning-mfe-config.env.jsx
+++ b/src/bridge/settings/openedx/mfe/slot_config/mitxonline/learning-mfe-config.env.jsx
@@ -1,3 +1,4 @@
+import { getConfig } from '@edx/frontend-platform';
 import config from './common-mfe-config.env.jsx';
 
 import(
@@ -10,7 +11,7 @@ import(
   /* webpackIgnore: true */
  "/static/smoot-design/remoteTutorDrawer.es.js").then(module => {
    module.init({
-     messageOrigin: "http://local.openedx.io:8000",
+     messageOrigin: getConfig().LMS_BASE_URL,
      transformBody: messages => ({ message: messages[messages.length - 1].content }),
    })
 })
--- a/src/bridge/settings/openedx/mfe/slot_config/mitxonline/learning-mfe-config.env.jsx
+++ b/src/bridge/settings/openedx/mfe/slot_config/mitxonline/learning-mfe-config.env.jsx
@@ -1,11 +1,18 @@
-import { getConfig } from '@edx/frontend-platform';
 import config from './common-mfe-config.env.jsx';
-import * as remoteTutorDrawer from "./mitodl-smoot-design/dist/bundles/remoteTutorDrawer.umd.js"
 
-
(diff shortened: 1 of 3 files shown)


src/bridge/settings/openedx/mfe/slot_config/mitxonline/learning-mfe-config.env.jsx
```diff
@@ -1,11 +1,18 @@
...
-  messageOrigin: getConfig().LMS_BASE_URL,
-  transformBody: messages => ({ message: messages[messages.length - 1].content }),
+import(
+  /**
+   * remoteTutorDrawer is already bundled to include its own version
+   * of React and ReactDOM.
+   *
+   * Add webpackIgnore to avoid bundling it again.
+   */
+  /* webpackIgnore: true */
```
ChristopherChudzicki, line 10: I don't entirely understand what was causing the error in https://github.com/mitodl/hq/issues/7259

However, previously:
1. We were bundling the code (including transpile/minification) via Vite during `smoot-design`'s build phase. *The purpose of this pre-bundling was largely to avoid conflicts with OpenEdx's version of React*)
2. The code was getting bundled (including minification/transpile) again via Webpack during the MFE build.

I noticed that replacing the synchronous import with an async import loading smoot via https://unpkg.com/@mitodl/smoot-design@6.4.0/dist/bundles/remoteTutorDrawer.es.js made the error go away entirely. This leads me to believe the double-bundling was causing problems.

So, lets not double bundle.

ChristopherChudzicki: BTW: We could use https://unpkg.com/@mitodl/smoot-design@6.4.0/dist/bundles/remoteTutorDrawer.es.js in production rather than copying and loading from the static folder.

However, I don't think this a great idea. Unpkg does work pretty well, but I don't think it's intended to be a production CDN.

jonkafton (approved):
Makes sense and lgtm.

src/bridge/settings/openedx/mfe/slot_config/mitxonline/learning-mfe-config.env.jsx
```diff
@@ -1,11 +1,18 @@
...
+ "/static/smoot-design/remoteTutorDrawer.es.js").then(module => {
+   module.init({
+     messageOrigin: "http://local.openedx.io:8000",
```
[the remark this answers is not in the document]
ChristopherChudzicki: Yikes. Thank you for catching that. 🤦

Issue #3119, opened 2025-04-15
Keycloak metrics


shaidar:
### Description/Context

We currently rely on https://github.com/aerogear/keycloak-metrics-spi to expose Keycloak metrics which we then ship to hosted Grafana through [vector](https://github.com/mitodl/ol-infrastructure/blob/454add77476340d9df6a8834cf9da5e67a5c85a1/src/bilder/images/keycloak/templates/vector/keycloak_logs_metrics.yaml#L6). The new [release of Keycloak](https://www.keycloak.org/docs/latest/release_notes/index.html#guides-for-metrics-and-grafana-dashboards) provides guides to two Grafana dashboards and the plugin that we have been relying on does not appear to be that active lately given that the last commit was about 5 months ago.

We should remove the metrics plugin from our pipeline and make the necessary Keycloak changes to expose Keycloak metrics with the built-in feature.

shaidar:
The keycloak metrics endpoint along with the healthcheck one are both configured and up and running on CI. Turning on events in the UI, doesn't appear to enable user and admin events for the metrics endpoint. Had to pass another startup flag to get the metrics endpoint to show those stats. Setup the Grafana dashboards from the Keycloak repo however i wasn't seeing any metrics flowing through. Verified that the vector client is scraping the new metrics url (port 9000) and verified that the metrics endpoint has data for the stats that the Grafana dashboard is querying for.

blarghmatey:
Closed: Keycloak runs on Kubernetes with metrics-enabled and a ServiceMonitor (applications/keycloak/__main__.py:530). The aerogear metrics SPI and the vector-based EC2 pipeline this described were removed along with the AMI build (#3833).

PR #3202, opened 2025-05-12
Keycloak metrics

shaidar:

https://github.com/mitodl/ol-infrastructure/issues/3119

This enables Keycloak metrics based on https://www.keycloak.org/observability/configuration-metrics. From a keycloak instance, I can load the results from the localhost:9000/metrics endpoint, however after adding the Grafana dashboards - https://www.keycloak.org/observability/grafana-dashboards, I'm not seeing any of the data in Grafana yet.

### DIFF
--- a/src/bilder/images/keycloak/files/docker-compose.yaml
+++ b/src/bilder/images/keycloak/files/docker-compose.yaml
@@ -11,6 +11,9 @@ services:
     - --proxy-headers=xforwarded
     - --cache-stack=postgres-jdbc-ping-tcp
     - --cache-config-file=cache-ispn-jdbc-ping.xml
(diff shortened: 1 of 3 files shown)


blarghmatey:
This will require updating the Vector configuration for the Keycloak AMI to run a Prometheus metrics scrape against that endpoint. Prior art here: https://github.com/mitodl/ol-infrastructure/blob/main/src/bilder/images/concourse/templates/vector/concourse_metrics.yaml

shaidar:
@blarghmatey That already exists as we had a metrics endpoint using a separate spi

blarghmatey:
Copy that.

blarghmatey (approved):

PR #3225, opened 2025-05-22
nextjs pipeline changes

Ardiea:
Updated the k8s_apps pipeline to work for mit-learn-nextjs. Tested both meta and actual pipelines. This will allow us to build and deploy k8s app stacks with a little more flexibility.

### DIFF
--- a/src/ol_concourse/pipelines/infrastructure/k8s_apps/meta.py
+++ b/src/ol_concourse/pipelines/infrastructure/k8s_apps/meta.py
@@ -17,7 +17,7 @@ from ol_concourse.lib.models.pipeline import (
 from ol_concourse.lib.resources import git_repo
 
 
-def meta_job(app_name: str) -> Job:
+def meta_job(app_name: str, deployment_config: dict[str, str]) -> Job:
     return Job(
         name=Identifier(f"create-{app_name}-pipeline"),
         plan=[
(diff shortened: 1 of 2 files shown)


src/ol_concourse/pipelines/infrastructure/k8s_apps/meta.py
```diff
@@ -115,8 +122,29 @@ def meta_pipeline(app_names: list[str]) -> Pipeline:
...
+            "repo_name": "learn-ai",
+            "dockerfile_path": "./Dockerfile",
+        },
+        "unified-ecommerce": {
+            "repo_name": "unified-ecommerce",
+            "dockerfile_path": "./Dockerfile",
+        },
+        "mit-learn": {
+            "repo_name": "mit-learn",
+            "dockerfile_path": "./Dockerfile",
```
blarghmatey, line 136: Does this want the `./`?

blarghmatey (approved):

src/ol_concourse/pipelines/infrastructure/k8s_apps/meta.py
Ardiea: It doesn't mind it and I kind of like being explicit. This is clearly requesting the dockerfile in the root of repository.

PR #3235, opened 2025-05-28
switch to using http endpoint for open telemetry

shanbady:

Closes https://github.com/mitodl/hq/issues/7406

This PR switches us to using the http endpoint for open telemetry (instead of GRPC). There is an issue where it seems like since switching to k8s the opentelemetry endpoint/service may go away at times leading the requests to the GRPC endpoint to continuously retry. There are some folks experiencing something similar and apparently [switching to the http endpoint](https://github.com/open-telemetry/opentelemetry-collector/issues/6363#issuecomment-2816635115) fixes the continuous retries and allows it to gracefully reconnect when available.

### How can this be tested?
This is something we will have to deploy to rc and see if we experience the same celery task buildup we are seeing

### Additional Context
If this does indeed fix the issue of the celery task queue stalling - we may still need a separate ticket to investigate what makes that opentelemetry endpoint go away in the first place.


blarghmatey (approved):
This change should also be made to the CI stack for consistency and early validation

PR #3244, opened 2025-05-30
Md/k8s apps pipeline updates

Ardiea:
For the k8s_apps pipeline, Uses the pydantic model as a configuration map instead of a dictionary.
Reverts previous changes to pipeline. 
Also apparently adds a disk upgrade to core k8s nodes in applications clusters.

### DIFF
--- a/src/ol_concourse/pipelines/infrastructure/k8s_apps/docker_pulumi.py
+++ b/src/ol_concourse/pipelines/infrastructure/k8s_apps/docker_pulumi.py
@@ -1,7 +1,9 @@
-import sys
-from typing import Any, Optional
+# ruff: noqa: PLR0913
 
-from pydantic import BaseModel
+import sys
+from typing import Any, Optional, Union
+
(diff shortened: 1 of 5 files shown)


src/ol_concourse/pipelines/infrastructure/k8s_apps/docker_pulumi.py
```diff
@@ -34,23 +34,31 @@
...
+        return self
 
 
 pipeline_params = {
     "mitxonline": AppPipelineParams(app_name="mitxonline", build_target="production"),
     "mit-learn-nextjs": AppPipelineParams(
-        app_name="mit-learn-nextjs", build_target="build_skip_yarn"
+        app_name="mit-learn-nextjs",
+        build_target="build_skip_yarn",
+        repo_name="mit-learn",
```
blarghmatey, line 54: Doesn't this also need a separate dockerfile path?

blarghmatey (approved):

PR #3263, opened 2025-06-04
config: Add header rewrite for fixing logout bug from edX

blarghmatey:
https://github.com/mitodl/mitxonline/pull/2704

Adds headers to allow for logout via gateway from MITx Online

### How can this be tested?

Apply in CI and/or QA and test once Nathan's PR is applied


rhysyngsun (approved):
LGTM

PR #3315, opened 2025-06-18
add csrf cookie domain and allowed / trusted origins for mitxonline a…

gumaerc:
N/A

This PR adds the MITx Online and MIT Learn domains to `CORS_ALLOWED_ORIGINS` as well as `CSRF_TRUSTED_ORIGINS`, and sets the `CSRF_COOKIE_DOMAIN` on MITx Online deployments to be `.mitxonline.mit.edu` for now.


gumaerc:
@blarghmatey I scoped the domains more tightly for now, but eventually as we discussed MITx Online and the Learn frontend will need to be able to have domains that match the same wildcard, because the Learn frontend will be doing POST / PUT / PATCH to the MITx Online backend.

PR #3380, opened 2025-07-15
Start indexing learn.mit.edu

ChristopherChudzicki:
Followup to https://github.com/mitodl/hq/issues/7696 and https://github.com/mitodl/hq/issues/7426

- Removes MITOL_NOINDEX env var from the backend stack. It is unused.
- Set MITOL_NOINDEX on the nextjs app. Only production should be indexed.

### How can this be tested?
Deploy and check vars.

### DIFF
--- a/src/ol_infrastructure/applications/mit_learn/Pulumi.applications.mit_learn.CI.yaml
+++ b/src/ol_infrastructure/applications/mit_learn/Pulumi.applications.mit_learn.CI.yaml
@@ -66,7 +66,6 @@ config:
     MITOL_COOKIE_NAME: "learn-ci"
     MITOL_LOG_LEVEL: "INFO"
     MITOL_NEW_USER_LOGIN_URL: "https://ci.learn.mit.edu/onboarding"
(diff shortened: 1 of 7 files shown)


src/ol_infrastructure/applications/mit_learn/Pulumi.applications.mit_learn.CI.yaml
```diff
@@ -66,7 +66,6 @@ config:
     MITOL_COOKIE_NAME: "learn-ci"
     MITOL_LOG_LEVEL: "INFO"
     MITOL_NEW_USER_LOGIN_URL: "https://ci.learn.mit.edu/onboarding"
-    MITOL_NOINDEX: "true"
```
ChristopherChudzicki, line 69: The django server doesn't use this for anything. It is currently being set in `settings.py` but has a default, and
- https://github.com/mitodl/mit-learn/pull/2361
is open to remove it.

feoh (approved):

PR #3401, opened 2025-07-22
Add mitxonline keycloak redirect uri config

rhysyngsun:

Follow up to https://github.com/mitodl/ol-infrastructure/pull/3388

This adds valid redirect uris which were missing in the original client config I reused in my previous PR. This covers the redirect URI for when someone is doing an email/password change.

### How can this be tested?

Test in CI/RC?


src/ol_infrastructure/substructure/keycloak/__main__.py
```diff
@@ -794,6 +794,9 @@ def fetch_realm_public_key(keycloak_url: str, realm_id: str) -> str:
     standard_flow_enabled=False,
     implicit_flow_enabled=False,
     service_accounts_enabled=True,
+    valid_redirect_uris=keycloak_realm_config.get_object(
```
Ardiea, line 797:
```suggestion
    valid_redirect_uris=keycloak_realm_config.require_object(
```

PR #3404, opened 2025-07-23
chore: bump rapid response xblock

arslanashraf7:
None

- Bumps the rapid response version
- There is no additional functionality change in the new version, it is just that the implementation of the existing functionality was change. The change is only associated to the course outline on CMS.
- The implementation that was changed can be seen in https://github.com/mitodl/open-edx-plugins/pull/546
- The newer implementation uses edX's built-in update/publish mechanism. There is a screencast on the PR https://github.com/mitodl/open-edx-plugins/pull/546

- [ ] Desktop screenshots

### How can this be tested?
Once done, we should be able to test the rapid response as per the screencast shared in https://github.com/mitodl/open-edx-plugins/pull/546


blarghmatey (approved):
Once the Sumac changes are removed we can merge this.

PR #3412, opened 2025-07-25
mongodb output updates 

Ardiea:
Added two outputs to mongodb atlas stacks with information needed by edxapp running in k8s.


blarghmatey:
This looks good overall. One thing I experienced a while ago with these exports is that sometimes trying to re-use the same output object with different attributes will end up returning an empty value for the second call to it. Presuming you're not seeing that then this should be good to merge.

PR #3493, opened 2025-08-20
config: remove mit_et from UAI_COURSE_KEY_FORMATS

Anas12091101:
https://github.com/mitodl/hq/issues/8010

This PR removes the `MIT_ET` prefix from `UAI_COURSE_KEY_FORMATS`, since the correct prefix is `UAI_ET`. The `UAI_ET` case is already covered by the `UAI_` prefix.

- [ ] Desktop screenshots


Anas12091101:
> Please get an approval from @pdpinch.

Already discussed this with @pdpinch on Slack.

PR #3533, opened 2025-09-02
First attempt at changing OIDC realm to use olapps

dsubak:
After some brief discussion [here](https://mitodl.slack.com/archives/G1VJQHQHK/p1756835317176429?thread_ts=1756835110.397759&cid=G1VJQHQHK) I took a swing at getting the realm and client switched over to use the existing SSO values in vault.

Two main blockers:
- ~Unsure if this is the right approach. Rather than explicitly trying to reuse a vault secret by specifying it's path, folks might prefer if I just copy the relevant values into `sso/jupyterhub` - not sure what is preferred!~
- ~It's not actually working. Even after `pulumi up` it's giving me old values. I've tried rolling the jupyter namespaced pods, but I kinda didn't expect that to change things - do I need to kick apisix pods in some way/shape/form?~ Tobias rolled some state and corrected a few missing redirect urls, works in CI now!

- [ ] Desktop screenshots

### DIFF
--- a/src/ol_infrastructure/applications/jupyterhub/__main__.py
+++ b/src/ol_infrastructure/applications/jupyterhub/__main__.py
@@ -239,7 +239,7 @@ oidc_resources = OLApisixOIDCResources(
         oidc_use_session_secret=True,
         oidc_scope="openid email",
         vault_mount="secret-operations",
-        vault_path="sso/jupyterhub",
+        vault_path="sso/mitlearn",
         vaultauth=vault_k8s_resources.auth_name,
     ),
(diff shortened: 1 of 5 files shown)


blarghmatey:
That all looks correct. It's possible that the Vault secret is taking a little while to resolve. One way to prompt that to happen is by deleting the K8s secret associated with the Vault secret. That will force the Vault Secrets Operator to reconcile the secret, thereby recreating it (or erroring in a way that is visible).

blarghmatey (approved):

PR #3541, opened 2025-09-04
Md/gpu clusters

Ardiea:
- Creates a new class of core nodes that are equipped with GPUs.
- Taints these nodes by default. 
- Installs the nvidia_k8s_device_plugin and configures it to only run on nodes that have GPUs (as labeled by us). 
- Modifies the default node security groups to allow all traffic from within the k8s address space to within the k8s address space. There is probably a more elegant way to accomplish what is needed (cross talk between all node security groups to all node security groups...) but I don't see it.

### DIFF
--- a/src/ol_infrastructure/infrastructure/aws/eks/__main__.py
+++ b/src/ol_infrastructure/infrastructure/aws/eks/__main__.py
@@ -451,7 +451,6 @@ k8s_provider = kubernetes.Provider(
 
 # Loop through the node group definitions and add them to the cluster
 node_groups = []
-node_group_security_groups = []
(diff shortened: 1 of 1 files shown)


src/ol_infrastructure/applications/jupyterhub/__main__.py
```diff
@@ -239,7 +239,7 @@
         oidc_use_session_secret=True,
         oidc_scope="openid email",
         vault_mount="secret-operations",
-        vault_path="sso/jupyterhub",
+        vault_path="sso/mitlearn",
```
dsubak, line 242: I think this should already in main fwiw

blarghmatey (approved):

dsubak (approved):

PR #3544, opened 2025-09-04
feat: Added attempt at kubectl plugin.

feoh:
mitodl/hq#7186

Me taking a swing at adding kubectl plugin support to our botkube.

### DIFF
--- a/src/ol_infrastructure/applications/botkube/__main__.py
+++ b/src/ol_infrastructure/applications/botkube/__main__.py
@@ -148,7 +148,7 @@ botkube_application = kubernetes.helm.v3.Release(
             "commonLabels": k8s_global_labels,
             "sources": {},
             "executors": {
-                "tools": {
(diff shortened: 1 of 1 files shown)


src/ol_infrastructure/applications/botkube/__main__.py
```diff
@@ -155,7 +155,17 @@
...
                                 "name": f"#{slack_channel}",
                                 "bindings": {
-                                    "executors": ["k8s-default-tools"],
+                                    "plugins": {
+                                        "repositories": {
+                                            "botkube": {
+                                                "url": "https://github.com/kubeshop/botkube/releases/download/v1.14.0/plugins-index.yaml"
+                                            },
+                                        },
+                                    },
```
blarghmatey, line 164: I don't think the plugins bit goes under the communications block. I think executors also wants to be its own top-level block at the same level as sources and communications.

PR #3549, opened 2025-09-05
Md/gpu clusters 2

Ardiea:

1. Transitioned the karpenter CRD management to its own helm chart (this is the recommended approach and I missed it back-when...). Needed to do some patch jank to allow the new helm release to take over the existing CRDs but works like a charm. 
2. Added gpu node management to karpenter and validated through. 

### How can this be tested?
`kubectl get nodes -o custom-columns=NODE:.metadata.name,TAINTS:.spec.taints -w`
To watch the status of current nodes with taints. 

In a browser spawn a bunch of nodes that need GPU. When the resource threshold is tripped, a new node should be added by karpenter.

### DIFF
--- a/src/ol_infrastructure/substructure/aws/eks/karpenter.py
+++ b/src/ol_infrastructure/substructure/aws/eks/karpenter.py
@@ -10,6 +10,7 @@ from pulumi import Output, ResourceOptions, StackReference
 from bridge.lib.magic_numbers import AWS_EVENT_TARGET_GROUP_NAME_MAX_LENGTH
 from bridge.lib.versions import KARPENTER_CHART_VERSION
 from ol_infrastructure.components.aws.eks import OLEKSTrustRole, OLEKSTrustRoleConfig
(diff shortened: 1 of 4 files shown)


src/ol_infrastructure/substructure/aws/eks/karpenter.py
```diff
@@ -392,3 +437,79 @@ def setup_karpenter(  # noqa: PLR0913
...
+                            "effect": "NoSchedule",
+                        }
+                    ],
+                    "expireAfter": "24h",
+                    "terminationGracePeriod": "30m",
+                    "requirements": [
+                        {
+                            "key": "node.kubernetes.io/instance-type",
+                            "operator": "In",
+                            "values": ["g4dn.xlarge", "g4dn.2xlarge"],
```
blarghmatey, line 482: It would be good to move this into an Enum that we can manage in the `lib/` module space for evolving the instance types. Similarly for the other node types that we want set in the default nodepool.

blarghmatey (approved):

PR #3556, opened 2025-09-08
Remove all menu items except for `Run`

dsubak:
https://github.com/mitodl/hq/issues/8395

After discussion w/ Ferdi and Peter, we wanted to look into reducing the number of UI elements that are not necessary. This PR accomplishes that in two ways:
- When possible, we use overrides provided by [core JupyterLab configurables](https://jupyterlab.readthedocs.io/en/stable/extension/extension_points.html#id17). In this case, we leverage the mainmenu and toolbar extensions.
- When we must, we disable the extensions providing certain bits of on-screen functionality. In this case we disable the debugger, nbdime, and the logo entrypoint.

Here's what the current overrides do. We show a file and run menu, with a few options disabled, as well as a save and run button.
Screenshot 2025-09-09 at 11 14 59 AM
Screenshot 2025-09-09 at 11 15 03 AM
Screenshot 2025-09-09 at 11 15 08 AM

### How can this be tested?
These changes are up on nb.ci.learn.mit.edu, so simply using any course in that environment will show the new UI treatment.

### DIFF
--- a/src/ol_infrastructure/applications/jupyterhub/menu_override.json
+++ b/src/ol_infrastructure/applications/jupyterhub/menu_override.json
@@ -84,6 +84,10 @@
   "@jupyterlab/notebook-extension:panel": {
     "toolbar": [
       {
+        "name": "save",
+        "disabled": true
+      },
+      {
         "name": "insert",
         "disabled": true
       },


pdpinch:
Thanks for the proof of concept. Are there any functions we want to keep besides runs? Is there a way for folks to export the notebook / save it from the browser?

PR #3557, opened 2025-09-08
config: Support OIDC vault login

blarghmatey:
N/A

Updates the EKS script to use GH token _or_ OIDC login

### How can this be tested?

Run the script and verify that if you have no GH token in your env that it logs you in via OIDC


shaidar (approved):
Looks good overall other than the hassle of having to login to all three Keycloak envs

Issue #3562, opened 2025-09-09
Transition MITxOnline to Serveless Valkey as Redis Replacement


feoh:
Transition MITxOnline to Serveless Valkey as Redis Replacement

- [x] Create Pulumi ServerlessCache Component
- [ ] Integrate that component into MITx Online in addition to existing redis
- [ ] Deploy in CI and restore from current snapshot
- [ ] Do another deploy removing the old style cache

feoh:
Also I [tagged](https://github.com/mitodl/ol-infrastructure/releases/tag/CPATTI_SERVERLESS_VALKEY) the existing work in case the branch gets deleted.

feoh:
I'm not convinced this will ever make sense from a cost/reward perspective. Closing. Feel free to re-open if that's a misread.

PR #3563, opened 2025-09-09
Elasticache Serverless implementation with default usage in CI and QA

feoh:
#3562 

Add a serverless Valkey cache to help replace current instance based

### How can this be tested?

pulumi up

### DIFF
--- a/src/ol_infrastructure/applications/edxapp/__main__.py
+++ b/src/ol_infrastructure/applications/edxapp/__main__.py
@@ -1,5 +1,4 @@
 # ruff: noqa: E501
-
 """Provision and deploy the resources needed for an edxapp installation.
 
 - Create S3 buckets required by edxapp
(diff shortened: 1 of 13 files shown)


Ardiea (commented):
LGTM. 

Extra thoughts though. Should this really be its own component resource or should it be an implementation of [OLAmazonCacheConfig](https://github.com/mitodl/ol-infrastructure/blob/main/src/ol_infrastructure/components/aws/cache.py#L37-L89) named `OLAmazonServerlessValkeyConfig()`.

Honestly, though I'm suggesting it, I think I like your implementation better just because the serverless configuration is really SOOO MUCH simpler than traditional elasticache configuration. `OLAmazonCacheConfig` comes with so much stuff that just isn't needed. 

Only other thought is you've created one in the mitxonline stack but you don't actually point anything at it?

Ardiea (approved):

feoh:
This PR is ancient. If we ever want to undertake this effort we'll make it an epic, but this PR is way past its sell by date.

PR #3576, opened 2025-09-15
Build images for UAI notebooks

dsubak:
Related ticket https://github.com/mitodl/hq/issues/8533

Adds new Concourse builds for updated UAI notebook images

### DIFF
--- a/src/ol_concourse/pipelines/container_images/jupyter_courses.py
+++ b/src/ol_concourse/pipelines/container_images/jupyter_courses.py
@@ -24,6 +24,8 @@ class CourseImageInfo:
 # Used to make a parameterized pipeline which builds
 # and publishes the image from GH to ECR
 courses = [
+    # TODO (dsubak) - Remove the first 4 once the rest are tested # noqa: FIX002, TD004
+    # TODO (dsubak) - Derive from course IDs # noqa: FIX002, TD004
     CourseImageInfo(
         course_name="course_deep_learning_foundations_and_applications",
         repo_uri="git@github.mit.edu:ol-notebooks-qa/course_deep_learning_foundations_and_applications.git",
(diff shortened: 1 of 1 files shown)


src/ol_concourse/pipelines/container_images/jupyter_courses.py
```diff
@@ -44,6 +46,41 @@ class CourseImageInfo:
...
+        image_name="UAI_SOURCE-UAI.11-3T2025",
+    ),
+    CourseImageInfo(
+        course_name="UAI_SOURCE-UAI.12-3T2025",
+        repo_uri="git@github.mit.edu:ol-notebooks-qa/UAI_SOURCE-UAI.12-3T2025.git",
+        image_name="UAI_SOURCE-UAI.12-3T2025",
+    ),
+    CourseImageInfo(
+        course_name="UAI_SOURCE-UAI.13-3T2025",
+        repo_uri="git@github.mit.edu:ol-notebooks-qa/UAI_SOURCE-UAI.13-3T2025.git",
```
blarghmatey, line 81: Did we change the org name yet? Probably worth doing and updating all of the images in this before we go much further.

src/ol_concourse/pipelines/container_images/jupyter_courses.py
```diff
@@ -44,6 +46,41 @@ class CourseImageInfo:
         repo_uri="git@github.mit.edu:ol-notebooks-qa/course_clustering_and_descriptive_ai.git",
         image_name="clustering_and_descriptive_ai",
     ),
+    CourseImageInfo(
+        course_name="UAI_SOURCE-UAI.6-3T2025",
+        repo_uri="git@github.mit.edu:ol-notebooks-qa/UAI_SOURCE-UAI.6-3T2025.git",
+        image_name="UAI_SOURCE-UAI.6-3T2025",
```
blarghmatey, line 52:
```suggestion
        image_name="uai_source-uai.6",
```
Including a term specifier seems like it will become problematic. If we need to include that information it should probably be set as a label on the image.

Also, only lowercase letters are permitted in the image name https://docs.aws.amazon.com/AmazonECR/latest/userguide/repository-create.html

src/ol_concourse/pipelines/container_images/jupyter_courses.py
dsubak: I haven't renamed it yet - I figured I'd do so once the initial crop of these images were built and in ECR. 

Minor lift either way since this is the only place that cares about the org name at the moment, but it is on my list to do!

src/ol_concourse/pipelines/container_images/jupyter_courses.py
dsubak: No issue taking the term specifier out of the image name and/or the course name (nothing upstream or downstream really cares, and it's only in the repo name because it matches the mitxonline GH repo naming scheme) - I'll it pulled out of both for the time being. It's easy enough to change that if we need the info later.

Ah, didn't realize about those naming restrictions; will get that fixed shortly!

blarghmatey (approved):

PR #3578, opened 2025-09-16
config: redirect anonymous users to login in mitxonline learning MFE

Anas12091101:
https://github.com/mitodl/hq/issues/8298

This PR adds configuration to redirect anonymous users to login in the MITXOnline learning MFE. More details in the [ticket](https://github.com/mitodl/hq/issues/8298)

- [ ] Desktop screenshots

### How can this be tested?
- Copy the configurations in your local setup. Make sure to add the `DEPLOYMENT_NAME="mitxonline"` in your env file.
- In edx-platform, change the return to True [here](https://github.com/openedx/edx-platform/blob/master/openedx/core/djangoapps/user_authn/utils.py#L63)
- In a private window, visit a learning MFE URL, and validate that the user is redirected to login with the correct query params.
- Complete the login, the user should be redirected back to the learning MFE page from which the flow was sent to the login page.


src/bridge/settings/openedx/mfe/slot_config/Footer.jsx
asadali145, line 1: Here is how chatgpt refactored it:

```
const ForceLoginRedirect = () => {
  const config = getConfig();
  const { authenticatedUser } = useContext(AppContext);
  console.log(config.APP_ID, process.env.DEPLOYMENT_NAME, authenticatedUser);

  useEffect(() => {
    if (
      config.APP_ID === "learning" &&
      process.env.DEPLOYMENT_NAME?.includes("mitxonline") &&
      authenticatedUser === null
    ) {
      const destination = getLoginRedirectUrl(window.location.href);
      window.location.replace(destination);
    }
  }, [config, authenticatedUser]);

  return null;
};
```

Then we can place it inside `<footer><ForceLoginRedirect/>...ALL OTHER FOOTER CODE...</footer>`

asadali145 (commented):
Works well but can be refactored.

src/bridge/settings/openedx/mfe/slot_config/Footer.jsx
```diff
@@ -110,7 +113,26 @@ const config = {
...
   };
 
+const ForceLoginRedirect = () => {
+  const config = getConfig();
+  const { authenticatedUser } = useContext(AppContext);
+
+  useEffect(() => {
+    if (
+      config.APP_ID === "learning" &&
+      process.env.DEPLOYMENT_NAME?.includes("mitxonline") &&
```
asadali145, line 123: @pdpinch Do you think we should enable this redirect for xPro as well?

asadali145 (approved):
LGTM! Once we get the go-ahead from Peter to make it only for MITx Online.

Issue #3522, opened 2025-08-28
Implement Starburst Roles and Permissions Management with dbt Integration


shaidar:
### Description/Context

We need to implement a comprehensive solution for managing Starburst roles and permissions that integrates Pulumi infrastructure definitions with dbt model grants.

Currently, we don't have a systematic approach for:
- Defining Starburst roles in our infrastructure
- Managing permissions and grants for data models
- Integrating role-based access control between our infrastructure (Pulumi) and data transformation (dbt) layers

### Infrastructure (Pulumi)
- [x] Research and identify Pulumi providers for Starburst/Trino role management
- [x] Define a standardized schema for Starburst roles in Pulumi
- [x] Create reusable Pulumi components for common role patterns (e.g., read-only, analyst, admin)
- [x] Implement role definitions that can be consistently applied across environments

### dbt Integration
- [x] Set up dbt project structure if not already present
- [x] Configure `dbt_project.yml` to support Starburst grant management
- [x] Define grant patterns that reference Pulumi-managed roles
- [x] Create macros/hooks for applying grants to models based on role definitions

### Documentation & Testing
- [x] Document the end-to-end workflow from role definition to grant application
- [x] Create examples showing how models specify grants in dbt
- [ ] Test role and grant application across different environments (dev/qa/prod)
- [ ] Establish governance guidelines for role creation and permission assignment

### Phase 1: Infrastructure Setup
- [x] Investigate Pulumi providers for Starburst (Trino SQL, REST API, or custom)
- [x] Design role definition schema (role names, permissions, catalogs, schemas)
- [x] Implement Pulumi components for role creation and management
- [x] Set up role definitions for common patterns (analyst, engineer, admin, etc.)

### Phase 2: dbt Integration
- [x] Initialize dbt project structure if needed
- [x] Configure dbt profiles and connections for Starburst
- [x] Set up `dbt_project.yml` with grant configuration patterns
- [x] Create dbt macros for consistent grant application
- [x] Implement model-level grant specifications

### Phase 3: Integration & Testing
- [x] Establish workflow for role updates (Pulumi → Starburst → dbt)
- [x] Test grant application on sample models
- [x] Validate permissions work correctly across environments
- [x] Create CI/CD integration for role and grant deployment

shaidar:
# Starburst Galaxy Role and Privilege Management

[...]

PR #3583, opened 2025-09-17
Renaming data-platform roles

shaidar:

Changes inline with what https://github.com/mitodl/ol-infrastructure/issues/3522 introduced

Lining up role names in Keycloak with the roles names that we settled on in Starburst to avoid confusion down the line.

### DIFF
--- a/src/ol_infrastructure/substructure/keycloak/ol_data_platform.py
+++ b/src/ol_infrastructure/substructure/keycloak/ol_data_platform.py
@@ -187,7 +187,7 @@ def create_ol_data_platform_realm(  # noqa: PLR0913
             "maps to superset_admin"
         ),
         composite_roles=[
-            ol_data_platform_superset_client_role_refs["superset_admin"].id
+            ol_data_platform_superset_client_role_refs["ol_platform_admin"].id
         ],
         opts=resource_options,
     )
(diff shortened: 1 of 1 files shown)


blarghmatey (approved):
Might be worth updating the Superset role names in this same PR https://github.com/mitodl/ol-infrastructure/blob/main/src/ol_superset/pythonpath/superset_config.py#L81-L86

src/ol_superset/pythonpath/superset_config.py
```diff
@@ -79,10 +79,10 @@
...
 # https://superset.apache.org/docs/installation/configuring-superset#mapping-ldap-or-oauth-groups-to-superset-roles
 AUTH_ROLES_MAPPING = {
-    "superset_admin": ["Admin"],
-    "superset_researcher": ["Alpha", "sql_lab"],
-    "superset_alpha": ["Alpha"],
-    "superset_gamma": ["Gamma"],
+    "ol_platform_admin": ["Admin"],
+    "ol_researcher": ["Alpha", "sql_lab"],
+    "ol_data_engineer": ["Alpha"],
+    "ol_data_analyst": ["Gamma"],
```
blarghmatey, line 85: I think this also requires an updated role mapping in Keycloak

PR #3585, opened 2025-09-18
chore(deps): upgrade git auto export

asadali145:

None

Upgrades git auto-export for MITx Online


asadali145:
> My only thought on this is that we should first update it in MITx Online, test and see how it goes, and then make the changes for Residential and xPRO.

The reason I upgraded in a single PR is that we won't be doing any xpro/residential releases.

blarghmatey (approved):

PR #3590, opened 2025-09-19
config: remove the course org and number from edx header

Anas12091101:
https://github.com/mitodl/hq/issues/8624

This PR hides the course number and org from the header in MITxOnline.

**Before**

Screenshot 2025-09-19 at 3 49 55 PM

**After**

Screenshot 2025-09-19 at 3 47 40 PM

### How can this be tested?

- Copy-paste the config in your local MFEs env.config.jsx files that use the learning header
- Verify that the course number and org are hidden from the header


asadali145 (changes requested):
The issue is about doing it for UAI courses. We should confirm with @pdpinch before making the change for all of the courses.

asadali145 (approved):

PR #3595, opened 2025-09-19
Md/edxapp eks

Ardiea:

- [ ] Desktop screenshots


src/ol_infrastructure/applications/edxapp/__main__.py
```diff
@@ -1582,6 +1599,17 @@ def cloud_init_user_data_func(
     "fastly_access_logging_iam_role"
 )
 
+if edxapp_config.get("k8s_deployment"):
```
blarghmatey, line 1602: This should probably be `k8s_cutover`

src/ol_infrastructure/applications/edxapp/k8s_configmaps.py
blarghmatey, line 1: It's great that things are working, but this is definitely a prime target for some refactoring/re-architecting. This seems like a lot of maintenance overhead in the medium/long term.

src/ol_infrastructure/applications/edxapp/k8s_resources.py
```diff
@@ -0,0 +1,960 @@
...
+        f"staticfiles-production-{EDXAPP_DOCKER_IMAGE_DIGEST}.tar.gz"
+    )
+    nonprod_staticfiles_archive_name = (
+        f"staticfiles-nonprod-{EDXAPP_DOCKER_IMAGE_DIGEST}.tar.gz"
+    )
+    production_staticfiles_url = f"https://ol-eng-artifacts.s3.amazonaws.com/edx-staticfiles/{stack_info.env_prefix}/{OPENEDX_RELEASE}/{production_staticfiles_archive_name}"
+    nonprod_staticfiles_url = f"https://ol-eng-artifacts.s3.amazonaws.com/edx-staticfiles/{stack_info.env_prefix}/{OPENEDX_RELEASE}/{nonprod_staticfiles_archive_name}"
+
+    prod_dl_command = f"wget {production_staticfiles_url} -O /tmp/prod.tar.gz && tar -xf /tmp/prod.tar.gz --strip-components 2 -C /openedx/staticfiles"
+    nonprod_dl_command = f"wget {nonprod_staticfiles_url} -O /tmp/nonprod.tar.gz && tar -xf /tmp/nonprod.tar.gz --strip-components 2 -C /openedx/staticfiles"
```
blarghmatey, line 238: I would like to get rid of the external staticfiles management as it is no longer necessary with the ProctorTrack integration and it adds a lot of cognitive and build overhead.

blarghmatey (approved):
Overall this is great work. Definitely lots of follow-on work to break it into more manageable chunks, but it will be great to get things running in K8s so that we can iterate from there.

src/ol_infrastructure/applications/edxapp/k8s_resources.py
```diff
@@ -211,6 +272,22 @@ def create_k8s_resources(
         opensearch_hostname=opensearch_hostname,
     )
 
+    openedx_data_pvc = kubernetes.core.v1.PersistentVolumeClaim(
```
blarghmatey, line 275: Is this to avoid having to deal with sticky sessions in the CMS?

src/ol_infrastructure/applications/edxapp/k8s_resources.py
```diff
@@ -404,6 +607,21 @@ def create_k8s_resources(
             mount_path="/openedx/config",
         )
     )
+
+    # Finally, actually define the CMS deployment
+    # Start with a pre-deployment job that runs the migrations
+    cms_pre_deploy_migrate_job = _create_pre_deploy_job(
```
blarghmatey, line 613: We will want to make sure that the LMS migrations _always_ run before the CMS migrations

src/ol_infrastructure/applications/edxapp/k8s_resources.py
Ardiea: Yes

src/ol_infrastructure/applications/edxapp/k8s_resources.py
Ardiea: Ohhhhh, okay I didn't know that. Hm. Okay. I can make that happen I think.

Ardiea:
One thing I thought of this morning is that I need to ensure the environment vars are being populated in the pulumi job of the deployment pipelines. I'm not sure they are / or if they are named correctly. So not quite ready yet.

PR #3640, opened 2025-10-01
Add docs on authoring workflow

dsubak:
Just removes documents that will be moved to the platform eng handbook once https://github.com/mitodl/platform-engineering-site/pull/13 lands


blarghmatey:
Can you move this documentation into https://github.com/mitodl/platform-engineering-site? That will make it easier to share and view.

PR #3645, opened 2025-10-02
Deploy Dagster in Kubernetes

blarghmatey:
N/A

Migrates Dagster deployment to run in Kubernetes instead of Ec2


src/ol_infrastructure/applications/dagster/__main__.py
```diff
@@ -337,288 +352,610 @@
...
-    meta={
-        "external-node": True,
-        "external-probe": True,
+
+# ============================================================================
+# Kubernetes Deployment using Helm
+# ============================================================================
+
+# OLEKSAuthBinding for IRSA and Vault K8s auth
+dagster_auth_binding = OLEKSAuthBinding(
```
Ardiea: 👍

PR #3671, opened 2025-10-09
refactor: learn open edx slots

asadali145:

Might fix https://github.com/mitodl/hq/issues/8852

Refactors MFE slots for https://github.com/mitodl/hq/issues/8852

### How can this be tested?

- Test that the footer loads fine in the MFEs
- Test that the user menu shows the learn dashboard link when looking at a UAI_ course.
- Test that the user menu shows mitxonline links when looking at a non-UAI_ course.
- Test that the course key is not visible above the title in the header for the UAI course and visible for non-UAI course.


Anas12091101 (changes requested):
Works fine 👍  Just a couple of suggestions:

The xPRO and Residential configuration files should also be refactored to ensure consistency across all configurations.

pdpinch:
We closed this issue, but this PR is still open. Should this be closed?

PR #3708, opened 2025-10-15
learn-ai nginx config update 

Ardiea:
fix: updated nginx config for learn-ai to allow streaming of AI responses.

CC @mbertrand


mbertrand (approved):
👍

PR #3712, opened 2025-10-16
fix: Removed binderhub

Ardiea:
Removes binderhub from our juptyerhub deployment. 

Tested in CI with @dsubak's help. Probably should get a more rigorous round of testing.


dsubak (approved):
Looks good to me - left one comment but once that's sorted I think this is looking good!

blarghmatey (approved):
This looks like it will remove and replace the Helm deployment, so I imagine that will lead to some period of downtime. We'll want to coordinate that for production.

blarghmatey:
You'll want to talk to Peter and Simone about it since it will impact the UAI courses.

Issue #3297, opened 2025-06-13
K8S Observability Improvements


Ardiea:
### Description/Context
Grafana has a helm chart that includes basically everything we would need to utilzie their built in "Kubernetes Metrics" tab. It seems like a superset of our current alloy-only configuration so it may be a drop in replacement to switch to this helm chart rather than ours. 

### Plan/Design
https://grafana.com/docs/grafana-cloud/monitor-infrastructure/kubernetes-monitoring/

PR #3735, opened 2025-10-21
Md/grafana refactor

Ardiea:
Closes https://github.com/mitodl/ol-infrastructure/issues/3297 

Removed the standalone grafana alloy installation in favor of 'grafana-k8s-monitoring', which is a more comprehensive suite of monitoring tools for kubernetes clusters and applications running on them. 
Updates references to the tracing endpoints to point to the new one. 
Removes grafana metric collection for vault and consul nodes. Little value and high cardinality. 

Bonus: Refactor the keda related code out of `__main__.py` in the substructure into its own module. (Copilot)

### DIFF
--- a/src/ol_infrastructure/substructure/aws/eks/__main__.py
+++ b/src/ol_infrastructure/substructure/aws/eks/__main__.py
@@ -9,7 +9,7 @@ import pulumi_vault as vault
 from pulumi import Config, ResourceOptions, StackReference, export
 
 from bridge.lib.versions import (
-    KUBE_STATE_METRICS_CHART_VERSION,
+    GRAFANA_K8S_MONITORING_CHART_VERSION,
     VANTAGE_K8S_AGENT_CHART_VERSION,
 )
 from bridge.secrets.sops import read_yaml_secrets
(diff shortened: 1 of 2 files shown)


src/ol_infrastructure/substructure/aws/eks/__main__.py
```diff
@@ -364,440 +359,252 @@
...
+grafana_k8s_monitoring_helm_release = kubernetes.helm.v3.Release(
+    f"{cluster_name}-grafana-k8s-monitoring-helm-release",
     kubernetes.helm.v3.ReleaseArgs(
-        name="grafana-alloy",
-        chart="alloy",
-        version="",
-        namespace="operations",
+        name="grafana-k8s-monitoring",
+        chart="k8s-monitoring",
+        version="3.5.3",
```
blarghmatey: Would be good to put this in versions.py and use Renovate for managing it

blarghmatey (approved):
Do we also want to remove our Prometheus CRDs? It's a bit unclear if this chart handles that as well or not.

Ardiea:
I think we want to leave those. The helm chart does not include them.

PR #3760, opened 2025-10-27
config: persist the required disabled legacy_studio flags (MITx Online/Learn)

arslanashraf7:
https://github.com/mitodl/hq/issues/9012

These legacy_studio flags keep getting activated after each deployment. We need them disabled for RC except for video_editor, problem_editor and file uploads https://github.com/mitodl/hq/issues/9012#issuecomment-3448965739

- [ ] Desktop screenshots


arslanashraf7:
> Can you also make this change for production, so that we don't forget about it before the release?

@blarghmatey I've updated them in the production file as well. I initially deleted these but then decided to deactivate instead, let me know which one is better in this case.

cc: @pdpinch

PR #3789, opened 2025-11-04
chore: bump git-auto-export and related configs added

marslanabdulrauf:
https://github.com/mitodl/hq/issues/6962

This PR:
1. Bumps the `ol-openedx-git-auto-export` version to 0.7.0
2. Adds configuration required for this new version


src/bilder/images/edxapp_v2/templates/edxapp/mitxonline/cms_only.yml.tmpl
```diff
@@ -3,6 +3,8 @@
 COURSE_AUTHORING_MICROFRONTEND_URL: /authoring
 DISCUSSIONS_INCONTEXT_LEARNMORE_URL: https://openedx.atlassian.net/wiki/spaces/COMM/pages/3470655498/Discussions+upgrade+Sidebar+and+new+topic+structure
 GIT_REPO_EXPORT_DIR: /openedx/data/export_course_repos
+GITHUB_ORG_API_URL: https://github.mit.edu/api/v3/orgs/mitx
```
marslanabdulrauf: @pdpinch could you please confirm the organization where we need to create repositories and upload course content for MITxOnline ?

marslanabdulrauf: `git-auto-export` works on the Github url we add and that url could be for any organization or individual user.
Its communication with Github is handled via either adding `username:password` in repo url or by using SSH (which is the current case)

This repository creation is handled via APIs and Github Access Token

Ardiea (approved):

PR #3793, opened 2025-11-06
Authoring Jupyterhub deployment

dsubak:
Stands up a second Jupyterhub deployment to enable an authoring workflow. This deployment uses a different logical database, but shares the same physical database as the initial deployment. As a result, this provisions a second of everything in the current deployment excluding the following components:
- RDS instance
- DB Security group
- S3 bucket

A large number of the configurables are now set in the stack config under `jupyterhub:deployments`. Previously, as we only had one deployment, many of these values were specified inline.

The functional configuration differences between the regular deployment are as follows:
- All menu items are unsupressed
- All extensions are enabled
- We do not prepull any built images
- We only provision a single user_placeholder_replicas

### How can this be tested?
As of the time of writing, this is up and running in CI under `authoring.nb.ci.learn.mit.edu`. In the event it has been taken down (manually or by concourse) below is a video of it loading the default image.

https://github.com/user-attachments/assets/d09f55f2-3c62-42f6-8e76-ed202fec9eb2

### Release Process
There are a few manual steps required to stand this up in an environment.
- Firstly, the operator will need to manually create a new database called `jupyterhub_authoring` in the environment's existing RDS instance. See [this link](https://mitodl.slack.com/archives/C09DRCSS9S6/p1764702322516779) for details on how this was done in CI.
- Run `pulumi up` for the relevant stack
- Check the environment's ApisixTLS object for the `jupyterhub-authoring` k8s namespace. If the "Accepted" condition is not listed as true and your stack otherwise came up clean, delete the `jupyterhub-authoring-cert` ApisixTLS object via kubectl. This is due to a race condition between the creation of the `api-jupyterhub-authoring-tls-pair` secret and the ApisixTLS object.
- Once `jupyterhub-authoring-cert` has been deleted, run another `pulumi up` to recreate it. At this point you should be able to access the new domain using HTTPS

### Open Questions
- How do we lock this down even further? At the moment all users who can access the normal notebook env can access this as well - we don't need to solve this immediately, but we probably need to before installing the authoring plugin which can write to S3.
- Vault HCL: Is there any concern that the Vault HCL allows authoring and standard deployments to access each others secrets? It's parameterized in case we ever decide we do want to split it on a per-deployment basis within a single stack.
- Is the replacement of `ol-k8s-apisix-olapisixoidcresources-ci` an online operation or will that result in downtime?

### Resolved Questions
- ~Database: We could probably use the same one and just do two logical databases if we want. I can't see an obvious reason it wouldn't work - question is if we'd prefer hardware-level separation or not.~ We're going to cram everything into one physical database.
- ~Namespace: ATM I'm assuming we'll want this in its own namespace. This is because [resources within a namespace need to be unique](https://kubernetes.io/docs/concepts/overview/working-with-objects/namespaces/#when-to-use-multiple-namespaces), and I'm not sure I can easily control the names of the deployments (i.e. `hub`)~ We've got it in a jupyter-authoring namespace


dsubak:
@blarghmatey or @Ardiea - this still needs work, but I'd appreciate if ya'll could take a look. Of note, I'd love your opinions on the open questions listed in the description; guessing I'll need another pass once I've got some of those answers!

src/ol_infrastructure/applications/jupyterhub/__main__.py
```diff
@@ -137,396 +165,63 @@
...
+    vault_config=vault_config,
+    db_config=jupyterhub_authoring_db_config,
+    app_db=jupyterhub_db,
+    cluster_stack=cluster_stack,
+    application_labels=application_labels,
+    k8s_global_labels=k8s_global_labels,
+    extra_images=EXTRA_IMAGES,
+    menu_override_json=author_menu_override,
+    disabled_extensions_json=author_disabled_extensions,
+    extra_config=dynamic_image_config,
```
blarghmatey: This block would be cleaner if it was more driven by the stack config using a structure along the lines of:
```yaml
jupyterhub:deployments:
  - name: authoring
    domain: authoring.nb.learn.mit.edu
    namespace: jupyterhub-authoring
    db_name: authoring
    additional_labels:
        ol.mit.edu/audience: course-authors
  - name: default
    domain: nb.learn.mit.edu
    namespace: jupyterhub
    db_name: jupyterhub
    additional_labels:
        ol.mit.edu/audience: learners
```
It would also make sense to include key/value for relative file paths to add the the extraConfig, etc.

This makes this deployment code more flexible for being able to manage different JupyterHub deployments across different stacks/clusters without hardcoding as much of it to the authoring/learning use case specifically.

src/ol_infrastructure/applications/jupyterhub/deployment.py
```diff
@@ -0,0 +1,431 @@
...
+    # JupyterHub Helm Release
+    extra_images_list = extra_images or {}
+    admin_users_list = jupyterhub_config.get_object("admin_users", default=[])
+    allowed_users_list = jupyterhub_config.get_object("allowed_users", default=[])
+    return kubernetes.helm.v3.Release(
+        f"{base_name}-{env_name.upper()}-application-helm-release",
+        kubernetes.helm.v3.ReleaseArgs(
+            chart="jupyterhub",
+            cleanup_on_fail=True,
+            version="4.2.0",
```
blarghmatey: We should pull this value from `src/bridge/lib/versions.py` so that it's managed by renovate

src/ol_infrastructure/applications/jupyterhub/__main__.py
blarghmatey: For values that we want to load into the Helm deployment, we can also maintain those as separate JSON or Python dictionaries that get deep-merged into a default values structure so that we don't have to stuff as much of it into the stack config and explode the amount of YAML being written. Then just reference the `values.py` file to be loaded in the config block.

src/ol_infrastructure/applications/jupyterhub/deployment.py
```diff
@@ -0,0 +1,436 @@
...
+        stack_info: Stack information from parse_stack()
+        jupyterhub_deployment_config: Pulumi config for jupyterhub deployment
+        vault_config: Pulumi config for vault
+        cluster_stack: EKS cluster stack reference
+        application_labels: Labels to apply to Kubernetes resources
+        k8s_global_labels: Global Kubernetes labels
+        extra_images: Dictionary mapping image names to image configuration dictionaries (optional)
+        menu_override_json: JSON contents for menu override Jupyter config
+        disabled_extensions_json: JSON contents for disabled extensions Jupyter config
+        extra_config: Extra configuration values to merge into the Helm chart values
```
Ardiea: This docstring isn't grounded in reality. I'd just delete it or have an AI fix it. IMO the args are descriptive enough on their own.

src/ol_infrastructure/applications/jupyterhub/deployment.py
```diff
@@ -0,0 +1,436 @@
...
+                        "requests": {
+                            "cpu": "10m",
+                            "memory": "128Mi",
+                        },
+                        "limits": {
+                            "cpu": "100m",
+                            "memory": "512Mi",
+                        },
+                    },
+                },
```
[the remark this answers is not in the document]
dsubak: @Ardiea I put https://github.com/mitodl/ol-infrastructure/pull/3793/commits/a5f7a504d8931eb05010f46069515efc586380b6 in, but I'm thinking that it probably makes more sense to enable/disable based on passing in an extra_images list. It's not used for anything outside of configuring the prepuller block at the moment, so it being non-empty is as good as a flag. Let me know if ya disagree!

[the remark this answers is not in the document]
blarghmatey: I agree that making it conditional on the truthy/falsey evaluation of the list is the correct way to do it.

src/ol_infrastructure/applications/jupyterhub/__main__.py
```diff
@@ -137,396 +165,63 @@
...
-    jupyterhub_db_vault_backend_config, opts=ResourceOptions(depends_on=[jupyterhub_db])
-)
-jupyterhub_creds_secret_name = "jupyterhub-db-creds"  # noqa: S105  # pragma: allowlist secret
-
-# Create vault_k8s_resources to allow jupyter hub to access secrets in vault
-jupyterhub_vault_policy = vault.Policy(
-    f"ol-jupyterhub-vault-policy-{stack_info.env_suffix}",
-    name="jupyterhub",
-    policy=Path(__file__).parent.joinpath("jupyterhub_policy.hcl").read_text(),
+jupyterhub_authoring_db_config = OLPostgresDBConfig(
```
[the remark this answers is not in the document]
dsubak: Bit squirrely, but we got that done - involved making a different secrets engine mount for jupyter authoring because `OLVaultDatabaseBackend` encapsulates the roles, connection and mount itself. 

Technically we could've done a single mount with 2 connections (one per DB) and an additional role, but we'd have needed to drop below the ComponentResource level and work w/ raw resources. Since that'd be more work for little benefit, this'll play for now!

Resolving this!

src/ol_infrastructure/applications/jupyterhub/Pulumi.applications.jupyterhub.CI.yaml
```diff
@@ -2,17 +2,33 @@
...
+  - name: jupyterhub-authoring
+    domain: authoring.nb.ci.learn.mit.edu
+    namespace: jupyter-authoring
+    db_name: jupyterhub_authoring
+    apisix_ingress_class: apache-apisix
+    admin_users:
+    - "admin"
+    menu_override_file: author_menu_override.json
+    disabled_extension_file: author_disabled_extensions.json
+    extra_config_file: dynamicImageConfig.py
```
dsubak: Note that I'll need to do this for all other envs too. Not hard, just something to do once it's functionally complete!

src/ol_infrastructure/applications/jupyterhub/jupyterhub_policy.hcl
```diff
@@ -9,6 +9,10 @@ path "postgres-jupyterhub/creds/app" {
   capabilities = ["read"]
 }
 
+path "postgres-jupyterhub_authoring/creds/app" {
```
dsubak: Could easily split these out if we'd like. No reason to keep em all in one spot.

PR #3823, opened 2025-11-14
Create B2B partner storage S3 / SFTP

feoh:
https://github.com/mitodl/hq/issues/6965

This PR creates necessary storage for our UAI partners.

This storage:
- Offers S3 access with appropriate cross account IAM support so partners can access ONLY their prefix
- SFTP access using a publish ssh key supplied by the partner

### How can this be tested?
- [X] Test that partners can ONLY access s3 bucket context within their assigned prefix
- [X] SFTP to generated server using partner ssh public keys configured in the Pulumi project YAML.
- [x]  Ensure that the SFTP server actually stores content in the partner prefix

### DIFF
--- a/docs/b2b_partners_storage_partner_setup.md
+++ b/docs/b2b_partners_storage_partner_setup.md
@@ -1,8 +1,8 @@
-# UAI Partners Storage - Partner Account Setup Guide
+# B2B Partners Storage - Partner Account Setup Guide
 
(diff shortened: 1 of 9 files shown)


blarghmatey (approved):
Looks great! Just replace usage of `uai` with `b2b` because the UAI offering is just the first step of the broader b2b objective.

PR #3976, opened 2025-12-16
Refactor edX configs to use dictionaries and reduce duplication

blarghmatey:
N/A

- Migrate YAML string templates to use Python dictionaries to make them easier to compose together and customize per-deployment.
- Consolidate duplicated values into a single reference with per-deployment overrides to make it easier to see what is the same and what differs
- Moves secret templates to use dictionary -> yaml generation for simpler modification

### How can this be tested?

Run a pulumi up manually and review the secret and configmap changes to make sure that they make sense

### DIFF
--- a/src/ol_infrastructure/applications/edxapp/config_builder.py
+++ b/src/ol_infrastructure/applications/edxapp/config_builder.py
@@ -394,7 +394,6 @@ def get_deployment_overrides(env_prefix: str) -> ConfigDict:
         "PLATFORM_DESCRIPTION": "MITx Residential",
         "PRESS_EMAIL": "support@mitx.mit.edu",
         "MITX_REDIRECT_ENABLED": False,
-        "SYSADMIN_DEFAULT_BRANCH": "live",
         "EMAIL_HOST": "outgoing.mit.edu",
         "EMAIL_PORT": 587,
         "EMAIL_USE_TLS": True,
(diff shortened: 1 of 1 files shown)


src/ol_infrastructure/applications/edxapp/secrets_builder.py
```diff
@@ -0,0 +1,419 @@
...
+        },
+        "default": {**redis_cache_base, "KEY_PREFIX": "default", "VERSION": "1"},
+        "general": {**redis_cache_base, "KEY_PREFIX": "general"},
+        "mongo_metadata_inheritance": {
+            **redis_cache_base,
+            "KEY_PREFIX": "mongo_metadata_inheritance",
+            "TIMEOUT": 300,
+        },
+        "staticfiles": {**redis_cache_base, "KEY_PREFIX": "staticfiles"},
+    }
```
Ardiea: This is the CACHES to keep.

src/ol_infrastructure/applications/edxapp/secrets_builder.py
```diff
@@ -0,0 +1,419 @@
...
+        "USER": '{{ get .Secrets "username" }}',
+        "PASSWORD": '{{ get .Secrets "password" }}',
+    }
+
+    # Common database options
+    db_options = {
+        "ssl_mode": "REQUIRED",
+        "ssl": {"cipher": "TLSv1.2"},
+        "charset": "utf8mb4",
+    }
```
Ardiea: Interestingly, whatever is rendering this seems to create a yaml anchor for this. Not a concern just something I noticed. It looks right to me.

PR #3979, opened 2025-12-16
Front MIT Learn's media storage bucket with Fastly CDN

shaidar:

https://github.com/mitodl/hq/issues/9376

Front MIT Learn's media storage bucket with Fastly CDN


blarghmatey:
Functionally this looks good. I have some questions for @ChristopherChudzicki or anyone else working on Learn about which path prefixes or file types we should be keying off of to determine which requests to route to the bucket. I don't think `/media` is sufficient for this case.

PR #4022, opened 2025-12-29
chore: remove ulmo deprecated certificate waffle flag

arslanashraf7:
None, noticed it while working on https://github.com/mitodl/hq/issues/9693#issuecomment-3684103207

- This setting is deprecated as of Ulmo [As per this info](https://github.com/openedx/edx-platform/blob/master/cms/djangoapps/contentstore/toggles.py#L356) so we should stop managing it as well.
- I would rather delete this waffle flag but I don't think that can be done via infra, and we may have to do it manually?

- [ ] Desktop screenshots

### How can this be tested?
Once deployed, everything around certificates look/feel, and settings should stay the same.

### DIFF
--- a/src/ol_infrastructure/applications/edxapp/Pulumi.applications.edxapp.mitx-staging.CI.yaml
+++ b/src/ol_infrastructure/applications/edxapp/Pulumi.applications.edxapp.mitx-staging.CI.yaml
@@ -35,7 +35,6 @@ config:
     "--everyone"]
   - ["openresponseassessment.enhanced_staff_grader", "--create", "--superusers", "--everyone"]
   - ["studio.enable_checklists_quality", "--create", "--superusers", "--everyone"]
-  - ["legacy_studio.certificates", "--create", "--everyone"]
   edxapp:business_unit: residential-staging
   edxapp:canvas_base_url: https://mit.test.instructure.com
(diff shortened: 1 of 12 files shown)


pdpinch:
Merging this PR is a good plan, but perhaps we should delete the waffle flag _first_ in RC and confirm that it's harmless. After that, we can delete it in production as well. And then merge this so it doesn't come back again in future deployments.

arslanashraf7:
> Merging this PR is a good plan, but perhaps we should delete the waffle flag _first_ in RC and confirm that it's harmless. After that, we can delete it in production as well. And then merge this so it doesn't come back again in future deployments.

Sure. I've removed the flag present at https://courses.rc.learn.mit.edu/admin/waffle/flag/36/change/. For posterity, this was the flag state before removing:
image

arslanashraf7:
> Ok. What should we do? Do we need an upstream PR to fix the behavior of `ENABLE_CERTIFICATE_PAGE` ?

Yes. Although we need to figure out how these three flags should work together:
1. legacy_studio.certificates (Platform)
2. FEATURES["CERTIFICATES_HTML_VIEW"] (Platform)
3. ENABLE_CERTIFICATE_PAGE (MFE)

PR #4049, opened 2026-01-09
fix: use same domain to manage authoring MFE reloads

asadali145:

None

Fixes the authoring MFE reload logic. We used a cookie to avoid multiple reloads but due to the studio domain,  learning MFE was not clearing it when updating the language. This PR fixes that stuff.

### How can this be tested?

- Setup auto language selection by following the readme https://github.com/mitodl/open-edx-plugins/tree/main/src/ol_openedx_course_translations
- Test that language changes for the learning MFE based on the course language and resets to English for the Authoring MFE.

### DIFF
--- a/src/bridge/settings/openedx/mfe/slot_config/Footer.jsx
+++ b/src/bridge/settings/openedx/mfe/slot_config/Footer.jsx
@@ -161,30 +161,49 @@ const AutoSelectLanguage = () => {
     const match = document.cookie.match(new RegExp(`(^| )${safeName}=([^;]+)`));
     return match ? decodeURIComponent(match[2]) : null;
   };
-  const setCookie = (name, value, days = 1, domain = null) => {
+  const setCookie = (name, value, days = 1, domainAttr = null) => {
     const expires = new Date(Date.now() + days * 864e5).toUTCString();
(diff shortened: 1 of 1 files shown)


src/bridge/settings/openedx/mfe/slot_config/Footer.jsx
```diff
@@ -161,12 +161,31 @@ const AutoSelectLanguage = () => {
...
+
+  // Helper to extract hostname from URL
+  const getHostnameFromUrl = (url) => {
+    // Remove subdomain 'studio.' if present e.g., studio.courses.learn.mit.edu -> courses.learn.mit.edu
+    try {
+      const hostname = new URL(url).hostname;
+      if (hostname.startsWith("studio.")) {
+        return hostname.replace(/^studio\./, "");
+      }
+      return hostname;
```
arslanashraf7: Should we make it dynamic so that it doesn't break when this URL changes in the future, e.g, let's say we make this URL cms.courses.learn.mit.edu, it will break. What we should do instead is to compare the URL with lmsBaseUrl and remove any additional part. I know we may still face issues, but keeping it dynamic this way is still more robust.

PR #4072, opened 2026-01-16
chore(deps): pin openai for ol-openedx-course-translations

asadali145:

https://github.com/mitodl/hq/issues/9913

Pins openai dependency as it is not used in edx-platform and causing troubles with course translations plugin. Here is the edx-platform issue to remove the dependency https://github.com/openedx/edx-platform/issues/35268


arslanashraf7 (approved):
Looks good.
We should probably go with `openai>=2.15.0` but It's fine as is too because of the below comment:

Can you please add a link to the ticket in edx-platform to remove the `OpenAI` package? And also keep a track of it so that we revert this PR once openai package is removed from Open edX.

blarghmatey (approved):

Issue #4069, opened 2026-01-15
Populate K8s label objects for all our applications


feoh:
### Description/Context
using the fully populated application label list defined in #3973, actually create pull requests per application that define the appropriate filled in kubernetes labels.

feoh:
Structural stub for hints from Tobias - [branch](https://github.com/mitodl/ol-infrastructure/tree/k8s_labels_hierarchical)

PR #4120, opened 2026-01-26
Cpatti/k8s labels ontology

feoh:
Related to https://github.com/mitodl/ol-infrastructure/issues/4069

This change is a step in the direction of creating Application / Service / Compoenent objects that translate to K8s labels in our running/deployed apps.
 
### How can this be tested?
 pulumi up. Check labels through Headlamp.

### DIFF
--- a/src/ol_infrastructure/applications/keycloak/__main__.py
+++ b/src/ol_infrastructure/applications/keycloak/__main__.py
@@ -46,7 +46,7 @@ from ol_infrastructure.lib.ol_types import (
     Application,
     AWSBase,
     BusinessUnit,
-    K8sGlobalLabels,
+    K8sAppLabels,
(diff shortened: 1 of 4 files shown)


src/ol_infrastructure/applications/mit_learn_nextjs/__main__.py
```diff
@@ -161,7 +172,8 @@ def determine_colors(last_active_color: str) -> dict[str, str]:
     )
 
 application_labels = k8s_global_labels | {
-    "ol.mit.edu/service": "nextjs",
+    "ol.mit.edu/application": Application.mit_learn,
```
blarghmatey: Another case where we can remove this in favor of the label already present in the Pydantic object.

PR #4161, opened 2026-02-03
fix: remove OCW_EXTRA_COURSE_THEMES from prod

zawan-ila:
Related to https://github.com/mitodl/hq/issues/9941

This PR unsets the variable `OCW_EXTRA_COURSE_THEMES` for production, so that it is properly [interpreted](https://github.com/mitodl/ocw-studio/blob/master/main/settings.py#L1271-L1276) as an empty list.

Setting an env var to an empty string and then interpreting it as a list using `get_delimited_list` will end up setting it to a list with an empty string `['']` whereas the original intention is more likely to be to set it to an empty list `[]`. This is a somewhat unintuitive behavior of the `get_delimited_list` function but may not be easy to fix due to compatibility concerns with other apps.

### How can this be tested?
After this is merged, verify the values of OCW_EXTRA_COURSE_THEMES from RC and prod django shell. On prod, it should be `[]`. On RC `['ocw-course-v3']`


feoh:
@zawan-ila Rebase to fix this pre-commit issue and I'll approve as I see no devops implications for this theming change.

PR #4168, opened 2026-02-05
Add audit logging to OL S3 Component and convert OCW Site project to use it

feoh:

Fixes https://github.com/mitodl/hq/issues/9680

Add audit logging to OL S3 Component and convert OCW Site project to use it

### How can this be tested?
Run pulumi up in QA and ensure that audit logs are appearing in the newly created audit logging bucket.

### DIFF
--- a/src/ol_infrastructure/applications/ocw_site/__main__.py
+++ b/src/ol_infrastructure/applications/ocw_site/__main__.py
@@ -465,6 +465,7 @@ draft_offline_bucket_config = S3BucketConfig(
         )
     ],
     website_configuration=s3.BucketWebsiteConfigurationArgs(
+        bucket=draft_offline_bucket_name,
         index_document=s3.BucketWebsiteConfigurationIndexDocumentArgs(
             suffix="index.html",
         ),


src/ol_infrastructure/applications/ocw_site/__main__.py
```diff
@@ -478,66 +501,67 @@
...
 )
 
-live_offline_bucket_cors = s3.BucketCorsConfiguration(
-    "ol-live-offline-bucket-cors",
-    bucket=live_offline_bucket_name,
+# live_offline bucket
+live_offline_bucket_config = S3BucketConfig(
+    bucket_name=live_offline_bucket_name,
+    versioning_enabled=True,
+    ownership_controls="BucketOwnerPreferred",
```
blarghmatey: Similarly for the live bucket, let's do the draft and test ones first before we update the live offline bucket.

blarghmatey (changes requested):
Overall this looks great. Let's just remove the live bucket modifications until we can prove this out on the draft buckets.

src/ol_infrastructure/applications/ocw_site/__main__.py
```diff
@@ -341,11 +366,35 @@
...
     ),
+    logging_target_bucket=audit_log_bucket_name,
+    logging_target_prefix=f"ocw-site/{draft_backup_bucket_name}/",
+    logging_target_object_key_format=s3.BucketLoggingTargetObjectKeyFormatArgs(
+        partitioned_prefix=s3.BucketLoggingTargetObjectKeyFormatPartitionedPrefixArgs(
+            partition_date_source="EventTime"
+        )
+    ),
+    logging_expected_bucket_owner=aws_account.account_id,
+    tags=aws_config.tags,
```
blarghmatey: I don't think we need to care about logging on the backup buckets. The only thing that would be modifying that is the backup pipeline or someone trying to do a restore.

PR #4186, opened 2026-02-11
fix: git auto export partial configs added for xpro and residential

marslanabdulrauf:
https://github.com/mitodl/hq/issues/10161#issuecomment-3879324760

This PR adds `git-auto-export` configuration for xPro and MITx - only for QA/CI environment

### Additional Context
1. We need to add `Github Access Token` through vault
2. We need to add SSH key for these 2 applications and make sure they have required rights e.g; clone/pull/push
3. Finally we need to run management command for these environments to create respective repositories in the respective organizations

```python
python manage.py migrate_giturl --all
```

**NOTE:** This Command might take time depending upon how many courses we have

### DIFF
--- a/src/bilder/images/edxapp_v2/templates/edxapp/mitx-staging/cms_only.yml.tmpl
+++ b/src/bilder/images/edxapp_v2/templates/edxapp/mitx-staging/cms_only.yml.tmpl
@@ -1,7 +1,6 @@
 # -*- mode: yaml -*-
 # ALTERNATE_WORKER_QUEUES: lms  # Already has a sane default in the code
 COURSE_AUTHORING_MICROFRONTEND_URL: /authoring
-GITHUB_ORG_API_URL: {{ key "edxapp/github-auto-export-org-api-url" }}
 GIT_REPO_EXPORT_DIR: /openedx/data/export_course_repos
 GIT_EXPORT_DEFAULT_IDENT:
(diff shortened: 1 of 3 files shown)


src/ol_infrastructure/applications/edxapp/k8s_configmaps.py
```diff
@@ -452,10 +452,9 @@ def create_k8s_configmaps(  # noqa: PLR0915
         "SESSION_COOKIE_NAME": f"{env_name}-edx-studio-sessionid",
     }
 
-    if stack_info.env_prefix == "mitxonline":
```
marslanabdulrauf: Removed if check as we are enabling it for all apps now

src/ol_infrastructure/applications/edxapp/Pulumi.applications.edxapp.mitx-staging.CI.yaml
```diff
@@ -60,6 +60,7 @@ config:
     learner_dashboard: dashboard
     learning: learn
     ora_grading: ora-grading
+  edxapp:github_org_api_url: https://github.mit.edu/api/v3/orgs/{actual_org}
```
marslanabdulrauf: @pdpinch I don't have access to all the organizations, could you please add organization for MITx-staging here ?

PR #4495, opened 2026-04-14
feat(local-dev): holistic local development environment (k3d + Tilt + Pulumi)

blarghmatey:
N/A — tracks RFC discussion at https://github.com/mitodl/hq/discussions/10833

Introduces a fully self-contained local development environment for the MIT Learn application stack, running inside a k3d Kubernetes cluster with Tilt for live development and Pulumi for shared infrastructure.

**Apps supported:** mit-learn (Django + Next.js), learn-ai, mitxonline, odl-video-service

**What's included:**

- `local-dev/cluster/k3d-config.yaml` — k3d cluster definition (Traefik disabled, local registry on port 5000)
- `local-dev/scripts/setup.sh` — one-time bootstrap: validates prereqs, creates cluster, generates mkcert wildcard TLS certs, patches /etc/hosts
- `local-dev/scripts/teardown.sh` — cluster teardown with --keep-certs / --keep-hosts flags
- `local-dev/scripts/seed.sh` — CLI wrapper for running management commands inside pods
- `local-dev/infra/` — Pulumi stack that owns all shared in-cluster resources: CNPG (PostgreSQL), Redis, Qdrant, LiteLLM, Mailpit, APISIX ingress, official Keycloak Operator, and the `olapps` Keycloak realm (mirrored from production `olapps.py`)
- `local-dev/apps/{app}/` — hand-authored Kubernetes manifests for each app (Deployments, Services, ConfigMaps, Secrets, ApisixRoute CRDs)
- `Tiltfile` — root entry point with per-app includes, live-update rules, pre-built image fallback, and manual seed triggers
- `local-dev/README.md` — comprehensive documentation

**Design principles:**
- `setup.sh` owns only the cluster, certs, and /etc/hosts. Everything in-cluster is owned by Pulumi or Tilt — no mixing.
- Hostnames mirror production: `.edu` → `.dev` (e.g. `api.learn.mit.edu` → `api.learn.mit.dev`)
- Live source sync for Django apps (sub-second, no rebuild). Pre-built ECR image fallback when a repo isn't checked out.
- Keycloak SSO at `sso.ol.mit.dev` with `olapps` realm, fake-Touchstone IdP, and test users.

### How can this be tested?

Prerequisites: Docker Desktop (≥8 GB RAM), k3d, Tilt, Helm, mkcert, Pulumi CLI.

```bash
# Clone ol-infrastructure and check out the branch
git checkout feature/local-dev-consolidation

# Bootstrap the cluster
./local-dev/scripts/setup.sh

# Configure
cp tilt_config.json.example tilt_config.json

# Install Pulumi deps
cd local-dev/infra && pip install -r requirements.txt && cd ../..

# Start
tilt up
```

Then open https://learn.mit.dev (frontend) and https://api.learn.mit.dev/api/v0/ (backend).

Full instructions in `local-dev/README.md`.

### Additional Context
- Follows Path A from the RFC (hand-authored manifests derived from QA Pulumi configs). Path B (per-app Pulumi dev stacks using `OLApplicationK8s`) is a future phase.
- Phase 6A (per-app database isolation toggle) is defined but not yet implemented.
- The Keycloak Operator used here is the official upstream operator (same as production), not Bitnami.


daniellefrappier18:
@blarghmatey 
Tested on: macOS, Docker Desktop with 8 CPUs / 11 GB RAM allocated, all prerequisites installed.

Setup steps followed: setup.sh → pulumi install → cp tilt_config.json.example tilt_config.json → tilt up

Blocker: local-infra cannot complete due to Keycloak instability

The local-infra Tilt resource consistently fails because Keycloak enters a crash loop before Pulumi can finish configuring it. The pod repeatedly restarts (6+ times) with exit code 1, with no error in the logs. It starts successfully then crashes shortly after.

The failure pattern in Pulumi is:

error sending GET/POST request to https://sso.ol.mit.dev/...: 502 Bad Gateway
APISIX returns 502 because Keycloak is mid-restart when pulumi up --refresh runs. The --refresh flag causes Pulumi to fire ~40 concurrent API calls to Keycloak immediately, which appears to destabilize it before it has fully warmed up.

**Things I tried:**
- Increased Docker memory from 7.5 GB → 11 GB
- Patched Keycloak liveness/readiness probe timeouts from 1s → 10s
- Waited for kubectl wait pod/keycloak-0 --for=condition=Ready before retrying
- Ran pulumi cancel to clear stale locks and retried
- None resolved the crash loop reliably. The root issue seems to be that there's no readiness gate between Keycloak becoming Ready (health probe on port 9000 passes) and Pulumi configuring the realm (requires the HTTP service on port 8080 to be fully warmed up and able to handle concurrent load).

**Minor setup issues also found:**
- uv is not listed as a prerequisite but is required (Pulumi uses it to manage Python deps for the infra stack)
- Port 5000 conflicts with macOS AirPlay Receiver. It's worth noting in the prerequisites or README troubleshooting section that AirPlay Receiver needs to be turned off to free up that port. Or maybe just hardcode it to port 5001 instead.

jkachel:
Tested this on a Debian machine. I ran into a few hurdles with setup but was able to get the stack up and mostly running. (Setup stumbles include the `uv` dependency - I am testing on a fresh Debian install and it wasn't there already - and some issues from not reading closely enough.)

Did notice a couple of things once I got things going:

- The `mitlearn-nextjs` resource failed to build successfully a few times because there weren't enough inotify resources). It gives you a link to the Kind docs that explain this but probably easier to note it on the frontend.

- Once I fixed that, I noticed that the `mitlearn-nextjs` resource doesn't build because it's missing (at least) one `env` variable:
```
Error [ValidationError]: NEXT_PUBLIC_CSRF_COOKIE_NAME is a required field
    at ignore-listed frames {
  value: [Object],
  path: 'NEXT_PUBLIC_CSRF_COOKIE_NAME',
  type: 'required',
  params: [Object],
  errors: [Array],
  inner: []
}
```

From a bit of looking I'm not entirely sure that the environment is being pulled - it seems like it ought to come from the `mit-learn` namespace but it doesn't appear to be?

- I was able to get MITx Online to come up and run and the Learn Django app appears to be working OK. MITx Online seems to want to use its own auth flow though, which is a bit weird. I was able to auth into Keycloak by manually going to `mitxonline.mit.dev/login/`, but I got a 502 on the way back and was stuck. I am able to get into the Keycloak admin. 

- I also noticed that `odlvideo-webapp` fails to build too (it says it's missing `nplusone` which seems like there's some -sort of packaging issue there). I don't really know anything about OVS, though.

shanbady:
Tested this on a mac. some notes:

- I did not get a successful stack running. I followed the instructions and upon doing ```tilt up``` did not see any output and so tore everything down. subsequent attempts at restarting everything errors out and i see this in logs:
```
c… │ ERROR: Build Failed: namespaces "mitxonline" not found
odlvideo-web… │      Tried to apply objects to cluster:
odlvideo-web… │        → odlvideo-webapp:service
odlvideo-web… │        → odlvideo-webapp:deployment
odlvideo-web… │ 
odlvideo-web… │ ERROR: Build Failed: namespaces "odl-video-service" not found
```
- small annoyance - having to manually install deps one by one on my host machine: k3d, tilt, helm, mkcert, and pulumi. it would be nice to have the setup check and install these for me if possible
- instructions ```cd local-dev/infra && pip install -r requirements.txt && cd ../.```.  ——> update instructions - should be uv sync 
- I think it would be useful to have ```tilt up```  be run with ```--stream=true``` so logs are visible so I know something is happening

quazi-h:
I was able to get the Tilt UI running with a few minor issues. I checked the other reviews and can echo some of the same things - had to install a few of the prerequisites that I hadn't used before (tilt, k3d, tilt, and mkcert). I also ran into the macOS AirPlay receiver port conflict with 5000. I just disabled the AirPlay receiver on my machine since I never use it anyways.

I saw a few build failures that were showing as "Update error" with the following traceback: `namespaces "learn-ai" not found`

zawan-ila:
I ran into the same errors as already faced by Danielle and Shankar above. However, after a few tries at tearing everything down and restarting from scratch, I was able to get past those.

The local-infra resource is still erroring out though

```
Diagnostics:
  kubernetes:apps/v1:Deployment (litellm):
    error: 3 errors occurred:
    	* resource "urn:pulumi:local-dev.infra.Dev::ol-local-dev-infra::kubernetes:core/v1:Namespace$kubernetes:apps/v1:Deployment::litellm" was successfully created, but the Kubernetes API server reported that it failed to fully initialize or become live: 'litellm' timed out waiting to be Ready
    	* Minimum number of live Pods was not attained
    	* [Pod local-infra/litellm-859ddfc9cd-475nx]: containers with unready status: [litellm]

  kubernetes:helm.sh/v3:Release (opensearch):
    warning: Helm release "opensearch" was created but has a failed status. Use the `helm` command to investigate the error, correct it, then retry. Reason: context deadline exceeded
    error: 1 error occurred:
    	* Helm release "local-infra/opensearch" was created, but failed to initialize completely. Use Helm CLI to investigate: failed to become available within allocated timeout. Error: Helm Release local-infra/opensearch: context deadline exceeded

  kubernetes:apps/v1:Deployment (ollama):
    error: 3 errors occurred:
    	* resource "urn:pulumi:local-dev.infra.Dev::ol-local-dev-infra::kubernetes:core/v1:Namespace$kubernetes:apps/v1:Deployment::ollama" was successfully created, but the Kubernetes API server reported that it failed to fully initialize or become live: 'ollama' timed out waiting to be Ready
    	* Minimum number of live Pods was not attained
    	* [Pod local-infra/ollama-b4fb75c47-bf79k]: containers with unready status: [ollama]

  kubernetes:helm.sh/v3:Release (apisix):
    warning: Helm release "apache-apisix" was created but has a failed status. Use the `helm` command to investigate the error, correct it, then retry. Reason: context deadline exceeded
    error: 1 error occurred:
    	* Helm release "operations/apache-apisix" was created, but failed to initialize completely. Use Helm CLI to investigate: failed to become available within allocated timeout. Error: Helm Release operations/apache-apisix: context deadline exceeded

  pulumi:pulumi:Stack (ol-local-dev-infra-local-dev.infra.Dev):
    error: update failed

Outputs:
  + keycloak_url  : "https://sso.ol.mit.dev"/
  + litellm_url   : "http://litellm.local-infra.svc.cluster.local:4000"/
  + mailpit_ui_url: "http://mailpit.local-infra.svc.cluster.local:8025"/
  + ollama_url    : "http://ollama.local-infra.svc.cluster.local:11434"/
  + opensearch_url: "http://opensearch-master.local-infra.svc.cluster.local:9200"/
  + postgres_host : "local-pg-rw.local-infra.svc.cluster.local"
  + qdrant_url    : "http://qdrant.local-infra.svc.cluster.local:6333"/
  + tika_url      : "http://tika.local-infra.svc.cluster.local:9998"/
  + valkey_host   : "valkey.local-infra.svc.cluster.local"

Resources:
    + 49 created
    1 unchanged
    5 errored

Duration: 18m21s

Build Failed: Command "LOCAL_DEV_ROOT_DOMAIN=mit.dev PULUMI_CONFIG_PASSPHRASE='' pulumi up --yes --skip-preview --refresh --stack local-dev.infra.Dev" failed: exit status 1
```

Hitting any of the routes e.g mitxonline.mit.dev results in `{"error_msg":"404 Route Not Found"}`. I am assuming this is due to the apisix failures above.

daniellefrappier18:
@blarghmatey 
Tested against the current branch. Got the stack mostly running but hit some issues worth flagging:

- Port 5000 conflict — k3d-registry.localhost tries to bind port 5000, which is occupied by macOS AirPlay Receiver on modern Macs. I was able to work around this by disabling AirPlay Receiver in System Settings.

- Python 3.14 incompatibility — Pulumi's asyncio runtime crashes with KeyboardInterrupt on Python 3.14 (the default on a fresh macOS install with uv). Root cause: uv picks up the root workspace requires-python = ">=3.13, <3.15" and resolves to 3.14, which has breaking changes in kqueue/asyncio behavior. Workaround: manually create venv directories in local-dev/infra/core and local-dev/infra/apps_infra using Python 3.12. 

- odl-video-service crashes on nplusone — The bootstrap container fails with ModuleNotFoundError: No module named 'nplusone'. The local-dev config sets DEBUG=True which causes Django to load nplusone, but the production image (mitodl/odl-video-service-app:latest) doesn't have it installed. 

- learn-ai fixture deserialization error — Bootstrap crashes with DeserializationError: Problem installing fixture '/src/ai_chatbots/fixtures/initial_llm_models.json': Error deserializing object: 'model'. This appears to be a mismatch between the fixture and the current model definitions — possibly a renamed model or missing migration.

local-dev/apps/mit-learn/deployment.yaml
```diff
@@ -0,0 +1,401 @@
...
+        livenessProbe:
+          httpGet:
+            path: /health/
+            port: 8061
+          initialDelaySeconds: 30
+          periodSeconds: 30
+          failureThreshold: 3
+        volumeMounts:
+        - name: mkcert-root-ca
+          mountPath: /etc/ssl/local-dev
```
mbertrand: Add `timeoutSeconds: 5` to both probes (default is 1s):

```yaml
        readinessProbe:
          httpGet:
            path: /health/
            port: 8061
          initialDelaySeconds: 15
          periodSeconds: 10
          timeoutSeconds: 5
          failureThreshold: 6
        livenessProbe:
          httpGet:
            path: /health/
            port: 8061
          initialDelaySeconds: 30
          periodSeconds: 30
          timeoutSeconds: 5
          failureThreshold: 3
```

Why: `/health/` aggregates several backend checks (cache, migrations, celery_ping, redis, db) and routinely takes more than 1s during warmup. With the default `timeoutSeconds: 1`, the liveness probe SIGKILLs the container before it can ever pass — observed 75 restarts in a row with the container never reaching Ready.

local-dev/apps/mit-learn-nextjs/deployment.yaml
```diff
@@ -0,0 +1,74 @@
...
+            cpu: 100m
+            memory: 256Mi
+          limits:
+            memory: 1Gi
+        readinessProbe:
+          httpGet:
+            path: /
+            port: 3000
+          initialDelaySeconds: 15
+          periodSeconds: 10
```
mbertrand: Add `timeoutSeconds: 5` (default 1s):

```yaml
        readinessProbe:
          httpGet:
            path: /
            port: 3000
          initialDelaySeconds: 15
          periodSeconds: 10
          timeoutSeconds: 5
```

Why: same reason as the mit-learn webapp — Next.js root takes more than 1s to respond during startup. Without this, the pod stays `0/1 Ready` indefinitely and Tilt's resource never goes green.

local-dev/apps/mit-learn/secrets.yaml
```diff
@@ -0,0 +1,49 @@
...
+stringData:
+  # ── Database ───────────────────────────────────────────────────────────────
+  # Format: postgres://<user>:<password>@<host>:<port>/<db>
+  # Password matches the pg-app-credentials Secret created by the Pulumi infra stack.
+  DATABASE_URL: "postgres://app:localdev@local-pg-rw.local-infra.svc.cluster.local:5432/mitlearn"  # pragma: allowlist secret
+
+  # ── Valkey ─────────────────────────────────────────────────────────────────
+  CELERY_BROKER_URL: "redis://valkey.local-infra.svc.cluster.local:6379/0"
+  REDISCACHE_URL: "redis://valkey.local-infra.svc.cluster.local:6379/1"
+  CELERY_RESULT_BACKEND: "redis://valkey.local-infra.svc.cluster.local:6379/2"
```
mbertrand, line 26: Add `REDIS_URL` alongside the existing `REDISCACHE_URL`:

```yaml
  CELERY_BROKER_URL: "redis://valkey.local-infra.svc.cluster.local:6379/0"
  REDISCACHE_URL: "redis://valkey.local-infra.svc.cluster.local:6379/1"
  CELERY_RESULT_BACKEND: "redis://valkey.local-infra.svc.cluster.local:6379/2"
  # health_check.contrib.redis reads settings.REDIS_URL directly; without it
  # the health probe throws AttributeError and /health/ returns 500.
  REDIS_URL: "redis://valkey.local-infra.svc.cluster.local:6379/1"
```

Why: `django-health-check`'s redis backend uses `getattr(settings, "REDIS_URL", "redis://localhost/1")`. mit-learn's `settings_celery.py` defines `REDIS_URL = get_string("REDIS_URL", get_string("REDISCLOUD_URL", None))` — both env vars are unset in local-dev, so `REDIS_URL` becomes `None`, and `redis.from_url(None)` throws `AttributeError: 'NoneType' object has no attribute 'startswith'`. That short-circuits `/health/` to 500 and the readiness probe never passes.

local-dev/apps/mit-learn/Tiltfile
```diff
@@ -0,0 +1,95 @@
...
+k8s_resource(
+    "mitlearn-worker-embeddings",
+    labels=["mit-learn"],
+    resource_deps=["mitlearn-webapp"],
+)
+
+k8s_resource(
+    "mitlearn-beat",
+    labels=["mit-learn"],
+    resource_deps=["mitlearn-webapp"],
```
mbertrand: Change worker/beat `resource_deps` from `["mitlearn-webapp"]` → `["local-infra-apps"]` (apply to all four):

```python
k8s_resource(
    "mitlearn-worker-default",
    labels=["mit-learn"],
    resource_deps=["local-infra-apps"],
)
# ... and similarly for mitlearn-worker-edx-content, mitlearn-worker-embeddings, mitlearn-beat
```

Why: deadlock. The webapp's `/health/` includes `health_check.contrib.celery_ping`, which fails until at least one worker is reachable on the broker. With workers gated on the webapp being Ready, the workers never deploy → the celery health check always fails → webapp never Ready → workers never deploy. Letting workers come up in parallel with the webapp (same dep as the webapp itself) breaks the cycle.

local-dev/apps/mit-learn/apisix-routes.yaml
```diff
@@ -0,0 +1,235 @@
...
+        realm: olapps
+        introspection_endpoint_auth_method: client_secret_post
+        scope: "openid profile email ol-profile"
+        redirect_uri: "https://api.learn.mit.dev/login/ol-oidc/callback/"
+        logout_path: "/logout/oidc"
+        post_logout_redirect_uri: "https://api.learn.mit.dev/logout/"
+        unauth_action: pass
+        session:
+          secret: "local-dev-apisix-oidc-session-secret"  # pragma: allowlist secret
+        ssl_verify: false
```
mbertrand, line 89: Each `openid-connect` plugin block here (4 in this file) needs two new fields:

```yaml
    - name: openid-connect
      enable: true
      secretRef: ol-mitlearn-oidc
      config:
        bearer_only: false
        # Verify id_token RS256 signature using Keycloak's JWKS rather than
        # falling back to client_secret (which only works for HS* algorithms).
        use_jwks: true
        realm: olapps
        # APISIX runs in-cluster; sso.ol.mit.dev resolves to 127.0.0.1 inside the pod.
        # Use the in-cluster Keycloak service so APISIX can fetch the discovery doc.
        discovery: "http://keycloak-service.local-infra.svc.cluster.local:8080/realms/olapps/.well-known/openid-configuration"
        introspection_endpoint_auth_method: client_secret_post
        # ...rest unchanged
```

Why two changes:
1. **`discovery` is required** — APISIX 3.15 makes the openid-connect plugin's `discovery` property mandatory. Without it, the ingress controller fails to sync the entire batch of routes (`property "discovery" is required`), which means even unrelated routes like `mitlearn-nextjs-route` fail to load and `learn.mit.dev` returns a generic APISIX 404. Pointing at the in-cluster Keycloak URL lets APISIX fetch the doc — `sso.ol.mit.dev` from inside the pod resolves to `127.0.0.1` (its own loopback) and won't work.
2. **`use_jwks: true`** — id_tokens are signed RS256. Without this, lua-resty-openidc tries to verify the JWT signature using the client_secret (HMAC), fails, and the OIDC callback returns 500 with `Decode secret is not a valid cert/public key`. Setting `use_jwks: true` makes APISIX fetch the JWKS from the discovery doc and verify properly.

(Same pattern needs to land in `learn-ai/apisix-routes.yaml` and `mitxonline/apisix-routes.yaml`.)

There's also a related runtime piece: APISIX needs to call Keycloak's `token_endpoint` server-to-server, and the discovery doc emits `https://sso.ol.mit.dev/...` URLs. From inside the pod, that hostname resolves to 127.0.0.1 → connection refused. Fixed by adding a CoreDNS override that maps `sso.ol.mit.dev` → APISIX gateway ClusterIP for in-cluster traffic. That override was applied live but is not in this PR — it should be added as a Pulumi-managed ConfigMap (`coredns-custom` in `kube-system`, key `sso.ol.mit.dev.server`).

local-dev/apps/learn-ai/apisix-routes.yaml
```diff
@@ -0,0 +1,168 @@
...
+        ssl_verify: false
+        renew_access_token_on_expiry: true
+        logout_path: "/logout"
+        post_logout_redirect_uri: "/"
+        session_contents:
+          access_token: true
+          enc_id_token: true
+          id_token: true
+          user: true
+        unauth_action: pass
```
mbertrand, line 94: All 3 `openid-connect` plugin blocks in this file need three new fields:

```yaml
      config:
        bearer_only: false
        use_jwks: true
        discovery: "http://keycloak-service.local-infra.svc.cluster.local:8080/realms/olapps/.well-known/openid-configuration"
        introspection_endpoint_auth_method: client_secret_basic
        scope: "openid profile email"
        ssl_verify: false
        renew_access_token_on_expiry: true
        logout_path: "/logout"
        post_logout_redirect_uri: "/"
        session:
          secret: "local-dev-learn-ai-oidc-session-secret"  # pragma: allowlist secret
        session_contents:
          ...
```

Why:
- **`discovery`** + **`use_jwks`** — same as the mit-learn block (see comment on `apps/mit-learn/apisix-routes.yaml`). Without `discovery`, the whole batch sync fails. Without `use_jwks`, RS256 id_token verification fails.
- **`session.secret`** — APISIX 3.15 requires `session.secret` whenever `bearer_only: false`. The block currently has no `session` field at all, so the route fails schema validation (`property "session.secret" is required when "bearer_only" is false`).

local-dev/apps/mitxonline/apisix-routes.yaml
```diff
@@ -0,0 +1,190 @@
...
+        session:
+          cookie:
+            domain: ".mitxonline.mit.dev"
+            lifetime: 1209600
+        session_contents:
+          access_token: true
+          enc_id_token: true
+          id_token: true
+          user: true
+        unauth_action: pass
```
mbertrand, line 100: All 3 `openid-connect` plugin blocks in this file need three new fields:

```yaml
      config:
        bearer_only: false
        use_jwks: true
        discovery: "http://keycloak-service.local-infra.svc.cluster.local:8080/realms/olapps/.well-known/openid-configuration"
        introspection_endpoint_auth_method: client_secret_basic
        scope: "openid profile email organization:*"
        ssl_verify: false
        renew_access_token_on_expiry: true
        logout_path: "/logout/oidc"
        post_logout_redirect_uri: "https://mitxonline.mit.dev/logout/"
        session:
          secret: "local-dev-mitxonline-oidc-session-secret"  # pragma: allowlist secret
          cookie:
            domain: ".mitxonline.mit.dev"
            lifetime: 1209600
        ...
```

Why:
- **`discovery`** + **`use_jwks`** — same as the mit-learn block (see comment on `apps/mit-learn/apisix-routes.yaml`).
- **`session.secret`** — the existing `session.cookie` block is present but no `session.secret`, which APISIX 3.15 requires whenever `bearer_only: false`.

local-dev/infra/modules/identity_core.py
```diff
@@ -0,0 +1,272 @@
...
+            ),
+            # The image is pre-built (kc.sh build was run during Docker build).
+            # startOptimized=True uses that build; False would re-run kc.sh build
+            # on every pod start (slow). The image already has organization baked
+            # in, so we do NOT set features.enabled here — that would cause a
+            # build-time options mismatch and crash with --optimized.
+            "startOptimized": True,
+            "hostname": {
+                "hostname": keycloak_hostname,
+            },
```
mbertrand, line 120: Make Keycloak's hostname a full URL with scheme:

```python
            "hostname": {
                # Full URL (with scheme) so Keycloak emits public-facing URLs in
                # the OIDC discovery doc instead of leaking its in-cluster listen
                # port (8080). Without the scheme, authorization_endpoint becomes
                # https://sso.ol.mit.dev:8080/... and breaks the browser redirect.
                "hostname": f"https://{keycloak_hostname}",
            },
```

Why: with just the bare hostname, Keycloak doesn't know its public scheme/port, so the discovery doc's `authorization_endpoint` (and other URLs) come out as `https://sso.ol.mit.dev:8080/...`. The browser is then redirected to port 8080 on the host, where nothing is listening → "refused to connect." With a full URL Keycloak emits clean `https://sso.ol.mit.dev/...` links.

(Same change is needed in `identity.py`.)

local-dev/infra/modules/identity.py
```diff
@@ -0,0 +1,281 @@
...
+            ),
+            # The image is pre-built (kc.sh build was run during Docker build).
+            # startOptimized=True uses that build; False would re-run kc.sh build
+            # on every pod start (slow). The image already has organization baked
+            # in, so we do NOT set features.enabled here — that would cause a
+            # build-time options mismatch and crash with --optimized.
+            "startOptimized": True,
+            "hostname": {
+                "hostname": keycloak_hostname,
+            },
```
mbertrand: Same Keycloak hostname fix as `identity_core.py` — make this a full URL:

```python
            "hostname": {
                "hostname": f"https://{keycloak_hostname}",
            },
```

See the comment on `identity_core.py` for the rationale.

local-dev/infra/modules/keycloak.py
```diff
@@ -0,0 +1,600 @@
...
+                        {
+                            "client_id": args["client_id"],
+                            "client_secret": args["client_secret"],
+                            "realm_id": args["realm_id"],
+                            "realm_name": args["realm_name"],
+                            "url": args["url"],
+                        }
+                    ),
+                }
+            ),
```
mbertrand, line 367: Drop the top-level `secret` key from the OIDC secret:

```python
        return k8s.core.v1.Secret(
            resource_name,
            metadata={"name": secret_name, "namespace": namespace},
            # NOTE: do NOT include a top-level `secret` key here. APISIX's
            # secretRef merges all secret data into the openid-connect plugin
            # config as `conf.<key>`, so `secret` would land as `opts.secret`
            # in lua-resty-openidc and be misinterpreted as the JWT
            # verification key, breaking RS256 id_token validation. The
            # APISIX session cookie secret is set inline in each route's
            # session.secret field instead.
            string_data=Output.all(
                client_id=client.client_id,
                client_secret=client_secret_value,
                realm_id=client.realm_id,
                realm_name="olapps",
                url=client.realm_id.apply(lambda rid: f"{keycloak_url}/realms/{rid}"),
            ).apply(
                lambda args: {
                    "client_id": args["client_id"],
                    "client_secret": args["client_secret"],
                    "realm_id": args["realm_id"],
                    "realm_name": args["realm_name"],
                    "url": args["url"],
                    "oidc_credentials": json.dumps({...}),
                }
            ),
            opts=k8s_opts,
        )
```

Why: this was the root cause of the OIDC callback returning a 500 with `Decode secret is not a valid cert/public key`. APISIX's `secretRef` flattens every key in the referenced k8s Secret into the plugin's `conf` table. The previous version stored `session_secret_value` under the key `secret`, which APISIX then passed to lua-resty-openidc as `opts.secret`. lua-resty-openidc treats `opts.secret` as the deprecated JWT verification key (logs `using deprecated option opts.secret for asymmetric key`) and uses it to verify RS256 id_tokens — which fails because it's not a key.

Removing the field is enough; the session cookie secret is already set inline in each ApisixRoute's `session.secret` field. The function still takes `session_secret_value` for now — that parameter (and the call sites that pass it) can be cleaned up if you'd like, since it's no longer used.

local-dev/apps/mit-learn/configmaps/app-env.yaml
```diff
@@ -0,0 +1,135 @@
...
+  MITOL_DB_CONN_MAX_AGE: "0"
+  MITOL_DEFAULT_SITE_KEY: "micromasters"
+  MITOL_FRONTPAGE_DIGEST_MAX_POSTS: "10"
+  USE_X_FORWARDED_HOST: "True"
+  USE_X_FORWARDED_PORT: "True"
+  SECURE_CROSS_ORIGIN_OPENER_POLICY: "None"
+  NODE_MODULES_CACHE: "False"
+  # Required — ImproperlyConfigured raised if unset.
+  MITOL_APP_BASE_URL: "https://learn.mit.dev"
+  MITOL_API_BASE_URL: "https://api.learn.mit.dev"
```
mbertrand, line 30: Add `MITOL_NEW_USER_LOGIN_URL` to the configmap:

```yaml
  MITOL_APP_BASE_URL: "https://learn.mit.dev"
  MITOL_API_BASE_URL: "https://api.learn.mit.dev"
  # Default in settings.py is the legacy http://open.odl.local:8062/onboarding —
  # override so first-time users land on the local-dev frontend, not a dead host.
  MITOL_NEW_USER_LOGIN_URL: "https://learn.mit.dev/onboarding"
```

Why: after a successful first-time OIDC login, mit-learn redirects new users to `MITOL_NEW_USER_LOGIN_URL`, which defaults in `settings.py` to the legacy `http://open.odl.local:8062/onboarding`. That hostname doesn't resolve in the local-dev environment, so users land on a "site can't be reached" page instead of `/onboarding`.

local-dev/README.md
```diff
@@ -0,0 +1,733 @@
...
+
+- **Docker Desktop WSL integration:** Settings → Resources → WSL Integration → enable your distro. The k3d cluster runs inside WSL; Docker Desktop forwards `127.0.0.1` to Windows so browsers can reach it.
+- **`/etc/hosts` persistence:** WSL regenerates `/etc/hosts` by default on every restart. `setup.sh` adds `generateHosts = false` to `/etc/wsl.conf` automatically. If it does, run `wsl --shutdown` in Windows PowerShell then reopen your terminal before proceeding.
+- **Windows hosts file:** Your Windows browser resolves DNS from `C:\Windows\System32\drivers\etc\hosts`, not WSL's `/etc/hosts`. `setup.sh` attempts to update the Windows hosts file automatically; if it cannot (the file requires Windows admin elevation), it prints the entries and an `Add-Content` PowerShell command to paste into an elevated terminal.
+- **TLS trust on Windows:** The mkcert root CA installed in WSL is not trusted by Windows. After running `setup.sh`, run the `certutil` command it prints in an **elevated** Windows PowerShell to add the CA to the Windows Root certificate store, then restart your browser.
```
mbertrand, line 71: Add a bullet to the WSL2 prerequisites about the kubeconfig location:

> - **Kubeconfig location:** If `~/.kube` is a symlink to your Windows kubeconfig (e.g. `/mnt/c/Users/<you>/.kube`), Windows-side `kubectl` or Docker Desktop can rewrite the file and drop the `local-dev` context (or unset the current context) without warning, breaking `local-infra-core`. Prefer a Linux-native kubeconfig — either remove the symlink before running `setup.sh`, or set `export KUBECONFIG=$HOME/.kube/local-dev-config` in your shell rc and let `setup.sh` write there.

Why: hit this in WSL2 — `~/.kube/config` was a symlink to the Windows-side file (which only had EKS contexts, no `local-dev`, current-context unset). Pulumi failed every refresh with `Loading Kubernetes config: invalid configuration: no configuration has been provided`, and Tilt's `local-infra-core` resource showed "Update error" with no obvious hint that the cause was the kubeconfig.

local-dev/README.md
```diff
@@ -0,0 +1,733 @@
...
+### `tilt up` fails on `local-infra` (Pulumi errors)
+
+```bash
+cd local-dev/infra
+pulumi up --logtostderr -v=3   # Verbose output
+```
+
+Common causes:
+- **kubeconfig not set:** Ensure `k3d kubeconfig merge local-dev --kubeconfig-merge-default` has been run, or set `ol-local-dev-infra:kubeconfig` in `Pulumi.local-dev.infra.Dev.yaml`.
+- **Cert files missing:** Run `setup.sh --skip-hosts` to regenerate certs without touching `/etc/hosts`.
```
mbertrand: Add a dedicated troubleshooting subsection just below the existing "kubeconfig not set" bullet (and tweak that bullet to cross-link). The current single line gets you started but doesn't show the symptom or the full recovery sequence — which I had to figure out from scratch.

```markdown
- **kubeconfig not set:** Ensure `k3d kubeconfig merge local-dev --kubeconfig-merge-default` has been run, or set `ol-local-dev-infra:kubeconfig` in `Pulumi.local-dev.infra.Dev.yaml`. See also the dedicated entry below for the full recovery sequence.
- **Cert files missing:** Run `setup.sh --skip-hosts` to regenerate certs without touching `/etc/hosts`.

### `local-infra-core` shows "Update error" with `invalid configuration: no configuration has been provided`

Pulumi can't find a kubeconfig. Every Kubernetes resource refresh in the `local-infra-core` Tilt resource will fail with `Loading Kubernetes config: invalid configuration: no configuration has been provided, try setting KUBERNETES_MASTER environment variable` and Tilt's overview will mark the resource red.

Check what kubectl sees:

​```bash
kubectl config current-context
kubectl config get-contexts | grep local-dev
​```

If the context is missing or current-context is unset, restore it:

​```bash
k3d kubeconfig merge local-dev --kubeconfig-merge-default
kubectl config delete-context local-dev 2>/dev/null || true
kubectl config rename-context k3d-local-dev local-dev
kubectl config use-context local-dev
kubectl get nodes   # should list 3 k3d nodes Ready
​```

Then trigger a rebuild of `local-infra-core` from the Tilt UI (the refresh icon next to the resource).

**WSL2:** if this recurs, your `~/.kube/config` is likely symlinked to the Windows kubeconfig — see the [Windows (WSL2)](#windows-wsl2) prereqs note about kubeconfig location.
```

Why: the symptom (Tilt "Update error" with a wall of Pulumi errors deep in the log) is non-obvious, and the existing bullet only hints at the merge command — it doesn't mention `kubectl config use-context`, which is what was actually missing in my case.

local-dev/scripts/setup.sh
```diff
@@ -0,0 +1,413 @@
...
+        export PULUMI_CONFIG_PASSPHRASE=""
+        # Init the stack if it doesn't exist yet.
+        if ! pulumi stack ls --json 2>/dev/null \
+            | python3 -c "import json,sys; stacks=[s['name'] for s in json.load(sys.stdin)]; sys.exit(0 if 'local-dev.infra.Dev' in stacks else 1)" 2>/dev/null; then
+            pulumi stack init local-dev.infra.Dev --secrets-provider=passphrase
+        fi
+    )
+    ok "Pulumi infra stack ready."
+fi
+
```
mbertrand, line 409: After commit `ed040e89b` removed the monolithic `local-dev/infra/Pulumi.yaml` project, this block runs `pulumi stack init local-dev.infra.Dev` from a directory that no longer has a `Pulumi.yaml` — Pulumi falls back to the cloud backend, the default-org lookup fails, and emits a misleading `"this command requires logging in"` error. The Tiltfile already inits `local-dev.core.Dev` and `local-dev.apps-infra.Dev` from inside the subproject dirs (where each `Pulumi.yaml` declares the file backend), so this block is redundant. Suggest deleting it.

```suggestion
```

local-dev/apps/mit-learn-nextjs/Tiltfile
```diff
@@ -0,0 +1,71 @@
...
+
+if os.path.exists(os.path.join(APP_DIR, '.git')):
+    docker_build(
+        IMAGE,
+        context=APP_DIR,
+        dockerfile=APP_DIR + '/frontends/main/Dockerfile.web',
+        build_args={
+            # Core — required by validateEnv.js
+            'NEXT_PUBLIC_ORIGIN': 'https://learn.{rd}'.format(rd=_rd),
+            'NEXT_PUBLIC_MITOL_API_BASE_URL': 'https://api.learn.{rd}'.format(rd=_rd),
```
mbertrand, line 27: The mit-learn frontend's [`common/certificateUtils.ts`](https://github.com/mitodl/mit-learn/blob/main/frontends/main/src/common/certificateUtils.ts) does `invariant(API_BASE_URL, "NEXT_PUBLIC_MITX_ONLINE_LEGACY_BASE_URL must be set")` at module load. The Dockerfile declares the ARG, but the Tiltfile never passes it, so `yarn build` aborts when collecting page data for `/certificate/[certificateType]/[uuid]`.

```suggestion
            'NEXT_PUBLIC_MITOL_API_BASE_URL': 'https://api.learn.{rd}'.format(rd=_rd),
            'NEXT_PUBLIC_MITX_ONLINE_LEGACY_BASE_URL': 'https://mitxonline.{rd}'.format(rd=_rd),
```

local-dev/apps/learn-ai/deployment.yaml
```diff
@@ -0,0 +1,272 @@
...
+        imagePullPolicy: IfNotPresent
+        command:
+        - /bin/sh
+        - -c
+        - |
+          python manage.py collectstatic --noinput --clear
+          python manage.py migrate --noinput
+          python manage.py createcachetable
+          RUN_DATA_MIGRATIONS=true python manage.py migrate --noinput
+          python manage.py loaddata initial_llm_models
```
mbertrand: `ai_chatbots/fixtures/initial_llm_models.json` is a flat list of dicts (`{provider, name, litellm_id, enabled}`), not a Django serialized fixture — `loaddata` crashes with `KeyError: 'model'`. The same JSON is already consumed by data migration `ai_chatbots/migrations/0005_llmmodel.py`, which runs via the `migrate` line above and `get_or_create`s every row. So the table is populated correctly without this step.

```suggestion
```

local-dev/apps/learn-ai/deployment.yaml
```diff
@@ -0,0 +1,272 @@
...
+        envFrom:
+        - configMapRef:
+            name: learn-ai-env
+        - secretRef:
+            name: learn-ai-secrets
+        volumeMounts:
+        - name: static-files
+          mountPath: /src/staticfiles
```
mbertrand: The configmap sets `SSL_CERT_FILE=/etc/ssl/local-dev/rootCA.pem` and `REQUESTS_CA_BUNDLE=/etc/ssl/local-dev/rootCA.pem`, but the bootstrap init container doesn't mount that path. Any httpx import path during `migrate` (e.g. a Qdrant or LiteLLM client constructed at import time) calls `ssl.create_default_context(cafile=os.environ["SSL_CERT_FILE"])` and dies with `FileNotFoundError`. The runtime `app` container already mounts this volume — mirror it here.

```suggestion
        volumeMounts:
        - name: static-files
          mountPath: /src/staticfiles
        - name: mkcert-root-ca
          mountPath: /etc/ssl/local-dev
          readOnly: true
```

local-dev/apps/learn-ai/deployment.yaml
```diff
@@ -0,0 +1,272 @@
...
+      - name: celery-beat
+        image: mitodl/learn-ai-app:latest
+        imagePullPolicy: IfNotPresent
+        command:
+        - celery
+        - -A
+        - main.celery:app
+        - beat
+        - --scheduler
+        - celery.beat.PersistentScheduler
```
mbertrand: The `mitodl/learn-ai-app` image has stale `celerybeat-schedule*` files baked into `/src` as `root:root 0644`. The container runs as the `mitodl` user (uid 1000), can't overwrite them, and beat crashes with `dbm.sqlite3.error: attempt to write a readonly database` on every start. Pointing the schedule at `/tmp` sidesteps the read-only files in the image.

```suggestion
        - --scheduler
        - celery.beat.PersistentScheduler
        - --schedule
        - /tmp/celerybeat-schedule
```

local-dev/apps/mit-learn/deployment.yaml
```diff
@@ -0,0 +1,401 @@
...
+            name: mitlearn-secrets
+        resources:
+          requests:
+            cpu: 100m
+            memory: 512Mi
+          limits:
+            memory: 1Gi
+        volumeMounts:
+        - name: static-files
+          mountPath: /src/staticfiles
```
mbertrand: Same fix as learn-ai's bootstrap. The configmap sets `SSL_CERT_FILE=/etc/ssl/local-dev/rootCA.pem`, but this init container doesn't mount that path. `create_qdrant_collections` (run during `migrate`) constructs an httpx client, which calls `ssl.create_default_context(cafile=os.environ["SSL_CERT_FILE"])` and crashes with `FileNotFoundError`. The runtime `app` container already mounts this volume — mirror it here.

```suggestion
        volumeMounts:
        - name: static-files
          mountPath: /src/staticfiles
        - name: mkcert-root-ca
          mountPath: /etc/ssl/local-dev
          readOnly: true
```

local-dev/apps/mit-learn/configmaps/app-env.yaml
```diff
@@ -0,0 +1,135 @@
...
+
+  # ── OpenSearch ─────────────────────────────────────────────────────────────
+  OPENSEARCH_INDEX: "mitlearn-local"
+  OPENSEARCH_URL: "http://opensearch-master.local-infra.svc.cluster.local:9200"
+  OPENSEARCH_DEFAULT_TIMEOUT: "30"
+  OPENSEARCH_INDEXING_CHUNK_SIZE: "75"
+
+  # ── Qdrant (local — use SparseHashEncoder which runs locally via sklearn) ──
+  # Use _V2 suffixes — these are the env var names the app reads in settings.py.
+  QDRANT_HOST_V2: "http://qdrant.local-infra.svc.cluster.local:6333"
```
mbertrand: mit-learn reads `QDRANT_HOST` (no `_V2`) in `main/settings.py:796`. Only `QDRANT_SPARSE_MODEL_V2` and `QDRANT_SPARSE_ENCODER_V2` use the `_V2` suffix — the comment overgeneralizes. With `QDRANT_HOST_V2` set but `QDRANT_HOST` unset, Django falls back to the default `http://qdrant:6333` and `create_qdrant_collections` fails with `DNS resolution failed for qdrant:6334`.

```suggestion
  # Host/encoder vars are read without _V2; sparse model/encoder vars are read
  # with _V2 (see main/settings.py).
  QDRANT_HOST: "http://qdrant.local-infra.svc.cluster.local:6333"
```

local-dev/apps/mit-learn/deployment.yaml
```diff
@@ -0,0 +1,401 @@
...
+        - configMapRef:
+            name: mitlearn-env
+        - secretRef:
+            name: mitlearn-secrets
+        resources:
+          requests:
+            cpu: 100m
+            memory: 512Mi
+          limits:
+            memory: 1Gi
```
mbertrand: Bump from `1Gi` to `4Gi`. Running `./manage.py backpopulate_xpro_data` (and similar batch management commands via `kubectl exec`) gets OOMKilled at 1Gi (exit 137). 4Gi gives the webapp `app` container headroom for large backpopulates without rotating limits up and down per task.

local-dev/apps/mit-learn/deployment.yaml
```diff
@@ -0,0 +1,401 @@
...
```
mbertrand: I instantly got an OOM error when trying the mgmt command `backpopulate_xpro_data` for mit-learn.  So I bumped this from `1Gi` to `2Gi`, and did the same to the `celery` containers in the other two worker deployments in this file.

local-dev/apps/mit-learn/deployment.yaml
```diff
@@ -0,0 +1,401 @@
...
+          failureThreshold: 6
+        livenessProbe:
+          httpGet:
+            path: /health/
+            port: 8061
+          initialDelaySeconds: 30
+          periodSeconds: 30
+          failureThreshold: 3
+        volumeMounts:
+        - name: mkcert-root-ca
```
mbertrand: Pointing `REQUESTS_CA_BUNDLE`/`SSL_CERT_FILE` at the mkcert-only `rootCA.pem` (see `app-env.yaml`) replaces the system CA bundle entirely, so public HTTPS hosts like `xpro.mit.edu`, GitHub, and the LiteLLM model-prices URL all fail cert verification from inside the pods.

Suggest: keep the mkcert ConfigMap volume but only mount it into a new `ca-bundle-merge` initContainer that concatenates the image's system bundle (`/etc/ssl/certs/ca-certificates.crt`) with the mkcert root and writes the result into a shared `combined-ca` emptyDir. Main containers then mount `combined-ca` at `/etc/ssl/local-dev`, so the env var values in `app-env.yaml` don't need to change but the bundle they point to now trusts both public and locally-signed services.

```yaml
initContainers:
- name: ca-bundle-merge
  image: mitodl/mit-learn-app:latest
  imagePullPolicy: IfNotPresent
  command: ["/bin/sh", "-c"]
  args:
  - cat /etc/ssl/certs/ca-certificates.crt /mkcert-src/rootCA.pem > /output/rootCA.pem
  volumeMounts:
  - {name: mkcert-root-ca, mountPath: /mkcert-src, readOnly: true}
  - {name: combined-ca, mountPath: /output}
volumes:
- name: combined-ca
  emptyDir: {}
```

The same pattern should be applied to all 5 pod templates in this file (webapp + 3 workers + beat). For the webapp, the new init must run **before** the existing `bootstrap` init so migrations / qdrant setup / LiteLLM also see the merged bundle. The accompanying comment update in `app-env.yaml` (line 32) should be paired with this.

local-dev/apps/mit-learn/configmaps/app-env.yaml
```diff
@@ -0,0 +1,135 @@
...
+  # DATABASE_URL is set in secrets (contains password).
+  # MITOL_DB_DISABLE_SSL is set above.
+
+  # ── Valkey / Celery ────────────────────────────────────────────────────────
+  # CELERY_BROKER_URL and REDISCACHE_URL are set in secrets.
+  CELERY_TASK_ALWAYS_EAGER: "False"
+
+  # ── OpenSearch ─────────────────────────────────────────────────────────────
+  OPENSEARCH_INDEX: "mitlearn-local"
+  OPENSEARCH_URL: "http://opensearch-master.local-infra.svc.cluster.local:9200"
```
mbertrand: Service name is wrong — `kubectl get svc -n local-infra` shows `opensearch-cluster-master` (the helm chart's default), not `opensearch-master`. With the current value, every indexing task and `./manage.py recreate_index --all` fails with `NameResolutionError`, so the `mitlearn-local*` indices never get created and `/search` returns nothing.

```yaml
OPENSEARCH_URL: "http://opensearch-cluster-master.local-infra.svc.cluster.local:9200"
```

mbertrand (changes requested):
I had to make some local changes (w/help from Claude) to get mitlearn running with a functional login process, ETL mgmt commands, and opensearch indexing.  Added comments for each change.  I was running this via WSL for what it's worth.

tilt_config.json.example
```diff
@@ -0,0 +1,19 @@
...
+    "mit-learn=0.62.0",
+    "mit-learn-nextjs=0.62.0",
+    "learn-ai=0.28.3",
+    "mitxonline=1.144.5",
+    "odl-video-service=0.85.0"
+  ],
+
+  "openai_api_key": "",
+  "langsmith_api_key": "",
+  "canvas_ai_token": ""
```
mbertrand, line 18: Five new PostHog config keys for the three-mode wiring (real keys / bootstrap flags / disabled). `tilt_config.json` is gitignored, so it's safe to paste real credentials.

```suggestion
  "canvas_ai_token": "",

  "posthog_project_api_key": "",
  "posthog_personal_api_key": "",
  "posthog_api_host": "https://us.i.posthog.com",
  "posthog_project_id": "",

  "posthog_bootstrap_flags": []
```

Tiltfile
```diff
@@ -0,0 +1,204 @@
...
+config.define_string_list("prebuilt_tags", usage="Prebuilt image tag overrides per app, e.g. mit-learn=0.62.0 learn-ai=0.28.3")
+config.define_string("openai_api_key", usage="OpenAI API key for AI/embedding features (optional)")
+config.define_string("langsmith_api_key", usage="LangSmith API key for tracing (optional)")
+config.define_string("canvas_ai_token", usage="Canvas AI token (optional)")
+config.define_string("root_domain", usage="Root DNS domain for all local-dev services (default: mit.dev). Overrides LOCAL_DEV_ROOT_DOMAIN env var.")
+cfg = config.parse()
+
+enabled_apps = cfg.get("enabled_apps", ["mit-learn", "learn-ai", "mitxonline", "odl-video-service"])
+per_app_databases = cfg.get("per_app_databases", False)
+openedx_mode = cfg.get("openedx_mode", "qa")
```
mbertrand, line 20: Declare the PostHog keys via `config.define_*` so they show up in `tilt up --help`. Tilt's `include()` doesn't share globals into child Tiltfiles, so the actual values are re-read via `read_json(tilt_config.json)` in `mit-learn-nextjs/Tiltfile`, `mit-learn/Tiltfile`, and `learn-ai/Tiltfile`. The trailing comment block notes this so future maintainers don't try to wire the values through here.

```suggestion
config.define_string("root_domain", usage="Root DNS domain for all local-dev services (default: mit.dev). Overrides LOCAL_DEV_ROOT_DOMAIN env var.")
# PostHog: three modes (see README "PostHog feature flags" section)
#   1. Real keys supplied      → mirror prod/RC, server-side flags evaluated
#   2. Bootstrap flags only    → flags baked into the Next.js bundle, no network
#   3. Neither                 → PostHog off, AI features gated off (default)
config.define_string("posthog_project_api_key", usage="PostHog project (public) API key — enables real-keys mode")
config.define_string("posthog_personal_api_key", usage="PostHog personal API key — required for server-side flag evaluation")
config.define_string("posthog_api_host", usage="PostHog API host (default: https://us.i.posthog.com)")
config.define_string("posthog_project_id", usage="PostHog project numeric ID")
config.define_string_list("posthog_bootstrap_flags", usage="Flag names (kebab-case) to bake on at build time when no real key is supplied")
cfg = config.parse()

enabled_apps = cfg.get("enabled_apps", ["mit-learn", "learn-ai", "mitxonline", "odl-video-service"])
per_app_databases = cfg.get("per_app_databases", False)
openedx_mode = cfg.get("openedx_mode", "qa")

# PostHog config is read directly from tilt_config.json by the mit-learn and
# mit-learn-nextjs Tiltfiles via read_json(); declaring keys here makes them
# CLI-discoverable (tilt up --help) but the values aren't consumed in this file.
```

local-dev/apps/mit-learn-nextjs/Tiltfile
```diff
@@ -0,0 +1,71 @@
...
+            'NEXT_PUBLIC_POSTHOG_API_KEY': '',
+            'NEXT_PUBLIC_POSTHOG_API_HOST': '',
+            'NEXT_PUBLIC_POSTHOG_UI_HOST': '',
+            'NEXT_PUBLIC_POSTHOG_PROJECT_ID': '',
+            'NEXT_PUBLIC_SENTRY_DSN': '',
+            'NEXT_PUBLIC_SENTRY_ENV': 'local',
+            'NEXT_PUBLIC_APPZI_URL': '',
+            'NEXT_PUBLIC_HUBSPOT_PORTAL_ID': '',
+            'NEXT_PUBLIC_VERSION': 'local',
+        },
```
mbertrand, line 53: Lift `build_args` out of the `docker_build` call into a module-level `_build_args` variable so we can branch on PostHog mode before the build. The four `NEXT_PUBLIC_POSTHOG_*` entries are no longer hardcoded blanks — they're populated by the conditional:

- **Mode 1** (real key set in tilt_config.json) → wire `NEXT_PUBLIC_POSTHOG_API_KEY/HOST/UI_HOST/PROJECT_ID` so the frontend talks to the configured PostHog project.
- **Mode 2** (bootstrap flags only) → set a dummy key + bogus host (so `posthog.init` runs and the bootstrap path activates) and translate each flag name to `NEXT_PUBLIC_FEATURE_FOO_BAR=True`. `next.config.js` picks those up and bakes them into the bundle as bootstrapped feature flags.
- **Mode 3** (neither) → build_args stay blank, `ConfiguredPostHogProvider` skips `posthog.init` entirely, AI gates stay closed.

```suggestion

# PostHog: three modes (see local-dev/README.md "PostHog feature flags").
#   1. Real keys   → posthog_project_api_key set → live flag evaluation
#   2. Bootstrap   → posthog_bootstrap_flags non-empty → flags baked at build time
#   3. Disabled    → neither set → AskTIM and other AI gates stay off
_tilt_config = read_json(os.path.join(config.main_dir, "tilt_config.json"), default={})
_ph_key = _tilt_config.get("posthog_project_api_key", "")
_ph_host = _tilt_config.get("posthog_api_host", "https://us.i.posthog.com")
_ph_project_id = _tilt_config.get("posthog_project_id", "")
_ph_bootstrap_flags = _tilt_config.get("posthog_bootstrap_flags", [])

_build_args = {
    # Core — required by validateEnv.js
    'NEXT_PUBLIC_ORIGIN': 'https://learn.{rd}'.format(rd=_rd),
    'NEXT_PUBLIC_MITOL_API_BASE_URL': 'https://api.learn.{rd}'.format(rd=_rd),
    'NEXT_PUBLIC_SITE_NAME': 'MIT Learn',
    'NEXT_PUBLIC_MITOL_SUPPORT_EMAIL': 'mitlearn-support@mit.edu',
    'NEXT_PUBLIC_CSRF_COOKIE_NAME': 'csrftoken',
    'NEXT_PUBLIC_MITOL_AXIOS_WITH_CREDENTIALS': 'true',

    # AI endpoints (point to learn-ai local service)
    'NEXT_PUBLIC_LEARN_AI_RECOMMENDATION_ENDPOINT':
        'https://ai.learn.{rd}/api/v0/recommendations/'.format(rd=_rd),
    'NEXT_PUBLIC_LEARN_AI_SYLLABUS_ENDPOINT':
        'https://ai.learn.{rd}/api/v0/syllabus/'.format(rd=_rd),

    # PostHog — populated below based on mode (real / bootstrap / disabled)
    'NEXT_PUBLIC_POSTHOG_API_KEY': '',
    'NEXT_PUBLIC_POSTHOG_API_HOST': '',
    'NEXT_PUBLIC_POSTHOG_UI_HOST': '',
    'NEXT_PUBLIC_POSTHOG_PROJECT_ID': '',

    # Optional — blank disables analytics/monitoring
    'NEXT_PUBLIC_EMBEDLY_KEY': '',
    'NEXT_PUBLIC_SENTRY_DSN': '',
    'NEXT_PUBLIC_SENTRY_ENV': 'local',
    'NEXT_PUBLIC_APPZI_URL': '',
    'NEXT_PUBLIC_HUBSPOT_PORTAL_ID': '',
    'NEXT_PUBLIC_VERSION': 'local',
}

if _ph_key:
    # Mode 1: real keys → live flag evaluation from PostHog
    _build_args['NEXT_PUBLIC_POSTHOG_API_KEY'] = _ph_key
    _build_args['NEXT_PUBLIC_POSTHOG_API_HOST'] = _ph_host
    _build_args['NEXT_PUBLIC_POSTHOG_UI_HOST'] = _ph_host
    _build_args['NEXT_PUBLIC_POSTHOG_PROJECT_ID'] = str(_ph_project_id)
elif _ph_bootstrap_flags:
    # Mode 2: bootstrap. Non-empty key triggers posthog.init in
    # ConfiguredPostHogProvider; the bogus host means real PostHog network
    # calls fail (harmless console noise), so only bootstrapped flags apply.
    _build_args['NEXT_PUBLIC_POSTHOG_API_KEY'] = 'local-dev-bootstrap-noop'  # pragma: allowlist secret
    _build_args['NEXT_PUBLIC_POSTHOG_API_HOST'] = 'https://localhost.invalid'
    for _flag in _ph_bootstrap_flags:
        # next.config.js translates NEXT_PUBLIC_FEATURE_FOO_BAR=True →
        # bootstrap flag "foo-bar" via lowercase + underscore-to-dash.
        _build_args['NEXT_PUBLIC_FEATURE_' + _flag.upper().replace('-', '_')] = 'True'

if os.path.exists(os.path.join(APP_DIR, '.git')):
    docker_build(
        IMAGE,
        context=APP_DIR,
        dockerfile=APP_DIR + '/frontends/main/Dockerfile.web',
        build_args=_build_args,
```

local-dev/apps/mit-learn/Tiltfile
```diff
@@ -0,0 +1,95 @@
...
+# ─────────────────────────────────────────────────────────────
+
+# Apply all manifests in order.
+k8s_yaml([
+    os.path.join(config.main_dir, "local-dev/apps/mit-learn/configmaps/nginx.yaml"),
+    os.path.join(config.main_dir, "local-dev/apps/mit-learn/configmaps/app-env.yaml"),
+    os.path.join(config.main_dir, "local-dev/apps/mit-learn/secrets.yaml"),
+    os.path.join(config.main_dir, "local-dev/apps/mit-learn/deployment.yaml"),
+    os.path.join(config.main_dir, "local-dev/apps/mit-learn/apisix-routes.yaml"),
+])
```
mbertrand, line 54: Conditionally render a `mitlearn-posthog` ConfigMap when the developer supplies real PostHog keys in `tilt_config.json`. The deployment's `envFrom` references it as `optional: true`, so in bootstrap or disabled mode the ConfigMap simply doesn't exist and the env vars are absent — Django's `features.py:30` is a no-op when `POSTHOG_PROJECT_API_KEY` is blank.

```suggestion
])

# ────────────────────────────────────────────────────────────
# Optional PostHog ConfigMap (real-keys mode only)
#
# The deployment references "mitlearn-posthog" as optional: true via envFrom.
# When the developer supplies real PostHog keys in tilt_config.json, render
# the ConfigMap so Django's features.py can evaluate flags server-side.
# In bootstrap or disabled mode this is skipped entirely.
# ────────────────────────────────────────────────────────────
_tilt_config = read_json(os.path.join(config.main_dir, "tilt_config.json"), default={})
_ph_key = _tilt_config.get("posthog_project_api_key", "")
if _ph_key:
    _ph_personal = _tilt_config.get("posthog_personal_api_key", "")
    _ph_host = _tilt_config.get("posthog_api_host", "https://us.i.posthog.com")
    _ph_project_id = _tilt_config.get("posthog_project_id", "")
    k8s_yaml(blob("""---
apiVersion: v1
kind: ConfigMap
metadata:
  name: mitlearn-posthog
  namespace: {ns}
data:
  POSTHOG_PROJECT_API_KEY: "{key}"
  POSTHOG_PERSONAL_API_KEY: "{personal}"
  POSTHOG_API_HOST: "{host}"
  POSTHOG_PROJECT_ID: "{project_id}"
""".format(
        ns=NAMESPACE,
        key=_ph_key,
        personal=_ph_personal,
        host=_ph_host,
        project_id=_ph_project_id,
    )))
```

local-dev/apps/mit-learn/deployment.yaml
```diff
@@ -0,0 +1,401 @@
...
+
+          echo "==> Pruning subscription queries..."
+          python manage.py prune_subscription_queries
+
+          echo "==> Bootstrap complete."
+        envFrom:
+        - configMapRef:
+            name: mitlearn-env
+        - secretRef:
+            name: mitlearn-secrets
```
mbertrand: Add an optional `configMapRef: mitlearn-posthog` to the bootstrap init container's `envFrom`. `optional: true` means the deployment is fine in bootstrap/disabled mode (no ConfigMap exists). When the developer supplies real PostHog keys, the Tiltfile renders the ConfigMap and Django picks up `POSTHOG_PROJECT_API_KEY` + friends here.

**This same 3-line addition needs to be applied to all 6 `envFrom:` blocks in this file** (bootstrap, app, worker-default, worker-edx-content, worker-embeddings, beat) so every Django process sees the same env.

```suggestion
        envFrom:
        - configMapRef:
            name: mitlearn-env
        - secretRef:
            name: mitlearn-secrets
        - configMapRef:
            name: mitlearn-posthog
            optional: true
```

local-dev/apps/learn-ai/Tiltfile
```diff
@@ -0,0 +1,44 @@
...
+        ],
+    )
+
+k8s_yaml([
+    'configmaps/app-env.yaml',
+    'configmaps/nginx.yaml',
+    'secrets.yaml',
+    'deployment.yaml',
+    'apisix-routes.yaml',
+])
```
mbertrand, line 34: Same pattern as mit-learn/Tiltfile: render a `learnai-posthog` ConfigMap only when real PostHog keys are supplied. learn-ai's `main/features.py:31` also checks `POSTHOG_PROJECT_API_KEY` and is a no-op when blank, so this is safe to leave out in bootstrap/disabled mode.

```suggestion
])

# ────────────────────────────────────────────────────────────
# Optional PostHog ConfigMap (real-keys mode only)
#
# The deployment references "learnai-posthog" as optional: true via envFrom.
# When the developer supplies real PostHog keys in tilt_config.json, render
# the ConfigMap so Django's features.py can evaluate flags server-side.
# In bootstrap or disabled mode this is skipped entirely.
# ────────────────────────────────────────────────────────────
_tilt_config = read_json(os.path.join(config.main_dir, "tilt_config.json"), default={})
_ph_key = _tilt_config.get("posthog_project_api_key", "")
if _ph_key:
    _ph_personal = _tilt_config.get("posthog_personal_api_key", "")
    _ph_host = _tilt_config.get("posthog_api_host", "https://us.i.posthog.com")
    _ph_project_id = _tilt_config.get("posthog_project_id", "")
    k8s_yaml(blob("""---
apiVersion: v1
kind: ConfigMap
metadata:
  name: learnai-posthog
  namespace: {ns}
data:
  POSTHOG_PROJECT_API_KEY: "{key}"
  POSTHOG_PERSONAL_API_KEY: "{personal}"
  POSTHOG_API_HOST: "{host}"
  POSTHOG_PROJECT_ID: "{project_id}"
""".format(
        ns=NAMESPACE,
        key=_ph_key,
        personal=_ph_personal,
        host=_ph_host,
        project_id=_ph_project_id,
    )))
```

local-dev/apps/learn-ai/deployment.yaml
```diff
@@ -0,0 +1,272 @@
...
+          python manage.py collectstatic --noinput --clear
+          python manage.py migrate --noinput
+          python manage.py createcachetable
+          RUN_DATA_MIGRATIONS=true python manage.py migrate --noinput
+          python manage.py loaddata initial_llm_models
+        envFrom:
+        - configMapRef:
+            name: learn-ai-env
+        - secretRef:
+            name: learn-ai-secrets
```
mbertrand: Add an optional `configMapRef: learnai-posthog` to the bootstrap init container's `envFrom`. Same pattern as the mit-learn deployment.

**This same 3-line addition needs to be applied to all 5 `envFrom:` blocks in this file** (bootstrap, web, celery-default, celery-edx-content, celery-beat).

```suggestion
        envFrom:
        - configMapRef:
            name: learn-ai-env
        - secretRef:
            name: learn-ai-secrets
        - configMapRef:
            name: learnai-posthog
            optional: true
```

local-dev/README.md
```diff
@@ -0,0 +1,733 @@
...
+    "mit-learn=0.62.0",
+    "mit-learn-nextjs=0.62.0",
+    "learn-ai=0.28.3",
+    "mitxonline=1.144.5",
+    "odl-video-service=0.85.0"
+  ],
+
+  "openai_api_key": "",
+  "langsmith_api_key": "",
+  "canvas_ai_token": ""
```
mbertrand, line 392: Mirror the new keys from `tilt_config.json.example` in the docs.

```suggestion
  "canvas_ai_token": "",

  "posthog_project_api_key": "",
  "posthog_personal_api_key": "",
  "posthog_api_host": "https://us.i.posthog.com",
  "posthog_project_id": "",
  "posthog_bootstrap_flags": []
```

local-dev/README.md
```diff
@@ -0,0 +1,733 @@
...
+
+| Key | Default | Description |
+|-----|---------|-------------|
+| `enabled_apps` | all four | Apps to deploy. Omit any to skip it entirely. |
+| `per_app_databases` | `false` | `true` deploys isolated CNPG + Valkey per namespace (Phase 6A). |
+| `prebuilt_tags` | see file | `["app=tag"]` list of image tags used when the app repo is not checked out locally. |
+| `openai_api_key` | `""` | OpenAI API key — required for learn-ai vector search and LiteLLM. |
+| `langsmith_api_key` | `""` | LangSmith API key for tracing (optional). |
+| `canvas_ai_token` | `""` | Canvas AI token (optional). |
+
```
mbertrand, line 404: Document the new config keys in the reference table and add a dedicated "PostHog feature flags" section explaining the three modes. The section is written for someone with no prior context — when to use each mode, what gets wired where, and the rebuild caveat for Mode 2.

````suggestion
| `canvas_ai_token` | `""` | Canvas AI token (optional). |
| `posthog_project_api_key` | `""` | PostHog project (public) API key. Set to enable real flag evaluation. See [PostHog feature flags](#posthog-feature-flags). |
| `posthog_personal_api_key` | `""` | PostHog personal API key. Needed alongside the project key for server-side flag evaluation in Django. |
| `posthog_api_host` | `https://us.i.posthog.com` | PostHog API host (US cloud by default). |
| `posthog_project_id` | `""` | PostHog project numeric ID. |
| `posthog_bootstrap_flags` | `[]` | List of flag names (kebab-case) to bake on at build time when no real key is supplied. |

### PostHog feature flags

AI features in the mit-learn frontend (AskTIM, recommendation bot, syllabus chat, etc.) are gated behind PostHog feature flags. Without configuration, every gate is closed and the features stay hidden. There are three modes:

**Mode 1 — Real PostHog keys (recommended for QA-like behavior):**

Paste credentials from the RC project (or your own personal PostHog project) into `tilt_config.json`:

```json
"posthog_project_api_key": "phc_xxx",
"posthog_personal_api_key": "phx_xxx",
"posthog_api_host": "https://us.i.posthog.com",
"posthog_project_id": "12345"
```

This wires `NEXT_PUBLIC_POSTHOG_*` build args into the Next.js image **and** renders an optional `*-posthog` ConfigMap into the mit-learn and learn-ai namespaces so Django can evaluate flags server-side too. Flag changes in PostHog take effect on page reload (no rebuild). Your local-dev events will flow into the configured project, so consider using your own personal project for noisy iteration.

**Mode 2 — Bootstrap flags (offline-friendly, no PostHog account):**

List the flag names you want forced on. They're baked into the Next.js bundle via `NEXT_PUBLIC_FEATURE_*` build args (see [`next.config.js` in mit-learn](../../mit-learn/frontends/main/next.config.js) for the conversion rule):

```json
"posthog_bootstrap_flags": [
  "recommendation-bot",
  "home-page-recommendation-bot",
  "lr-drawer-chatbot",
  "pr-drawer-chatbot",
  "universal-ai",
  "universal-ai-search-banner"
]
```

PostHog will be initialized with a dummy key pointing at a bogus host, so the bootstrapped flag values apply but no telemetry leaves the cluster (one harmless failed network call per page load). Backend (Django) feature flags stay disabled in this mode. **Changing this list triggers a full Next.js rebuild** because the values are compile-time constants.

**Mode 3 — Disabled (default):**

Leave both `posthog_project_api_key` and `posthog_bootstrap_flags` empty. No PostHog client is initialized; AI gates stay closed. This is the lowest-friction starting state.

If `posthog_project_api_key` is set, it always wins — bootstrap flags are ignored.

````

**Review by mbertrand — COMMENTED** [review]
Another review for PostHog feature-flag wiring:  these feature flags need to be set for learn-ai "AskTIM" functionality to be available in the mit-learn front end.  The proposed changes add three modes (real keys / bootstrap flags / disabled) through tilt_config.json, the Next.js build, and optional Django ConfigMaps. Suggestions inline for each change

File: local-dev/README.md [hunk 34]
```diff
@@ -0,0 +1,733 @@
...
+- Increase Docker Desktop memory to 10+ GB
+- Or use a prebuilt image by removing `mit-learn` from `enabled_apps` in `tilt_config.json` and letting Tilt use the `prebuilt_tags` value instead
+
+### macOS: Port 5000 already in use
+
+On macOS, AirPlay Receiver listens on port 5000 by default. If you see `Address already in use` during cluster creation:
+
+**Option 1 (recommended):** Disable AirPlay Receiver in System Settings → General → AirDrop & Handoff → disable "AirPlay Receiver".
+
+**Option 2:** Modify the k3d registry port in `local-dev/cluster/k3d-config.yaml`:
```
ChristopherChudzicki: @daniellefrappier18 Mentioned this in https://github.com/mitodl/ol-infrastructure/pull/4495#issuecomment-4374015142, but  port 5000 if it is commonly use on MacOS for system stuff. We shouldn't use it by default.

Additionally, Option 2 doesn't make sense. The file is committed. Non-issue, hopefully, once we use a different port.

local-dev/README.md
```diff
@@ -0,0 +1,733 @@
...
+
+### `tilt up` fails on `local-infra` (Pulumi errors)
+
+```bash
+cd local-dev/infra
+pulumi up --logtostderr -v=3   # Verbose output
+```
+
+Common causes:
+- **kubeconfig not set:** Ensure `k3d kubeconfig merge local-dev --kubeconfig-merge-default` has been run, or set `ol-local-dev-infra:kubeconfig` in `Pulumi.local-dev.infra.Dev.yaml`.
```
ChristopherChudzicki: > ...or set `ol-local-dev-infra:kubeconfig` in `Pulumi.local-dev.infra.Dev.yaml`.

I don't think that makes sense. The `Pulumi.local-dev.infra.Dev.yaml` file is committed, right? So we wouldn't edit it.

ChristopherChudzicki (changes requested):
@blarghmatey Finally got a chance to try this out today. Ran into a few issues: the port 5000 thing that has been mentioned by @daniellefrappier18 and a pulumi error. I got the following pulumi error from `➜ ./local-dev/scripts/setup.sh`

```
 Manage your Pulumi stacks by logging in.                                                      
  Run `pulumi login --help` for alternative login options.                                     
  Enter your access token from https://app.pulumi.com/account/tokens                           
      or hit <ENTER> to log in using your browser   
```
AI's summary of the cause of ☝️
> setup.sh still runs Pulumi from the old local-dev/infra path and initializes the obsolete local-dev.infra.Dev stack, so it misses the new split projects’ file-backend config and falls back to Pulumi Cloud login.



 Suggestion: It would help if start.sh ended with a note like “When you’re done, run ./local-dev/scripts/stop.sh”. Stopping `./local-dev/scripts/start.sh` or exiting `tilt up` does not stop the local-dev cluster or its running workloads.

local-dev/infra/core/__main__.py
```diff
@@ -0,0 +1,149 @@
...
+
+create_ai_services(_k8s, namespaces["local-infra"], None, _infra_dir)
+
+messaging = create_messaging(_k8s, namespaces["local-infra"])
+
+# Create database cluster (needed by Keycloak instance) before identity
+db = create_database(_k8s, namespaces["local-infra"], cnpg_version)
+
+# Deploy Keycloak operator and instance (but not realm — that's in apps-infra)
+identity = create_identity_core(
```
mbertrand, line 167: Still hitting the `sso.ol.mit.dev` in-cluster DNS problem described [here](https://github.com/mitodl/ol-infrastructure/pull/4495#discussion_r3228239295).

local-dev/apps/mit-learn/deployment.yaml
```diff
@@ -0,0 +1,406 @@
...
+        resources:
+          requests:
+            cpu: 100m
+            memory: 512Mi
+          limits:
+            memory: 2Gi
+        volumeMounts:
+        - name: static-files
+          mountPath: /src/staticfiles
+        - name: mkcert-root-ca
```
mbertrand: Still seeing the CA bundle replacement problem described [here](https://github.com/mitodl/ol-infrastructure/pull/4495#discussion_r3228693716).

tilt_config.json.example
mbertrand, line 18: AskTIM not showing up in learn due to PostHog issue described [here](https://github.com/mitodl/ol-infrastructure/pull/4495#discussion_r3229030048).

PR #4503, opened 2026-04-16
chore: cleanup heroku references in micromasters pulumi code.

Ardiea:
chore: cleanup heroku references in micromasters pulumi code.

### DIFF
--- a/src/ol_infrastructure/applications/micromasters/__main__.py
+++ b/src/ol_infrastructure/applications/micromasters/__main__.py
@@ -15,12 +15,10 @@ import pulumi
 import pulumi_fastly as fastly
 import pulumi_kubernetes as kubernetes
 import pulumi_vault as vault
-import pulumiverse_heroku as heroku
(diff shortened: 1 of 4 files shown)


src/ol_infrastructure/applications/micromasters/__main__.py
```diff
@@ -280,186 +274,6 @@
...
-heroku_app_id = heroku_config.require("app_id")
-micromasters_heroku_configassociation = heroku.app.ConfigAssociation(
-    f"micromasters-{stack_info.env_suffix}-heroku-configassociation",
-    app_id=heroku_app_id,
-    sensitive_vars=sensitive_heroku_vars,
-    vars=heroku_vars,
-)
-
-
 if micromasters_config.get_bool("deploy_k8s"):
```
blarghmatey: It seems like we should remove this conditional as part of this change.

blarghmatey (approved):

Issue #4504, opened 2026-04-16
Mailgun under pulumi management


shaidar:
### Description/Context

The Mailgun domains and Route53 DNS records across all environments (CI, QA/RC, Production — ~20 domains total) were created manually and are not under Pulumi management. This means changes are made ad-hoc outside of version control, there is no drift detection, and there is no consistent record of what DNS records or SMTP credentials exist per domain.

We need to:
1. Bring all existing Mailgun domains, SMTP credentials, and Route53 DNS records into Pulumi state via a one-time import
2. Refactor the `mailgun` Pulumi project (`src/ol_infrastructure/applications/mailgun/`) to manage multiple domains per stack from a structured config list
3. Update each stack's YAML config to declare all domains with their per-domain settings and encrypted credentials

#### Phase 1 — One-time import script

A script `src/ol_infrastructure/applications/mailgun/import_existing.py` reads a per-environment JSON domain list, resolves Route53 zone IDs via boto3, and runs `pulumi import` for each existing resource per domain:

| Resource | Pulumi type | Import ID format |
|---|---|---|
| `mailgun.Domain` | `mailgun:index/domain:Domain` | `us:{domain}` |
| `mailgun.DomainCredential` (one per login) | `mailgun:index/domainCredential:DomainCredential` | `us:{login}@{domain}` |
| SPF TXT record | `aws:route53/record:Record` | `{zone_id}_{domain}_TXT` |
| DKIM TXT record | `aws:route53/record:Record` | `{zone_id}_mx._domainkey.{domain}_TXT` |
| MX record | `aws:route53/record:Record` | `{zone_id}_{domain}_MX` |
| Tracking CNAME | `aws:route53/record:Record` | `{zone_id}_email.{domain}_CNAME` |
| DMARC TXT record | `aws:route53/record:Record` | `{zone_id}__dmarc.{domain}_TXT` |

Usage (run from mailgun project dir with the target stack already selected):
```bash
pulumi stack select applications.mailgun.Production
uv run python import_existing.py --config domains-production.json --dry-run  # preview
uv run python import_existing.py --config domains-production.json             # execute
```

Repeat for QA and CI stacks. After each run, `pulumi stack --show-urns` should list all imported resources.

#### Phase 2 — Refactor __main__.py

Replace the current single hardcoded `mitlearn` domain with a loop over `mailgun_config.require_object("domains")`. Key points:

- Resource logical names must exactly match what the import script registered, e.g.:
[...]

PR #4505, opened 2026-04-16
Mailgun management under Pulumi

shaidar:

https://github.com/mitodl/ol-infrastructure/issues/4504

### DIFF
--- a/src/ol_infrastructure/applications/mailgun/__main__.py
+++ b/src/ol_infrastructure/applications/mailgun/__main__.py
@@ -13,9 +13,7 @@ from ol_infrastructure.lib.mailgun_helper import (
     mailgun_credential_opts,
     mailgun_domain_opts,
 )
-from ol_infrastructure.lib.pulumi_helper import parse_stack
 
-stack_info = parse_stack()
 mailgun_config = Config("mailgun")
 ol_mailgun_config = Config("ol_mailgun")
 api_key = mailgun_config.require("apiKey")
--- a/src/ol_infrastructure/applications/mailgun/__main__.py
+++ b/src/ol_infrastructure/applications/mailgun/__main__.py
@@ -17,8 +17,9 @@ from ol_infrastructure.lib.pulumi_helper import parse_stack
 
 stack_info = parse_stack()
 mailgun_config = Config("mailgun")
+ol_mailgun_config = Config("ol_mailgun")
 api_key = mailgun_config.require("apiKey")
-domains: list[dict[str, Any]] = mailgun_config.require_object("domains")
+domains: list[dict[str, Any]] = ol_mailgun_config.require_object("domains")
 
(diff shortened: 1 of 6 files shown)


src/ol_infrastructure/lib/aws/route53_helper.py
```diff
@@ -85,6 +85,55 @@ def zone_opts(domain: str) -> pulumi.ResourceOptions:
...
+    if records:
+        existing = records[0]
+        if (
+            existing["Name"].rstrip(".") == record_name.rstrip(".")
+            and existing["Type"] == record_type
+        ):
+            return pulumi.ResourceOptions(
+                import_=f"{zone_id}_{record_name}_{record_type}",
+                delete_before_replace=True,
+            )
```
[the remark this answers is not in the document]
shaidar: This is a **false positive**. Here's why:

Pulumi's `import_` in `ResourceOptions` is a conditional hint, not an unconditional directive. Pulumi only acts on `import_` when the resource **is not already tracked in the state file**. On subsequent runs where the resource is already in state, Pulumi ignores the `import_` field entirely and proceeds with a normal refresh/update cycle.

This means:
- **Run 1** (`managed: false`, record exists in Route53, not in state): `import_` fires → record enters Pulumi state.
- **Run 2+** (`managed: false`, record now in state): boto3 finds the record, returns `import_=...` again, but Pulumi sees the resource is already tracked and skips the import. No error.

The `managed: true` flag is a **performance optimization** (avoids a boto3 API call per record per run), not a correctness requirement. The stack won't break if it stays `false`.

The design is intentional and valid — it mirrors the same pattern used in `zone_opts()` in this file for hosted zones, and the `mailgun_domain_opts()`/`mailgun_credential_opts()` helpers in mailgun_helper.py.

src/ol_infrastructure/applications/mailgun/__main__.py
```diff
@@ -1,84 +1,132 @@
...
+        open_tracking=domain_cfg.get("open_tracking", True),
+        region=region,
+        smtp_password=pulumi.Output.secret(domain_cfg["smtp_password"]),
+        spam_action=domain_cfg.get("spam_action", "disabled"),
+        use_automatic_sender_security=domain_cfg.get(
+            "use_automatic_sender_security", True
+        ),
+        web_scheme=domain_cfg.get("web_scheme", "https"),
+        wildcard=domain_cfg.get("wildcard", False),
+        opts=mailgun_domain_opts(name, api_key, managed=managed, region=region),
```
blarghmatey, line 58: Is the intent to run this and then manually toggle the 'managed' flag for each stack layer?

blarghmatey (approved):

src/ol_infrastructure/applications/mailgun/__main__.py
shaidar: Yes, that's the theory

PR #4509, opened 2026-04-17
Make STS header last longer

ChristopherChudzicki:
Related to
- https://github.com/mitodl/hq/issues/10971

This PR increases expiration age on the [Strict Transport Security](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Strict-Transport-Security) header, which instructors browsers to redirect HTTP to HTTPS automatically on future visits. Currently we do this for 5 minutes, but it would be good to have this set longer. 

Especially since a large chunk of our traffic is HTTP -> HTTPS redirects, something like 10-20%.
```sh
# synthetic responses... almost all 301 redirects
 curl -s -u "105204:${GRAFANA_TOKEN}" \
    "https://logs-prod-us-central1.grafana.net/loki/api/v1/query" \
    --data-urlencode 'query=sum by (response_status) (count_over_time({service="fastly",application="mit-learn"} | json | environment="production" |
  url=~"^/$" | response_state="HIT-SYNTH" [24h]))' \
    | jq '.data.result'

# total response count
  curl -s -u "105204:${GRAFANA_TOKEN}" \
    "https://logs-prod-us-central1.grafana.net/loki/api/v1/query" \
    --data-urlencode 'query=sum(count_over_time({service="fastly",application="mit-learn"} | json | environment="production" | url=~"^/$" [24h]))' \
    | jq '.data.result'
```

### How can this be tested?
Unsure


src/ol_infrastructure/applications/mit_learn/__main__.py
```diff
@@ -920,7 +920,7 @@
             action="set",
             destination="http.Strict-Transport-Security",
             name="Generated by force TLS and enable HSTS",
-            source='"max-age=300"',
+            source='"max-age=31536000"',
```
ChristopherChudzicki, line 923: we could add `[includeSubDomains](https://developer.mozilla.org/en- US/docs/Web/HTTP/Reference/Headers/Strict-Transport-Security#includesubdomains)`, too, probably? But the only subdomain we hit is api.learn.mit.edu, and we're hitting that ourselves, so should already be https

**Question 1:** Should we add includeSubDomains?

**Question 2:** All our other STS header max ages are set at 300 seconds, too. OCW similarly has about 10% HIT-Syntehtic (at least on the homepage). Any objection to setting the others to longer?

blarghmatey: I think we can leave out the include subdomains for now, though it should be harmless since we serve everything over HTTPS. 

I do think that extending the lifetime across our Fastly distributions is worthwhile.

blarghmatey (approved):

src/ol_infrastructure/applications/mit_learn/__main__.py
feoh: When I asked Claude about the implications of this change, its reply left me rather concerned.

Can you help me understand how this *doesn't* open us up to man in the middle and other attacks?

Claude's response:

https://claude.ai/share/89ecadf6-ce94-4ad2-8e5d-c487eac81f4d

ChristopherChudzicki: _Note: This improves MITM defenses because it eliminates a small subset of HTTP requests. The risk Claude mentioned to you is an operational failure mode. (See below.)_

I also had not heard of this header before, btw, until embarking on this quest. I think it's best understood with an example. Should probably have included this or something like it in the PR description...

1. User types http://learn.mit.edu (or follows a plain http link)
2. Fastly returns 301 redirect → https://learn.mit.edu
3. Browser retries over HTTPS
4. Fastly returns the HTTPS response with the HSTS header, which the browser now caches for `max-age`[^1]
5. **From then on** (up to `max-age`), any http://learn.mit.edu URL the browser encounters is rewritten to https:// client-side. The rewrite happens **without network requests**.
    - currently, this is only happening for 5 minutes, which is basically useless. If a user retries http://learn.mit.edu an hour later, they go through the whole redirect shenanigans again, 2 round trips to fastly instead of 1.

[^1]: The 301 response to the original http request actually includes the HSTS header also, but browsers won't honor it since it was returned over HTTP.

**The Risk:** The operational risk... is what happens if HTTPS breaks for our site? E.g., an expired certificate.
- Without HSTS (or with short-lived HSTS): users see a browser warning akin to ":warning: Your connection is insecure". Users *can* click through that if they want to.
- With (long) HSTS: The site is broken, period[^2]. Users cannot click through the insecure warning.

[^2]: Well, not quite. Power users can clear the HSTS list. E.g., in chrome via chrome://net-internals/#hsts

**Suggestion:** My understanding is that (A) our certificate renewal is pretty well automated and (B) "your connection is insecure" is a pretty crappy failure mode anyway. So I think ramping up to 1 year directly seems OK. 

**Mentioning for interest:** There is actually a bit farther we could go in enforcing HTTPS:
-  the `includeSubDomains` mentioned earlier is one option—that forces HTTPS on all subdomains, too
- the HSTS preload list... See https://hstspreload.org/ Basically, this is a list of guaranteed HTTPS sites maintained by Chromium project. Browser makers then hardcode that list of sites into new browser versions, and you get automatic client side redirects on the very first HTTP request, no 301 redirect required.
    - requires `includeSubDomains` and `max-age` at least one year

feoh: Ha! And tobias approved an hour ago. Thanks for the extra digging Chris! Very interesting and useful.

feoh (approved):

PR #4558, opened 2026-04-27
fix: Bump Sentry trace sample rate down

### LINKED ISSUE #4543: Reduce Sentry Span Usage & Monthly Nastygram
### Description/Context
At the start of every month we get warnings from Sentry that we've over run our Spans quota and that spans are being dropped.

We must fix this, especially since we're already doing OTEL metric collection in Grafana.

### Plan/Design
- [x] Identify top Sentry span generating projects
- [x] Mitigate our span overage by reducing sampling rate or turning them off entirely as appropriate

feoh:
To keep from hitting our span usage limit like we do in the first few days of every month.

Fixes #4543 

Reduces trace sample rate from 0.25 to 0.1 for MITxOnline and MIT-Learn.

### How can this be tested?
Push fix. Number go down? :)


blarghmatey (approved):

src/ol_infrastructure/applications/mit_learn/Pulumi.applications.mit_learn.Production.yaml
```diff
@@ -105,7 +109,7 @@ config:
     POSTHOG_TIMEOUT_MS: 1000
     SEE_API_ENABLED: "true"
     SENTRY_PROFILES_SAMPLE_RATE: "0.25"
```
blarghmatey: These should get dropped to 0.1 as well

feoh: Great. Wasn't sure since profiling is a different beast.

PR #4598, opened 2026-05-05
config: Toggle the olapps email theme to use keycloakify

blarghmatey:
https://github.com/mitodl/hq/issues/8794

Toggles on the keycloakify emails in production

### How can this be tested?

This is already set up on QA


pdpinch (approved):
LGTM! 

I approve this decision, but I have no idea if the syntax is correct.

PR #4615, opened 2026-05-11
fix: escape dollar-quote delimiters in Vault readonly role creation statement

quazi-h:
N/A

PR #4576 added a `DO $$...$$;` block to the `readonly` role creation statements in `src/ol_infrastructure/lib/vault.py` to grant `rds_iam` to the `read_only_role`. However, the statements are wrapped in Python `string.Template` objects and rendered via `.substitute()`, which converts every `$$` → `$`. This means Vault was storing:

```sql
DO $
BEGIN
    ...
END $;
```

PostgreSQL rejects a bare `$` as a dollar-quote delimiter with `ERROR: syntax error at or near "$" (SQLSTATE 42601)`, causing a 500 error whenever credentials are generated for the `readonly` role on any `postgres-*` Vault secrets engine (first observed on `postgres-xpro`).

Fix: use `$$$$` (four dollar signs) in the Template string so `.substitute()` renders them as `$$` — the valid PostgreSQL anonymous dollar-quote delimiter. This matches the pattern used elsewhere in the same file (`$$do$$` style).

### How can this be tested?
1. After `pulumi up` re-applies the updated role definition, navigate to any `postgres-*` secrets engine in the Vault UI (e.g. `vault-production.odl.mit.edu/ui/vault/secrets-engines/postgres-xpro/overview`) and generate credentials for the `readonly` role — it should succeed without a 500 error.
2. Alternatively, verify via CLI: `vault read postgres-xpro/creds/readonly`

### Additional Context
The other `DO` blocks in the same file correctly use the `$$do$$...$$do$$` tag style (which renders as `$do$...$do$` — valid PostgreSQL dollar-quoting). Only the anonymous `$$` style added in #4576 was affected.

### DIFF
--- a/src/ol_infrastructure/lib/vault.py
+++ b/src/ol_infrastructure/lib/vault.py
@@ -199,14 +199,16 @@ postgres_role_statements = {
             # The DO block is a no-op on non-RDS Postgres where the role doesn't exist.
             Template(
                 """
-                DO $$
+                DO
+                $$do$$
                 BEGIN
                     IF EXISTS (
                         SELECT FROM pg_catalog.pg_roles WHERE rolname = 'rds_iam'
                     ) THEN
                         GRANT rds_iam TO "read_only_role";
                     END IF;
-                END $$;
+                END
+                $$do$$;
                 """
             ),
             # Create the read-only user and put it into the read-only-role


src/ol_infrastructure/lib/vault.py
```diff
@@ -199,14 +199,16 @@
             # The DO block is a no-op on non-RDS Postgres where the role doesn't exist.
             Template(
                 """
-                DO $$
+                DO
+                $$do$$
```
quazi-h, line 203: Following the same convention for this logic that already exists in this file for other readonly roles (see lines 152-154).

Ardiea (approved):

feoh (approved):

PR #4618, opened 2026-05-12
set YOUTUBE_CONFIG_URL

abeglova:
needed for https://github.com/mitodl/mit-learn/pull/3308
This also changes the base open-video-data used for all environments from the mitopen  to main


abeglova:
I think it's ok to have this pinned to just the main branch so we can update that branch without having to make an ol_infrastructure pr

feoh (approved):

PR #4644, opened 2026-05-18
config: Enable Fastly shielding in production for OCW and Learn

blarghmatey:
N/A

Toggles on shielding in Fastly for production on Learn and OCW. This has been enabled in RC for several weeks and I haven't heard any issues from it.

### How can this be tested?

The same configuration is enabled on RC currently

### Additional Context

Enabling shielding will reduce the amount of bandwidth to the origin and should speed up content delivery to non-US learners


gumaerc (approved):
Approving based on some basic publishing testing I did in OCW RC. I didn't see any caching issues which is what I imagine could arise from this. We should keep an eye on production once this is deployed though, as we don't have the same amount of traffic in RC obviously, paying attention to cache busting and making sure it continues to work as expected.

PR #4647, opened 2026-05-19
Update secrets for mitxonline

dsubak:
https://github.com/mitodl/hq/issues/11324

Sets hubspot:private-api-token to the value of the access token of the [app in mitxonlinedev-2026](https://app.hubspot.com/private-apps/51475747/40039666) and the value of hubspot:uai-private-api-token to the value of the access token of the [app in xprodev-2026](https://app.hubspot.com/private-apps/51359745/37338098)

In RC right now, those values are both the same and reference access tokens which don't appear in either accounts - my hunch is thatits the value for one of the old sandboxes. In either event, these should point at two different accounts - the UAI token should go to the xpro account, while the other points at the mitxonline account.

State of RC as of this afternoon.
Screenshot 2026-05-19 at 3 20 54 PM


dsubak:
Confirmed with Collin that the values are correct, so this should be ready to go!

blarghmatey (approved):

PR #4789, opened 2026-06-18
fix: fix MKTG ROOT url for learn

marslanabdulrauf:
https://github.com/mitodl/hq/issues/11838

This pull request makes a minor update to the edX app Kubernetes config map generation logic by ensuring that the marketing root URL uses the correct domain. This change helps maintain consistency in domain usage across the application configuration.

- Configuration Update:
  * Changed the `MKTG_URLS["ROOT"]` value in the `_build_interpolated_config_dict` function to use the domain from `mit_learn_domain` instead of `marketing_domain` in `k8s_configmaps.py`.

### DIFF
--- a/src/ol_infrastructure/applications/edxapp/k8s_configmaps.py
+++ b/src/ol_infrastructure/applications/edxapp/k8s_configmaps.py
@@ -342,7 +342,7 @@ def _build_interpolated_config_dict(
                     f"https://{edxapp_config.require('mit_learn_api_domain')}/logout",
                 ],
                 "MKTG_URLS": {
-                    "ROOT": f"https://{marketing_domain}/",
+                    "ROOT": f"https://{edxapp_config.require('mit_learn_domain')}/",
                 },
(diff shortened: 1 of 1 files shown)


asadali145:
@Anas12091101 can this change impact the behavior of different links for learn and mitxonline courses? I guess we still have some logic for different footer and header URLs?

blarghmatey (approved):

pdpinch:
I'm 90% sure we want to do this, and now is the right time. But I'd like to know where that URL is actually used in the MFEs and the theme. Footer changes and Studio link changes are totally fine. I'd like to know what else it touches.

marslanabdulrauf:
I had a call with @asadali145 and confirmed that we are not using this `MKTG_URLS` in the MFEs and themes.

Footers are already using learn domain from [MKTG_URL_OVERRIDES](https://github.com/mitodl/ol-infrastructure/blob/main/src/ol_infrastructure/applications/edxapp/k8s_configmaps.py#L347)

PR #4806, opened 2026-06-22
Remove ProctorTrack default from MITx Residential

feoh:
N/A — follow-up from Rootly/Sentry alert investigation for MITx Residential production.

MITx Residential (`mitx` / `mitx-staging`) does not use ProctorTrack, but the generated Open edX secrets defaulted every non-xPRO deployment to `proctortrack` even when no ProctorTrack URL was configured.

This changes the generated `PROCTORING_BACKENDS` so ProctorTrack is only used as the default when a deployment explicitly configures a `proctortrack_url`. Residential deployments without that URL now default to the `null` proctoring backend, while deployments that do configure ProctorTrack continue to generate the ProctorTrack backend and credentials template.

### How can this be tested?
- `uv run pytest tests/ol_infrastructure/applications/edxapp/test_secrets_builder.py`
- `uv run mypy src/ol_infrastructure/applications/edxapp/secrets_builder.py tests/ol_infrastructure/applications/edxapp/test_secrets_builder.py`
- `EDXAPP_DOCKER_IMAGE_DIGEST=sha256:fd0e1539e018e1b172003cf7ef9061ec169d328cf7b4e4b0216b5f30cfc3333c pulumi preview --stack mitx.Production --diff` from `src/ol_infrastructure/applications/edxapp`
  - Expected diff includes `PROCTORING_BACKENDS.DEFAULT: proctortrack -> null` in `10-general-secrets-yaml`.
  - No resource deletes or replacements were planned for `mitx.Production`.

### Additional Context
`uv run pre-commit run --all-files` was attempted. Python formatting, ruff, mypy, YAML, shellcheck, and other relevant checks passed, but the full run failed on unrelated repository-wide hooks: missing local `packer` binary and pre-existing hadolint warnings in unchanged Dockerfiles.

### DIFF
--- a/tests/ol_infrastructure/applications/edxapp/test_secrets_builder.py
+++ b/tests/ol_infrastructure/applications/edxapp/test_secrets_builder.py
@@ -169,26 +169,27 @@ class TestBuildBaseGeneralSecretsDict:
         assert secrets["PROCTORING_BACKENDS"]["DEFAULT"] == "null"
 
     def test_proctoring_default_backend_mitx(self, mock_stack_info_mitx):
-        """Verify mitx uses proctortrack as default backend."""
+        """Verify MITx Residential uses null as default proctoring backend."""
         secrets = build_base_general_secrets_dict(
             stack_info=mock_stack_info_mitx,
             redis_hostname="redis.example.com",
             lms_domain="mitx.example.com",
         )
 
-        assert secrets["PROCTORING_BACKENDS"]["DEFAULT"] == "proctortrack"
+        assert secrets["PROCTORING_BACKENDS"]["DEFAULT"] == "null"
 
-    def test_proctortrack_url_configuration(self, mock_stack_info_mitx):
+    def test_proctortrack_url_configuration(self, mock_stack_info_mitxonline):
         """Test proctortrack URL configuration."""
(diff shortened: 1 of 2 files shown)


tests/ol_infrastructure/applications/edxapp/test_secrets_builder.py
```diff
@@ -169,26 +169,27 @@ def test_proctoring_default_backend_xpro(self, mock_stack_info_xpro):
...
-    def test_proctortrack_url_configuration(self, mock_stack_info_mitx):
+    def test_proctortrack_url_configuration(self, mock_stack_info_mitxonline):
         """Test proctortrack URL configuration."""
         proctortrack_url = "https://proctortrack.example.com"
         secrets = build_base_general_secrets_dict(
-            stack_info=mock_stack_info_mitx,
+            stack_info=mock_stack_info_mitxonline,
             redis_hostname="redis.example.com",
-            lms_domain="mitx.example.com",
+            lms_domain="mitxonline.example.com",
```
shaidar, line 187: Doesn't matter since this doesn't really resolve into anything, but not sure if it's a bit confusing having mitxonline in here

shaidar (approved):

PR #4807, opened 2026-06-22
config(edxapp): double RATELIMIT_RATE from 600/m to 1200/m

feoh:
N/A

Doubles the edxapp `RATELIMIT_RATE` setting from `600/m` to `1200/m` in [`src/ol_infrastructure/applications/edxapp/k8s_configmaps.py`](https://github.com/mitodl/ol-infrastructure/blob/main/src/ol_infrastructure/applications/edxapp/k8s_configmaps.py#L638).

`RATELIMIT_RATE` is the default per-IP throttle Django's `ratelimit` decorator applies to rate-limited views in edx-platform (login, password reset, etc.). Raising it from 600 to 1200 requests per minute relaxes that throttle by 2x.

### How can this be tested?
Run `pulumi preview` against an edxapp stack and confirm the only change to the rendered ConfigMap is the `RATELIMIT_RATE` value flipping from `600/m` to `1200/m`. After deploy, the new value should be visible in the lms/cms ConfigMap and applied on pod restart.


rhysyngsun (approved):
LGTM

PR #4824, opened 2026-06-25
Disable CI alerts

shaidar:
### Summary

This change updates CloudWatch alarm behavior so CI RDS alarms remain visible for monitoring but do not trigger notifications.

### What Changed

- Added support for non-notifying alarms by honoring the existing `enabled` flag in the CloudWatch alarm component.
- Updated the CI RDS monitoring profile to create alarms with notifications disabled (`enabled: false`).

### Outcome

- CI still surfaces RDS issues in CloudWatch (state changes and history are preserved).
- CI no longer sends SNS/Rootly alerts for these alarms.
- QA and Production alert behavior is unchanged.


Ardiea (approved):
Love it.

PR #4829, opened 2026-06-25
Secure Opik with Keycloak OIDC via APISIX

Ardiea:
N/A

Puts the self-hosted Opik (Comet LLM observability) deployment behind Keycloak authentication. Opik's open-source edition has **no native OIDC** (it's an Enterprise-only feature), so authentication is enforced by an authenticating proxy in front of it rather than by Opik itself.

Key changes:

- **APISIX OIDC instead of the Traefik Gateway.** Opik moves off `OLEKSGateway` onto the data-cluster APISIX gateway using the legacy `OLApisixRoute` (v2) path, matching the established pattern for OIDC apps (Airbyte/Dagster/Marimo). This is required because secret-backed OIDC relies on the APISIX `secretRef` mechanism, which only exists on the v2 `ApisixRoute` path (the Gateway-API `PluginConfig` drops `secretRef`). TLS is terminated per-app via an `ApisixTls` CRD (`OLCertManagerCert`).
- **UI vs API auth split.** The Opik web UI and its API share an origin, and the SPA calls `/api/*` from the browser using only the session cookie. A single bearer-only policy would 401 the UI's own API calls, so `/api/*` is split by the presence of a Bearer `Authorization` header:
  - `/api/*` **with** a Bearer token → `bearer_only` validation (SDK / programmatic clients), rejects rather than redirects.
  - `/api/*` **without** one → session (cookie) auth, so the browser SPA works after interactive login.
  - everything else → interactive authorization-code flow for human browsers.
- **Keycloak client in the `ol-platform-engineering` realm.** A `CONFIDENTIAL` `ol-opik-client` is created in `ol-platform-engineering` (alongside Airbyte/Grafana/Dagster/Concourse/Vault), with `service_accounts_enabled=True` for the bearer-token path. Credentials are published to Vault at `secret-operations/sso/opik` and synced into the cluster by the Vault Secrets Operator. The client lives in `ol-platform-engineering` rather than `ol-data-platform` deliberately: that realm authenticates via username/password + passkey and has no Touchstone SAML IdP, avoiding the realm's `want_assertions_encrypted=True` requirement that caused Keycloak "Invalid Requester" (unencrypted-assertion) failures.
- **Reusable component addition.** `OLApisixRouteConfig` gains an optional `exprs` field so legacy `ApisixRoute` rules can match on headers/query/cookies (used here for the Bearer-header split). Backward-compatible — only emitted into `match` when set.
- **Vault policy + config.** The Opik Vault policy grants read on `secret-operations/sso/opik`; the redirect-URI stack config is keyed under `ol-platform-engineering-opik-redirect-uris`.

The Vault secret resource is aliased to its previous `ol-data-platform` resource name so the realm move is a non-destructive in-place update of the shared Vault path rather than a create+delete that would leave it empty.

### How can this be tested?
Validated with `pre-commit` (ruff + mypy) and `pulumi preview` on both the `keycloak` CI and `opik` CI stacks.

Deploy order:
1. `pulumi up` the **keycloak** CI stack — creates the `ol-platform-engineering` Opik client and rewrites `secret-operations/sso/opik`.
2. `pulumi up` the **opik** CI stack — APISIX route (with TLS, shared plugins, OIDC), Vault sync, and the `/api` split.

Reviewer validation against CI:
- `curl -sI https://opik-ci.ol.mit.edu/` → `302` to `sso-ci.ol.mit.edu` (interactive OIDC).
- APISIX ingress-controller logs show no `failed to match service port` for `opik-frontend` (the backend port is numeric `5173`, matched by number not name).
- Browser login uses `ol-platform-engineering` realm credentials; after login the SPA's `/api/v1/session` returns `200` (session rule), while `curl -H "Authorization: Bearer <invalid>" .../api/v1/session` returns `401` from bearer validation.

### Additional Context
- Authorization is coarse: this enforces authentication (any `ol-platform-engineering` realm user); Opik OSS has no per-user RBAC.
- External SDK / programmatic ingestion must present a Keycloak access token (service-account client-credentials) as a Bearer header. In-cluster producers posting directly to the `opik-frontend` service bypass APISIX and are unaffected.

### DIFF
--- a/src/ol_infrastructure/applications/opik/__main__.py
+++ b/src/ol_infrastructure/applications/opik/__main__.py
@@ -1,4 +1,3 @@
-# ruff: noqa: E501
 """Deploy Opik (Comet's open-source LLM observability platform) on the data EKS cluster.
 
 Opik provides tracing, evaluation, and prompt management for LLM applications.
@@ -33,11 +32,18 @@ from ol_infrastructure.components.applications.eks import (
     OLEKSAuthBinding,
     OLEKSAuthBindingConfig,
 )
-from ol_infrastructure.components.aws.eks import (
-    OLEKSGateway,
-    OLEKSGatewayConfig,
-    OLEKSGatewayListenerConfig,
-    OLEKSGatewayRouteConfig,
+from ol_infrastructure.components.services.apisix import (
+    OLApisixOIDCConfig,
+    OLApisixOIDCResources,
+    OLApisixPluginConfig,
+    OLApisixRoute,
+    OLApisixRouteConfig,
+    OLApisixSharedPlugins,
+    OLApisixSharedPluginsConfig,
+)
+from ol_infrastructure.components.services.cert_manager import (
+    OLCertManagerCert,
+    OLCertManagerCertConfig,
 )
 from ol_infrastructure.components.services.vault import (
     OLVaultK8SSecret,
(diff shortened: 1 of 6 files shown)


blarghmatey (approved):

Ardiea:

PR #4858, opened 2026-06-30
fix(mit-learn): reduce search API tail latency

mbertrand:
Related to https://github.com/mitodl/hq/issues/5101

Three independent infra-side fixes for `/api/v1/learning_resources_search` tail latency (10–30s, occasional 502/499). The slow tail does not seem to be the API/search query — it's request queuing, autoscaling, and transmission:

- **nginx JSON gzip** — the API bypasses Fastly and the sidecar had no `gzip_types`, so JSON bodies (up to ~1.26 MB) shipped uncompressed.
- **Higher per-pod concurrency** — Granian `workers` 1→4 (config-tunable) and CPU request 250m→500m. Sync Django views under the `asginl` ASGI interface serialize to ~1 in-flight search per process, so a single homepage's ~7-call fan-out saturated the pods.
- **KEDA autoscaling** — scale the webapp on APISIX request-rate + p95 latency (with a CPU backstop) instead of CPU/memory alone. Search blocks on I/O, so CPU stayed low and the old HPA never added replicas while requests queued. Mirrors the existing edxapp KEDA pattern.

### How can this be tested?
- `pulumi -C src/ol_infrastructure/applications/mit_learn preview --stack mit_learn.QA`. Expect: webapp Deployment update (workers/CPU), nginx ConfigMap update, and the CPU/memory HPA replaced by a KEDA `ScaledObject` + `TriggerAuthentication`.
- Post-deploy: `curl -H 'Accept-Encoding: gzip' -sI <search-url>` returns `content-encoding: gzip`; `kubectl -n mitlearn get scaledobject` shows READY with the Prometheus triggers returning data.

### Additional Context
KEDA Prometheus query labels (route regex, `mitlearn-app` pod selector) were confirmed against live Grafana metrics. App-side fixes from the diagnosis (cache-stampede coalescing, SSR fan-out reduction) are out of scope here and tracked separately.

### DIFF
--- a/src/ol_infrastructure/applications/mit_learn/__main__.py
+++ b/src/ol_infrastructure/applications/mit_learn/__main__.py
@@ -9,7 +9,6 @@ from string import Template
 
 import pulumi_fastly as fastly
 import pulumi_github as github
-import pulumi_kubernetes as kubernetes
 import pulumi_qdrant_cloud as qdrant_cloud
 import pulumi_vault as vault
 from pulumi import (
@@ -31,6 +30,10 @@ from bridge.lib.magic_numbers import (
     ONE_MEGABYTE_BYTE,
 )
 from bridge.secrets.sops import read_yaml_secrets
+from ol_infrastructure.applications.mit_learn.k8s_autoscaling import (
+    build_webapp_keda_config,
+    create_webapp_trigger_auth,
+)
 from ol_infrastructure.applications.mit_learn.k8s_secrets import (
     create_mitlearn_k8s_secrets,
 )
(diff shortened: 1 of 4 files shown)


blarghmatey (approved):

src/ol_infrastructure/applications/mit_learn/__main__.py
```diff
@@ -1493,7 +1511,7 @@
         application_cmd_array=["uwsgi"],
         application_arg_array=["/tmp/uwsgi.ini"],  # noqa: S108
         granian_config=GranianConfig(
-            workers=1,
+            workers=mitlearn_config.get_int("granian_workers") or 4,
```
blarghmatey: Maybe just set the default to 2 and we can scale up if needed.

PR #4865, opened 2026-06-30
reduce batch size for production

shanbady:

pulling back the increased batch size we introduced in https://github.com/mitodl/ol-infrastructure/pull/4836

This PR reduces the batch size for production. It seems there are potentially some bad interactions with pod memory and retries that are once again spiking the embedding costs.

Screenshot 2026-06-30 at 6 46 31 PM

### How can this be tested?
We will have to deploy this and see the embedding usage normalize


mbertrand (approved):
👍

PR #4904, opened 2026-07-06
fix: kill zombie retry loops in xqwatcher, learn-ai, odl-video-service

blarghmatey:
Part of https://github.com/mitodl/hq/issues/12179 (ops-signal hygiene: kill zombie retry loops and broken telemetry, >50% of fleet error volume)

Root-causes three of the confirmed zombie-retry-loop offenders from the 2026-07 SOA runtime/Sentry audit:

- **xqwatcher on mitx-staging (Production/CI/QA)**: `xqueue_server_url` pointed at the LMS domain (`staging.mitx.mit.edu` etc.) instead of a real xqueue host, so the watcher was stuck in an infinite login-loop 404 — 2.80M error lines/7d, ~40% of the whole fleet's Loki error volume. `edxapp:xqueue_domain` for these same stacks is literally `unused-placeholder`, so the integration is already disconnected end-to-end; rather than patch a URL nothing downstream will ever call, this scales the watcher to zero for all three mitx-staging stacks. Also fixes a `config.get_int(...) or 1` bug that silently coerced an explicit `min_replicas: 0` back to `1`, and skips HPA creation when `min_replicas == 0` (the k8s API rejects `minReplicas: 0` for an HPA whose metrics are all Resource-type/CPU-memory).
- **learn-ai (Production/QA)**: `OTEL_EXPORTER_OTLP_PROTOCOL` is `http/protobuf` but both OTLP endpoints pointed at port `4317` — the gRPC-only port on the `grafana-k8s-monitoring-alloy-receiver` Service. Corrected to `4318` (the HTTP OTLP port), matching the working config already used by `mit_learn`/`mitxonline`. This `ConnectionResetError` was 516k Sentry events/30d = 99.6% of the learn-ai project's volume, drowning out every real error.
- **odl-video-service**: dropped the vestigial `techtv.mit.edu`/`ttv.mit.edu` entries from `ovs:domains`. Confirmed dead both in code (only `ovs:default_domain` is ever read; `route53_managed_domains` was already pared down to just `video.odl.mit.edu` back in 2022) and in the live cluster (no route/cert exists for either host) — pure leftover config, zero behavior change.

### How can this be tested?
Verified with `pulumi preview` against every affected stack — diffs are exactly the intended change with no unrelated drift:
- `xqwatcher`: `mitx-staging.{Production,CI,QA}` each show the HPA being deleted (scale-to-zero) and no other changes; confirmed the min_replicas-bug fix and HPA-skip logic don't affect unrelated stacks (`mitx.Production` previewed clean aside from an artifact of setting a synthetic docker tag for preview purposes).
- `learn_ai`: `Production`/`QA` both show only the `4317`→`4318` change on the two OTLP endpoint env vars.
- `odl_video_service`: `Production` previews with **zero** resource changes from the domain removal (confirms the field is truly unread).

Also confirmed live via `kubectl` (read-only production access) that `edxapp:xqueue_domain: unused-placeholder` is set for all three mitx-staging stacks, meaning xqwatcher's queues were never receiving real submissions regardless of the URL bug.

### Additional Context
Not addressed in this PR (filed as follow-ups against the ops-signal-hygiene task, see the linked issue for full detail):
- Legacy "open" (mit-open/open-discussions) OpenSearch 403s + Celery beat retry storm — needs a product decision on decommission vs. fix; the beat schedule lives in the app repo, not here.
- `meilisearch.courses.learn.mit.edu` 503s — confirmed live via kubectl to be a StatefulSet pod bouncing between nodes with an EBS multi-attach conflict on reschedule (self-healed in ~15 min during observation), not a stale Pulumi route. Needs node-pinning/PodDisruptionBudget hardening rather than a config one-liner.
- Vault `reddit`/`bootcamps` secrets-engine mount cleanup — deferred; Redash (the actual poller generating this noise) is being retired wholesale per direct instruction, so patching its templates now would be wasted effort.
- `odl-video-service` YouTube-upload `MissingSchema` bug, `ocw-next` frontend Sentry SDK crash, `ocw-studio`'s stuck per-course Concourse pipeline — all app-code or ops-tooling fixes that live outside this repo.


blarghmatey:
Both addressed in 758319f:
- Made `ignore_changes=["spec.replicas"]` conditional on `min_replicas > 0` for both the main and edxorg Deployments, so Pulumi actively enforces `replicas=0` (rather than ignoring drift forever) once the HPA is gone.
- Mirrored the `min_replicas > 0` guard onto `xqwatcher_edxorg_hpa`, matching the main HPA — a stack combining `edxorg_xqueue_enabled=true` with `min_replicas=0` would otherwise still fail.

Verified with `pulumi preview`: mitx-staging.{Production,CI,QA} now show `replicas: 1 => 0` explicitly, and mitxonline.Production (min_replicas=1, edxorg_xqueue_enabled=true) previews clean with no regression.

PR #4939, opened 2026-07-08
Import mit-office-of-digital-learning Sentry org into Pulumi

### LINKED ISSUE #4908: Bring Sentry Alerting and Other Configuration Under Pulumi
### Description/Context
Graduate away from ClickOps and make Pulumi code the source of truth for our Sentry configuration.

### Plan/Design
Sentry provider already exists. Should just be an import?

feoh:
## Summary

Fixes #4908

Bring the existing `mit-office-of-digital-learning` Sentry organization under Pulumi management as a new `ol-infrastructure-sentry` project, importing live configuration instead of recreating it.

- New Pulumi project: `src/ol_infrastructure/infrastructure/sentry` (single `Production` stack, `pulumiverse-sentry==0.0.9` provider).
- New generator CLI: `bin/import-sentry-config` — inventories the live Sentry org via its API, generates `__main__.py` + `sentry_imports.json` + `IMPORT_SUMMARY.md`, and is safe to re-run to regenerate the whole project deterministically.
- Sentry credentials stored via SOPS at `src/bridge/secrets/sentry/account.yaml` (`auth_token`, `organization`), never in Pulumi config or generated code.
- New Concourse pipeline entry (`sentry`) added to `simple_pulumi` meta/pipeline for `Production`-only deployment, matching the Rootly pattern.
- Per review feedback, organization membership (`SentryOrganizationMember`) is intentionally left out of Pulumi management entirely; members continue to be administered directly in Sentry, and the generator only reports the live member count as an informational note.

## What's imported and managed

364 resources imported and confirmed via `pulumi preview --refresh`: **366 unchanged, 0 changes** (364 Sentry resources + 2 provider resources — a genuine no-op).

| Resource | Count |
|---|---|
| organization | 1 |
| team | 15 |
| project | 19 |
| key (DSN) | 20 |
| github_repository | 176 |
| code_mapping | 109 |
| dashboard | 20 |
| issue_alert | 4 |

All imported resources use an explicit `sentry_opts = ResourceOptions(provider=sentry_provider, protect=True)` (or the issue-alert variant with `ignore_changes`), so nothing can be accidentally deleted via `pulumi destroy`.

## What's intentionally excluded (documented in `IMPORT_SUMMARY.md`)

`pulumiverse-sentry==0.0.9` has real modeling/API limitations discovered during import; excluded resources remain live and unmanaged in Sentry, not deleted:

- **44 organization members** — left out of Pulumi management per review feedback; membership is administered directly in Sentry.
- **6 metric alerts** — provider fails to unmarshal live alerts whose actions contain numeric `targetIdentifier` values.
- **1 dashboard** (`Learn Performance`, id 140561) — uses `error-events`/`spans` widget types that the provider schema doesn't support (only `discover`/`issue`/`metrics`).
- **6 issue alerts** (Rootly/OpsGenie-integration rules) — driven exclusively by the `NotifyEventSentryAppAction` Sentry-app action. Confirmed via direct API calls that Sentry's classic Rules API accepts `GET` but rejects `PUT` with `404` for these rules regardless of payload — a hard Sentry API restriction, not a provider bug.
- **5 non-GitHub repositories** — provider only supports GitHub-integrated repos.

Two additional caveats for what *is* managed:
- Dashboard widget/query IDs are provider-computed and omitted from generated code.
- Issue-alert `actions`/`conditions`/`filters` bodies are set via `ignore_changes` (Sentry's API returns normalized data that doesn't perfectly round-trip through the provider schema); `actionMatch` itself is still actively managed.

## Validation performed

- `ruff format` / `ruff check` — clean, zero new errors vs. repo baseline.
- `mypy` (via `--explicit-package-bases --namespace-packages --scripts-are-modules`, matching the actual `.pre-commit-config.yaml` invocation) — zero errors in touched files.
- `pre-commit run` / `pre-commit.ci` — all hooks pass, including `detect-secrets` (added `# pragma: allowlist secret` on generated `SentryKey` resource names and excluded the generated `sentry_imports.json` from secret scanning, since both structurally contain Sentry's public DSN key IDs, which are not credentials).
- CodeQL — clean. An earlier revision tripped `py/clear-text-logging-sensitive-data` because a local variable named `secrets` (from `secrets = read_yaml_secrets(...)`) matched CodeQL's generic sensitive-name heuristic and tainted the non-sensitive organization slug extracted from it; renamed the variable and isolated the informational `print()` calls into helper functions that only take plain `str`/`int`/`Path` arguments.
- `pulumi import --file sentry_imports.json --preview-only` then for real, followed by `pulumi up` to adopt the explicit provider and normalize a handful of `actionMatch` values — iterated until `pulumi preview --refresh` converged to a true no-op.
- SOPS secret decrypts correctly and contains only `auth_token` + `organization`.

## Operational notes for reviewers

- To refresh from live Sentry state later: `uv run python bin/import-sentry-config generate` (regenerates all three files deterministically, ~45-90s), then diff/review before applying.
- `pulumi import --file sentry_imports.json --preview-only` is safe to re-run at any time to sanity-check drift without mutating state.
- All resources are `protect=True`; removing one from Pulumi state (not Sentry) requires an explicit `pulumi state delete --force`.
- The 62 excluded resources (44 members, 6 metric alerts, 1 dashboard, 6 issue alerts, 5 non-GitHub repos) remain fully functional in Sentry; they're just outside Pulumi's control until `pulumiverse-sentry` addresses these provider limitations, Sentry's API changes, or (for members) a future decision to manage them.


blarghmatey:
I think we can leave out the explicit user management from the imports/pulumi code

feoh:
Good call — removed explicit `SentryOrganizationMember` management entirely (44 resources). Membership continues to be administered directly in Sentry; `bin/import-sentry-config` now only reports the live member count as an informational note in `IMPORT_SUMMARY.md`, and the 44 previously-imported member resources have been removed from Pulumi state (no change to live Sentry). Managed resource count is now 364 (down from 408), and `pulumi preview --refresh` remains a true no-op.

PR #4950, opened 2026-07-09
fix(apisix): disable premature OIDC session timeouts across all apps (hq#8416)

blarghmatey:
Fixes https://github.com/mitodl/hq/issues/8416

MIT Learn users were getting logged out after ~15-60 minutes even though sessions are intended to last ~2 weeks, with a silent (no Keycloak UI) re-login on the next click — proof the Keycloak SSO session was still alive but the APISIX/OIDC session envelope had already expired.

Root cause: `OLApisixOIDCConfig`/`OLApisixOIDCResources` in [`src/ol_infrastructure/components/services/apisix.py`](https://github.com/mitodl/ol-infrastructure/blob/main/src/ol_infrastructure/components/services/apisix.py) only ever set `session.cookie.lifetime`, which APISIX's `openid-connect` plugin maps to lua-resty-session's `absolute_timeout` only. `idling_timeout`/`rolling_timeout` were left unset and fell back to lua-resty-session's compiled-in defaults of 900s/3600s (verified against [`apisix/plugins/openid-connect.lua`](https://github.com/apache/apisix/blob/3.17.0/apisix/plugins/openid-connect.lua#L36-L52) and [`lua-resty-session` v4.1.5](https://github.com/bungle/lua-resty-session/blob/v4.1.5/lib/resty/session.lua) at the APISIX version this repo actually pins — chart `2.15.0` / appVersion `3.17.0`, per `src/bridge/lib/versions.py:39`). So the server-side session was silently invalidated well inside the intended 14-day window, even though the browser cookie (14-day Max-Age) and Keycloak SSO session (336h idle/max) were both still valid.

Changes:
1. Add `oidc_session_idling_timeout` / `oidc_session_rolling_timeout` (`NonNegativeInt | None`, default `None`) to `OLApisixOIDCConfig`, emitted as flat `session.idling_timeout` / `session.rolling_timeout` keys (the schema APISIX 3.17+ actually validates against — not nested under the deprecated `session.cookie`). Defaulting to `None` means every other caller of this shared component (mitxonline, learn-ai, dagster, jupyterhub, etc.) is unaffected until they opt in.
2. Set both to `0` for mit-learn's two `OLApisixOIDCResources` (`mitlearn-k8s`, `mitlearn-k8s-no-prefix`). `0` explicitly disables that specific lua-resty-session check (verified in source — `if idling_timeout ~= 0 then ... end`, not "expire immediately"), deferring entirely to the already-configured 14-day `absolute_timeout`. This restores the originally-intended session length without introducing any new, undecided idle-timeout security policy; whether idle sessions on shared/public machines should eventually get a shorter cutoff is a separate open product/security decision.

### How can this be tested?
- `pulumi preview --stack QA` against `src/ol_infrastructure/applications/mit_learn` shows exactly the intended diff: `+ idling_timeout: 0` / `+ rolling_timeout: 0` added under `session` on both mit-learn `ApisixRoute` resources, nothing else touched.
- After deploy, a reviewer can confirm by staying idle on mit-learn QA for >15 minutes (past the old `idling_timeout` default) and >60 minutes (past the old `rolling_timeout` default), then interacting again — the session should remain valid instead of silently re-authenticating.

### Additional Context
- Filed a related, pre-existing bug discovered while reading the actual 3.17.0 schema: mitxonline's `oidc_session_cookie_domain` is a silent no-op (emitted into a nested `session.cookie.domain` path that lua-resty-session 4.x no longer reads — it wants a flat `session.cookie_domain`). Deliberately left out of this PR to keep it focused on hq#8416; tracked separately.
- Complementary, independent tracks for this issue: a "remember me" checkbox in ol-keycloakify (browser-restart persistence), and a broader team decision on long-term idle-timeout policy.


blarghmatey:
Update: confirmed with product that the mit-learn policy (no idling/rolling override — rely on the full `absolute_timeout`) should generalize to every APISIX-fronted app on this shared component, not just mit-learn. Added `oidc_session_idling_timeout=0` / `oidc_session_rolling_timeout=0` to all remaining `OLApisixOIDCConfig` call sites: dagster, jupyterhub, celery-monitoring, mitxonline (both resources), openlit, marimo-data, learn-ai, airbyte, and opik. Each of these was exposed to the same premature-logout bug as mit-learn.

Verified with `pulumi preview --diff` against mit-learn QA, mitxonline QA, and opik CI — each diff is exactly the two new `session.idling_timeout`/`session.rolling_timeout` keys, nothing else touched.

PR #4983, opened 2026-07-13
set featured podcast via NEXT_PUBLIC_PODCASTS_FEATURED_LIST_LEARNINGPATH_ID

ahtesham-quraish:
https://github.com/mitodl/mit-learn/pull/3576

This var is needed for populating the featured podcasts in learn, more you can check with above PR that has its used

### DIFF
--- a/src/ol_infrastructure/applications/mit_learn_nextjs/Pulumi.Production.yaml
+++ b/src/ol_infrastructure/applications/mit_learn_nextjs/Pulumi.Production.yaml
@@ -20,6 +20,7 @@ config:
   nextjs:origin: "https://learn.mit.edu"
   nextjs:pod_count: 5
   nextjs:posthog_api_host: "https://ph.ol.mit.edu"
+  nextjs:podcasts_featured_list_learningpath_id: "116826"
   nextjs:posthog_api_key:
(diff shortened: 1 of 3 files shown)


pdpinch:
@ChristopherChudzicki @mbertrand how/does the naming of this setting align with other Learn settings? I'm thinkig in particular of the feature learning paths that are used on the unit pages and the home page.

Ardiea (approved):

ChristopherChudzicki:
@pdpinch 

> @ChristopherChudzicki @mbertrand how/does the naming of this setting align with other Learn settings? I'm thinkig in particular of the feature learning paths that are used on the unit pages and the home page.

My opinion is that the name is fine / clear. I

t is deliberately different from the featured lists on the home page:

- **Podcast featured list:** This is a specific learning path that the frontend queries, determined via env var. Uses `api/v1/learning_paths/{id}` to fetch the data. Simple, but tie to this env var.
    - if the env var is unset or the associated list does not exist / is empty, the frontend just hies it.
    - The actual list is this one, https://learn.mit.edu/learningpaths/116826, whose title is "Featured Podcasts"
        - Annoyingly, the frontend at least required us to set at least 1 topic, so I set it to "Engineering"
- **Homepage featured list:** this is a bespoke API that aggregates across all `unit` channels, grabs the featured list from each unit channel, and returns them sorted by order `[1st from ocw, 1st from mitxonline, 1st from xpro, ... 2nd from ocw, 2nd from mitxonline, ... ]`. The `api/v1/featured/` endpoint is heavily tied to channels (in addition to learning lists) and doesn't transfer easily to podcasts.

PR #4993, opened 2026-07-14
Upload Next.js source maps to Sentry via decoupled Concourse task

ChristopherChudzicki:
# Issue

- Closes https://github.com/mitodl/hq/issues/12330

### Description:

Adds a concourse task to upload sourcemaps to Sentry for MIT Learn NextJS code.

- `next build` already injects Sentry **Debug IDs** into every JS chunk + `.map` (no token, no network — that's `@sentry/nextjs`/`withSentryConfig` in https://github.com/mitodl/mit-learn/blob/af19656020f05dc67eee9f724a589b035102c888/frontends/main/next.config.js#L128).
- This adds a separate pipeline task that uploads those already-generated maps.
  - The RC/release image build unpacks its rootfs (`UNPACK_ROOTFS`).
  - A separate `sentry-cli sourcemaps upload` task reads the maps from that rootfs and uploads them with the token.

Notes:

- Skipped for CI — QA/prod run the RC image, so the RC upload already covers them.
- Does not block release. Task is a `try`
- Debug IDs (not release name) are what bind a map to its minified code, so `--release` is only for organizing the artifact bundle in the Sentry UI.

### How can this be tested?
Not sure. Deploy?

### Release Checklist

- [ ]  a Sentry org auth token must be provisioned at `((sentry.mitlearn_auth_token))` (scopes `org:read`, `project:releases`).

### DIFF
--- a/src/ol_concourse/pipelines/infrastructure/k8s_apps/pipeline.py
+++ b/src/ol_concourse/pipelines/infrastructure/k8s_apps/pipeline.py
@@ -24,6 +24,7 @@ from ol_concourse.lib.models.pipeline import (
     ResourceType,
     TaskConfig,
     TaskStep,
+    TryStep,
 )
 from ol_concourse.lib.resource_types import (
     github_deployments_resource,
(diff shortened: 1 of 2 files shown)


src/ol_concourse/pipelines/infrastructure/k8s_apps/pipeline.py
```diff
@@ -165,6 +193,12 @@ def validate_fastly_config(self) -> "AppPipelineParams":
             "production": "learn.mit.edu",
         },
         fastly_purge_scope="html-pages",
+        sentry_sourcemaps=SentrySourcemapsConfig(
+            org="mit-office-of-digital-learning",
+            project="open-next",
```
blarghmatey, line 198: Not a blocker, but hard-coding this will be problematic as we try to streamline/standardize our project naming. cc @feoh

blarghmatey (approved):

src/ol_concourse/pipelines/infrastructure/k8s_apps/pipeline.py
ChristopherChudzicki: renaming that project would be wonderful...

PR #5011, opened 2026-07-15
feat(grafana_alerting): route overnight warning alerts to Slack instead of paging

feoh:

Part of https://github.com/mitodl/ol-infrastructure/issues/5012

On-call has been getting paged overnight by warning-severity alerts that last only a few minutes and self-resolve (e.g. Grafana Synthetic Monitoring "response times 30% above normal", `mitlearn-5xx-error-percentage`). These should not wake a human in the middle of the night, but we still want visibility that a service spiked past threshold.

This changes the Grafana Alertmanager notification policy (`src/ol_infrastructure/infrastructure/grafana_alerting/alertmanager.py`) so that **warning-severity** alerts are routed by time of day:

- **Daytime (09:00–17:00 `America/New_York`)** → Rootly (pages), unchanged.
- **Overnight (17:00–09:00)** → a new `#devops-alerts` Slack contact point **only** — no page. Awareness is preserved; nobody is woken.
- **Critical-severity** alerts → Rootly, 24/7. Unchanged.

Implementation notes:

- Two complementary `MuteTiming` resources (`overnight-17-to-09-eastern`, `daytime-09-to-17-eastern`). Times are `America/New_York`-local so the window tracks EST/EDT automatically. The overnight window is split into `00:00–09:00` + `17:00–24:00` because Alertmanager time ranges cannot span midnight.
- The daytime Rootly route uses `continue_=True` so a warning is offered to both the Rootly route and the overnight Slack route; whichever is muted for the current time drops out, leaving exactly one active destination. This is the idiom shown in the pulumiverse_grafana provider's own `NotificationPolicy` example.

### Scope / follow-up

This policy only governs the **Grafana pipeline** — alert sources `grafana` and `alertmanager` (~65% of Rootly alert volume over the last 30 days). **CloudWatch** (SNS→Rootly) and **Sentry** (→Rootly) webhook Rootly directly and bypass Grafana Alertmanager, so they are **not** time-gated by this change. Gating those is a planned fast-follow (likely at the Rootly layer, the one chokepoint all sources share).

Note on threshold: the original request was "must be in alert 2 minutes before paging," but the data showed most overnight noise already lasts 5–6 minutes, so a 2-minute duration gate would not have helped. This severity + time-of-day approach targets the actual noise while keeping criticals paging.

### How can this be tested?

- `uv run ruff check` and `uv run mypy` on `alertmanager.py` — both pass. `pre-commit run --files ...` passes all hooks.
- `pulumi preview` was **not** run locally: it requires Grafana Cloud credentials and the new `slack_devops_alerts_overnight_api_url` secret (see checklist), which are only present in the deploy environment. Please validate via `pulumi preview` in the CI pipeline / a stack with the secret populated before merge.
- Post-deploy validation: in Grafana Cloud → Alerting → Notification policies, confirm the two warning routes and the two mute timings render as expected; send a test warning alert during each window and confirm it reaches Rootly (day) vs `#devops-alerts` (night), and that a critical still pages in both.

### Additional Context

Open question for reviewers: this implements the **literal** 5 PM–9 AM window **every day**, so weekend daytime warnings still page. If weekends should be treated as fully overnight (Slack-only), say so and I'll add a weekday constraint to the mute timings.

### Checklist:
- [ ] Add `slack_devops_alerts_overnight_api_url` (an incoming webhook for `#devops-alerts`, or the chosen channel) to each `src/bridge/secrets/grafana_cloud/api.<env>.yaml` and re-encrypt with SOPS **before** deploying — `pulumi up` will fail without it.
- [ ] Confirm the target Slack channel name (`#devops-alerts` is a placeholder default in the contact point `recipient`).

### DIFF
--- a/src/ol_infrastructure/infrastructure/grafana_alerting/CLAUDE.md
+++ b/src/ol_infrastructure/infrastructure/grafana_alerting/CLAUDE.md
@@ -2,6 +2,7 @@
 
 This Pulumi program manages alerting and uptime monitoring for MIT Open Learning.
 It replaces two legacy systems:
+
 - **Pingdom (manual)** → Pingdom checks managed via a Pulumi dynamic provider (`pingdom_checks.py`)
 - **grafana-alerts repo + cortextool** → Grafana-managed alert rules and Alertmanager config
 
@@ -31,6 +32,7 @@ set of Grafana-managed alert rules. The Pulumi stacks (CI, QA, Production) map
 1:1 to the Grafana Cloud stacks.
 
 Secrets files must be encrypted with SOPS before committing:
+
 ```
 sops --encrypt --in-place src/bridge/secrets/grafana_cloud/api.<env>.yaml
 ```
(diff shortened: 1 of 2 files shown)


feoh:
This is the wrong approach. Closing.

PR #5027, opened 2026-07-17
trigger a restart of ocw deployments when secret-ocw-studio changes

zawan-ila:
None

This PR triggers a rolling restart of the ocw deployments when the SECRET_KEY vault secret changes. It is meant to help prevent [this issue](https://mitodl.slack.com/archives/G1VJQHQHK/p1784210417723189?thread_ts=1784128739.074629&cid=G1VJQHQHK).

### How can this be tested?
- Once the change is on QA, examine the appropriate vault static secret `kubectl get  vaultstaticsecret ocw-studio-collected-static-secret -n ocw-studio -o yaml`
- Verify that its spec.rolloutrestarttargets property is appropriately set to ocw deployments

### DIFF
--- a/src/ol_infrastructure/applications/ocw_studio/__main__.py
+++ b/src/ol_infrastructure/applications/ocw_studio/__main__.py
@@ -567,6 +567,17 @@ vault_k8s_resources = OLVaultK8SResources(
 db_instance_name = f"ocw-studio-db-applications-{stack_info.env_suffix}"
 rds_endpoint = f"{db_instance_name}.cbnm7ajau6mi.us-east-1.rds.amazonaws.com:{DEFAULT_POSTGRES_PORT}"
 
+# All deployments that consume the secret-ocw-studio application secrets and
+# need a rolling restart when those secrets change. Names must match the
+# deterministic naming in OLApplicationK8s (application_name="ocw-studio").
+ocw_studio_restart_deployment_names = [
+    "ocw-studio-app",  # webapp
+    "ocw-studio-default-celery-worker",
+    "ocw-studio-publish-celery-worker",
+    "ocw-studio-batch-celery-worker",
+    "ocw-studio-celery-beat",
+]
+
 # Create Kubernetes secrets
 secret_names, secret_resources = create_ocw_studio_k8s_secrets(
     stack_info=stack_info,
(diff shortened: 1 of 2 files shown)


feoh (approved):

src/ol_infrastructure/applications/ocw_studio/__main__.py
```diff
@@ -567,6 +567,17 @@
 db_instance_name = f"ocw-studio-db-applications-{stack_info.env_suffix}"
 rds_endpoint = f"{db_instance_name}.cbnm7ajau6mi.us-east-1.rds.amazonaws.com:{DEFAULT_POSTGRES_PORT}"
 
+# All deployments that consume the secret-ocw-studio application secrets and
+# need a rolling restart when those secrets change. Names must match the
+# deterministic naming in OLApplicationK8s (application_name="ocw-studio").
+ocw_studio_restart_deployment_names = [
+    "ocw-studio-app",  # webapp
+    "ocw-studio-default-celery-worker",
+    "ocw-studio-publish-celery-worker",
```
feoh, line 576: Nit: this hardcodes the five deployment names as string literals instead of reusing `ocw_studio_k8s_app.all_deployment_names`, which #4966 added to `OLApplicationK8s` specifically so callers wouldnt have to duplicate the naming logic (see its docstring example for `OLVaultK8SDynamicSecretConfig.restart_targets`).

Looks like this is because `create_ocw_studio_k8s_secrets()` runs at line 571, before `OLApplicationK8s` is instantiated at line 593, so `all_deployment_names` isnt available yet at this point. Not a blocker, but worth a comment noting the ordering constraint so the two deployment-name lists dont silently drift if the celery worker/beat config changes later.

zawan-ila: Yup, I tried using all_deployment_names but the ordering made it somewhat hard. Then I noticed that nobody uses that, probably because of the same reason, and didn't bother putting more time into it.

The only other usage of restart_targets I could find is [here](https://github.com/mitodl/ol-infrastructure/blob/main/src/ol_infrastructure/applications/edxapp/k8s_resources.py#L104-L111), which also hardcodes the names.

PR #5034, opened 2026-07-17
fix(concourse): scope simple_pulumi self-build to the app's subdirectory

blarghmatey:
Follow-up to https://github.com/mitodl/ol-infrastructure/pull/5030, https://github.com/mitodl/ol-infrastructure/pull/5031, https://github.com/mitodl/ol-infrastructure/pull/5033

After deploying `release_bot`, the Kubernetes Deployment came up `0/1`
ready with the pod cycling `Completed` (exit 0) every ~30s and **zero log
output**:

```
NAME                                      READY   STATUS      RESTARTS   AGE
release-bot-production-6489dbf49b-nrqdj   0/1     Completed   4          2m
```

Pulled the actual `release-bot-production:latest` image locally to
investigate:

```
$ docker run --rm --entrypoint sh <repo>:latest -c 'ls /app; pip show slack-bolt; python3 --version'
ls: cannot access '/app': No such file or directory
WARNING: Package(s) not found: slack-bolt
Python 3.14.6
```

No `/app`, no `slack-bolt`, and Python **3.14.6** — not the `3.13-slim`
pinned in
[`applications/release_bot/Dockerfile`](https://github.com/mitodl/ol-infrastructure/blob/main/src/ol_infrastructure/applications/release_bot/Dockerfile).
The pushed image was actually built from the **repo root's own
`Dockerfile`** (the generic `ol-infrastructure` Python package image,
`FROM python:3.14-slim`), which has no `CMD` — so it falls back to the
base image's bare `python3`, which hits EOF on stdin in a pod and exits 0
immediately. Confirmed the `latest` tag's digest matches a build tagged
with an unrelated commit SHA (`#5025`, an ElastiCache alarm fix).

Root cause in
[`_build_docker_image_job()`](https://github.com/mitodl/ol-infrastructure/blob/main/src/ol_concourse/pipelines/infrastructure/simple_pulumi/pipeline.py):
it built `CONTEXT`/`DOCKERFILE` from `pulumi_code.name` alone — the git
resource's checkout root, i.e. the *whole repo* (`paths:` only gates what
triggers a new version, it doesn't scope the checkout) — and never joined
in `pulumi_project_path`. So `oci-build-task` silently found and built the
repo's own top-level `Dockerfile` instead of erroring. `release-bot` is
the only `pipeline_params` entry currently using the self-build
(`BuildConfig`) path, so this had never been exercised before.

Fix: scope `CONTEXT`/`DOCKERFILE` to
`PULUMI_CODE_PATH.joinpath(pulumi_project_path)`, the same join already
used for the Pulumi deploy job elsewhere in this file
(`project_source_path=PULUMI_CODE_PATH.joinpath(params.pulumi_project_path)`).

### How can this be tested?
Regenerated the release-bot pipeline definition locally and confirmed the
build task's params now point at the right subdirectory:

```
CONTEXT:    ol-infrastructure-pulumi-release-bot/src/ol_infrastructure/applications/release_bot
DOCKERFILE: ol-infrastructure-pulumi-release-bot/src/ol_infrastructure/applications/release_bot/Dockerfile
```

`pre-commit` (ruff format/check, mypy) passes.

### Additional Context
Once this merges, `build-release-bot-image` needs to be re-triggered in
Concourse to push a correctly-built image before the k8s Deployment will
come up healthy -- the currently-running pod will keep restarting on the
bad image until that happens.

### DIFF
--- a/src/ol_concourse/pipelines/infrastructure/simple_pulumi/pipeline.py
+++ b/src/ol_concourse/pipelines/infrastructure/simple_pulumi/pipeline.py
@@ -232,7 +232,7 @@ def _ensure_ecr_repository_step(ecr_registry_image_resource: Any) -> TaskStep:
 
 def _build_docker_image_job(
     app_name: str,
-    pulumi_code: Any,
+    build_source: Any,
     docker_image_resource: Any,
     build_config: BuildConfig,
     pulumi_project_path: str,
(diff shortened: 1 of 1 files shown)


blarghmatey:
Pushed a follow-up commit addressing the same-day risk: the build job's trigger was riding on \`pulumi_code\`'s broad \`PULUMI_WATCHED_PATHS\` (shared \`lib/\`, \`components/\`, and *every* app's \`src/bridge/secrets/\`), which is literally how the original bad image got pushed as \`latest\` -- an unrelated commit (#5025) touching \`components/aws/cache.py\` triggered it.

Gave \`build-release-bot-image\` its own dedicated git resource (\`release-bot-docker-source\`) scoped only to \`applications/release_bot/\` + its own secrets folder. The deploy chain's \`pulumi_code\` resource is unchanged. Verified all 35 \`pipeline_params\` apps still regenerate cleanly.

PR #5107, opened 2026-07-23
Add MITOL_API_BASE_URL to mit-learn

rhysyngsun:

N/A

It looks like we've had this setting but it was unused since we got rid of direct OIDC login from the django app, but the variable was [still in `settings.py`](https://github.com/mitodl/mit-learn/blob/main/main/settings.py#L235) so I though we were setting it in deployments. The TLDR is that certain operations, particularly celery tasks, need to be able to generate an absolute URL to the server when they don't have an existing request from which to derive the domain.

### How can this be tested?

Only way to test this is by sending a specific email from a test system, so we should deploy to RC and test it there.


blarghmatey (approved):
It's worth double checking on Copilot's comment

PR #5119, opened 2026-07-27
fix: Set the proper env var for MITx Online to pick up the Fastly settings

blarghmatey:
Closes https://github.com/mitodl/hq/issues/12381

Aligns the keys set by the deployment with the keys that the app is reading

### How can this be tested?

Deploy and try updating an article to ensure that the cache purges and the content updates

### Additional Context

Fastly purging from the app is failing and it looks like it's because there's a mismatch between the keys being set and the keys that the app is expecting for the Fastly settings.

### DIFF
--- a/src/ol_infrastructure/applications/mitxonline/k8s_secrets.py
+++ b/src/ol_infrastructure/applications/mitxonline/k8s_secrets.py
@@ -226,9 +226,9 @@ def create_mitxonline_k8s_secrets(
             "base_name": "collected",
             "path": "collected-static-secrets",
             "templates": {
-                "FASTLY_API_KEY": '{{ index .Secrets "fastly" "api_key" }}',
-                "FASTLY_AUTH_TOKEN": '{{ index .Secrets "fastly" "api_key" }}',
-                "FASTLY_SERVICE_ID": '{{ index .Secrets "fastly" "service_id" }}',
+                "MITX_ONLINE_FASTLY_API_KEY": '{{ index .Secrets "fastly" "api_key" }}',
+                "MITX_ONLINE_FASTLY_AUTH_TOKEN": '{{ index .Secrets "fastly" "api_key" }}',
+                "MITX_ONLINE_FASTLY_SERVICE_ID": '{{ index .Secrets "fastly" "service_id" }}',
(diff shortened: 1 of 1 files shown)


ChristopherChudzicki:
The env vars were named correctly before this PR, in the sense that they matched what `get_string(name=...)` actually reads. They're now renamed to the Django *setting* names. So the settings now evaluate to `None`.

The accurate env var names, as of now, are:

- [`FASTLY_AUTH_TOKEN`](https://github.com/mitodl/mitxonline/blob/39798d2cc180e686d6d623957ceadac6840ead59/main/settings.py#L1354)
    - ready by `settings.MITX_ONLINE_FASTLY_AUTH_TOKEN`
- [`FASTLY_URL`](https://github.com/mitodl/mitxonline/blob/39798d2cc180e686d6d623957ceadac6840ead59/main/settings.py#L1360)
    - ready by `MITX_ONLINE_FASTLY_URL`
- [`FASTLY_SERVICE_ID`](https://github.com/mitodl/mitxonline/blob/39798d2cc180e686d6d623957ceadac6840ead59/main/settings.py#L1366)
    - ready by `MITX_ONLINE_FASTLY_SERVICE_ID`

The env var names are bad, but two things went wrong here:
-  the rename picked the wrong axis (the ambiguity worth fixing is *whose* Fastly service, not which app reads the var)
- and it changed only one side of a two-sided contract. `k8s_secrets.py` and `settings.py` have to move together or the read silently breaks.

The actual root cause of https://github.com/mitodl/hq/issue/12381 was never the names, though:

1. MITxOnline had the wrong **value**—it was purging the MITxOnline Fastly *service*, not Learn's. MITxOnline's own service caches no HTML and sets no surrogate keys, so the purge was a 200 OK that invalidated nothing.
2. There IS a related naming issue. No live code path uses these settings except purging Learn pages, so `FASTLY_SERVICE_ID` reads as "our service" when the value must be Learn's. Renaming it `LEARN_FASTLY_SERVICE_ID` makes that unmistakable.

    - I'd leave `FASTLY_AUTH_TOKEN` and `FASTLY_URL` alone — the token is an account-level credential (also read by the dead `call_fastly_purge_api`, which targets MITxOnline's own service), and `FASTLY_URL` is just the API base.

PR #5154, opened 2026-07-29
fix(concourse): point MFE feedback URLs at the mit-learn API domain

zamanafzal:
https://github.com/mitodl/hq/issues/11629

Points the MFE content-feedback URLs (`FEEDBACK_SUBMIT_URL`, `FEEDBACK_CSRF_PRIME_URL`) at the mit-learn API domain (`api.<env>.learn.mit.edu`, APISIX) via a new `mit_learn_api_url` var, instead of the frontend domain which 308-redirects `/api/...` and breaks the CORS preflight.

### How can this be tested?
After the pipeline runs, the preflight `OPTIONS` on `FEEDBACK_SUBMIT_URL` returns 200 (not 308).


asadali145 (approved):
LGTM!

PR #5214, opened 2026-08-03
Add zizmor GitHub Actions security scanning

blarghmatey:
N/A

Hardens GitHub Actions supply-chain security in two ways, and documents a third change that was attempted but had to be deferred:

1. **Adds `.github/workflows/actions-static-analysis.yml`** — a new CI workflow that runs [zizmor](https://github.com/zizmorcore/zizmor), a static analysis tool for GitHub Actions workflows, on push/PR whenever files under `.github/workflows/**` change. It uses the pinned `zizmorcore/zizmor-action` (pinned to a commit SHA, `v0.6.2`) with `min-severity: high` / `min-confidence: medium`, scoped to read-only `contents`/`actions` permissions. This mirrors the pattern already in use in the sibling repo `mitodl/mitxonline`.

2. **Adds a `zizmor` pre-commit hook** (`zizmorcore/zizmor-pre-commit`, `rev: v1.29.0`) to `.pre-commit-config.yaml`, placed next to the existing `actionlint` hook, so the same workflow-security linting runs locally before commit, not just in CI.

3. **Attempted but deferred: a 7-day dependency cool-down.** The plan was to add a root-level `[tool.uv]` table with `exclude-newer = "7d"` to `pyproject.toml` (uv 0.11+ supports a rolling relative-duration string here, stored as `exclude-newer-span` in the lockfile, so it keeps rolling forward rather than freezing to today's date) and regenerate `uv.lock`. This is meant to hold back newly-published package versions for 7 days to reduce exposure to newly-introduced malicious or compromised releases.

   However, `uv lock` failed outright with this workspace's current dependency pins:

   ```
   Resolving despite existing lockfile due to addition of global exclude newer ...
     × No solution found when resolving dependencies for split (markers:
     │ python_full_version == '3.14.*'):
     ╰─▶ Because ol-concourse==0.10.0 was published after the exclude newer
         time and only ol-concourse<=0.8.12 is available, we can conclude that
         ol-concourse>=0.9.0,<0.11 cannot be used.
         And because ol-infrastructure depends on ol-concourse>=0.9.0,<0.11 and
         your workspace requires ol-infrastructure, we can conclude that your
         workspace's requirements are unsatisfiable.

   hint: `ol-concourse` was filtered by `exclude-newer` to only include packages
   uploaded before the cutoff. The latest version satisfying the requirement is
   v0.10.0, published within the exclusion window. Consider using
   `exclude-newer-package` to override the cutoff for this package.
   ```

   The root `pyproject.toml`'s own `ol-concourse>=0.9.0,<0.11` constraint currently resolves only to `0.10.0`, which was published inside the proposed 7-day window — there's no earlier release in the `0.9.x`/`0.10.x` range to fall back to. Rather than force a workaround (bump the pin, add a per-package `exclude-newer-package` exemption, or shorten the window), that decision is left for a follow-up so the resolution strategy for `ol-concourse` can be made deliberately. This PR therefore contains only changes 1 and 2; no `pyproject.toml`/`uv.lock` changes are included.

**Note for reviewers:** this is a large uv workspace with many members. If a follow-up PR does add `exclude-newer`, expect the `uv.lock` diff to touch many resolved package versions (rolling back to whatever the cutoff was at generation time) — that's expected and intended, but worth a sanity-check pass rather than a rubber-stamp.

N/A — no UI changes.

### How can this be tested?
- Validate both new/changed files parse as YAML:
  ```
  python3 -c "import yaml; yaml.safe_load(open('.github/workflows/actions-static-analysis.yml'))"
  python3 -c "import yaml; yaml.safe_load(open('.pre-commit-config.yaml'))"
  ```
- Run the new hook locally against the new workflow file:
  ```
  pre-commit run zizmor --files .github/workflows/actions-static-analysis.yml
  ```
  This was run locally as part of preparing this PR and passed (`zizmor...Passed`), confirming the pre-commit hook environment resolves and the new workflow file itself is clean by zizmor's own standards.
- On merge, any future push/PR touching `.github/workflows/**` will trigger the new `GitHub Actions Static Analysis` workflow; confirm it shows up as a check and passes.

### Additional Context
The `exclude-newer` dependency cool-down (item 3 above) is intentionally not included here — see the Description for the exact `uv lock` failure and why it was deferred rather than forced through. Happy to follow up with that once there's a decision on how to handle the `ol-concourse` pin (bump the lower bound, add `exclude-newer-package` for it, or otherwise).


blarghmatey:
Re-added the `exclude-newer = "7d"` dependency cool-down. The earlier blocker was that `ol-concourse>=0.9.0,<0.11` only has one version in range (0.10.0), and it was published inside the 7-day window with no older fallback to resolve to.

Fixed by adding a `[tool.uv.exclude-newer-package]` table that exempts `ol-concourse` and every other package we publish from mitodl/django-aqueduct, mitodl/open-edx-plugins, and mitodl/ol-django (exclude-newer = "0d", i.e. always resolve to latest for those packages). The 7-day cooldown is meant to guard against newly-published *third-party* supply-chain risk, not slow down consumption of our own in-house releases, so this keeps that protection for everything else while unblocking our own packages.

`uv lock` now succeeds — 155 packages resolved, `ol-concourse` resolves to 0.10.0 as expected, and several third-party packages rolled back to versions outside the 7-day window.

PR #5219, opened 2026-08-03
fix(apisix): give MIT Learn's gateway session cookie an environment-scoped name

blarghmatey:
APISIX's openid-connect plugin (lua-resty-session 4.x) names its session
cookie "session" by default, which every OIDC-protected application here was
using unchanged. That is a poor fit for two reasons:

- The name is generic enough to collide with any other "session" cookie a
  *.mit.edu host sets, and says nothing about which system owns it when
  reading a Cookie header during an oversized-header investigation.
- Several of these cookies are deliberately scoped to a parent domain
  (.learn.mit.edu, see #5182; .mitxonline.mit.edu) so they reach sibling
  subdomains, which means the Production cookie is also sent to the RC and CI
  hosts under that parent. Under one shared name all three environments'
  cookies stack up on those hosts, where each gateway can be handed a session
  envelope encrypted with a different environment's secret.

## Naming

Adds `OLApisixOIDCConfig.oidc_session_cookie_name`, emitted as the flat
`session.cookie_name` key (same lua-resty-session 4.x schema as the existing
`session.cookie_domain`), and a shared
`apisix_oidc_session_cookie_name(application, env_suffix)` helper.
`mit_learn_session_cookie_name()` is a thin wrapper over it, so every resource
participating in the shared MIT Learn login session derives its name from one
place. The `apisix` segment names the gateway as owner: the application behind
it sets cookies of its own on the same host — mit-learn's Django app being the
obvious case — so when the question is which component wrote an oversized
Cookie header, `mitlearn_apisix_session` answers it and `mitlearn_session`
does not.

The field stays optional rather than required. Requiring it would rename the
cookie for all 14 `OLApisixOIDCConfig` call sites across 12 stacks in one
deploy — airbyte, dagster, gwarek, opik, leek, marimo and ol-analytics-api's
canonical host — each of which also wants its own cleanup plugin to avoid
leaving a dead cookie behind. That is a follow-up, where the component
defaults the name from `application_name` + stack environment so no caller can
forget it, and the field remains an override for resources that must share a
name they do not own.

Production is unsuffixed — `mitlearn_apisix_session`, not
`mitlearn_apisix_session_production` — matching how the rest of the codebase names
environment-scoped, user-visible strings (compare the `learn_csrftoken` /
`learn_rc_csrftoken` pair). That is safe precisely because the non-production
names *are* suffixed: a Production `.learn.mit.edu` cookie riding along to
`api.rc.learn.mit.edu` is called `mitlearn_apisix_session` there, which is not the
name the RC gateway reads (`mitlearn_apisix_session_qa`), so the two never contend.

| OIDC resource | Host | Cookie name (Production) |
| --- | --- | --- |
| mit-learn, prefix + no-prefix | `api.learn.mit.edu` | `mitlearn_apisix_session` |
| learn-ai `/ai/*` | `api.learn.mit.edu` | `mitlearn_apisix_session` |
| ol-analytics-api, Learn-scoped | `analytics.learn.mit.edu` | `mitlearn_apisix_session` |
| mitxonline `/mitxonline/*` | `api.learn.mit.edu` | `mitlearn_apisix_session` |
| mitxonline, own login | `*.mitxonline.mit.edu` | `mitxonline_apisix_session` |
| jupyterhub | `nb.learn.mit.edu` | `jupyterhub_apisix_session` |
| jupyterhub-authoring | `authoring.nb.learn.mit.edu` | `jupyterhub_authoring_apisix_session` |

The first four all share the `sso/mitlearn` Keycloak client and read the
cookie mit-learn's login flow set, so they must agree exactly — a mismatch
silently turns their `unauth_action="pass"` routes anonymous. mitxonline's
`/mitxonline/*` routes are the sharpest case: they are served from MIT Learn's
own host and are what the MIT Learn frontend calls as
`NEXT_PUBLIC_MITX_ONLINE_BASE_URL`. They also pick up mit-learn's cookie
domain, because their `reqauth` route performs a real login and without a
domain that write lands as a *host-only* cookie under the shared name,
shadowing the `.learn.mit.edu` one on that host.

The last three are separate sessions and get their own names. mitxonline's own
login needed one for the same reason mit-learn did (parent-domain cookie
stacking across environments); jupyterhub's host-only "session" was colliding
outright with mit-learn's `.learn.mit.edu` "session" on `nb.learn.mit.edu`,
with the gateway reading whichever the browser happened to send first.

Also points the gateway's `$session_name` variable at the shared helper. That
variable only feeds the `$oidc_session_bytes` access-log field and the
`apisix_oidc_session_cookie_bytes` dashboard (lua-resty-session 4.x ignores
nginx-variable configuration), and it previously named a cookie that never
existed — `{env}_gateway_session` — so the metric was measuring nothing.

## Evicting the old cookies

Renaming does not remove the old cookie. The browser keeps sending the
abandoned lua-resty-session envelope to every route on the host, where the
plugin now ignores it — and these are large, and session-scoped, so they only
go away when the user fully quits their browser. Most people never do.

`stale_session_cookie_cleanup_plugin()` emits the matching `Max-Age=0`
deletions from a `serverless-post-function` in the `header_filter` phase,
guarded on a stale cookie actually being present on the request. So it costs
nothing on the overwhelming majority of responses and stops firing for a given
browser after the one response that clears it. Chunked cookies are handled
too — lua-resty-session splits an oversized payload across `<name>`,
`<name>_2`, `<name>_3`… and each chunk is its own cookie to expire.

A cookie's identity is (name, domain, path), so the host-only and
domain-scoped variants are separate entries in the browser's jar and each
needs its own deletion. The caller has to name the domains explicitly:
a `Domain=.mitxonline.mit.edu` deletion emitted from `api.learn.mit.edu` is
simply rejected by the browser, and deriving a domain from the request host
risks clearing an unrelated `.mit.edu` cookie belonging to someone else. That
is why mitxonline attaches this per route group rather than through its shared
plugin config, which both groups reference.

learn-ai and ol-analytics-api are deliberately not covered: they sit on hosts
mit-learn's own routes already clean (a domain-scoped cookie is a single jar
entry, so clearing `.learn.mit.edu` from mit-learn's routes clears it
everywhere), and ol-analytics-api's shared plugin config is also used by
`analytics.ol.mit.edu`, where "session" is still a **live** cookie for a
different Keycloak client.

Safe to delete these plugin attachments once the old cookies have aged out of
circulation.

## Incidental

`OLApisixSharedPlugins` now orders the defaults ahead of caller-supplied
plugins, so adding one renders as a one-line append in `pulumi preview`
instead of shifting every entry down by one. APISIX dispatches by each
plugin's registered priority rather than array position, so the order is
cosmetic. No existing caller passed `plugins`. The config also now accepts
`OLApisixPluginConfig` objects directly instead of only raw dicts.

## Rollout

Existing sessions are invalidated for mit-learn, mitxonline and jupyterhub:
browsers keep sending the old "session" cookie, which the renamed plugins
ignore, so users log in again once. local-dev is left alone — it pins APISIX
chart 2.12.0, whose plugin schema predates the flat `session.*` keys.

## Known gap

learn-ai has the same host-only shadow problem and is **not** fixed here. Its
OIDC resource has no `oidc_session_cookie_domain`, and its `reqauth` route on
`api.<env>.learn.mit.edu` (`/ai/admin/login/*`, `/ai/http/login/*`) performs a
login — so it writes a host-only `mitlearn_apisix_session` alongside mit-learn's
domain-scoped one. Impact is mild (same session secret, so either decrypts to
the same user) but it doubles the cookie bytes on the host we measure and
makes logout clear only one of the two. Fixing it properly means splitting
learn-ai into two OIDC resources, one per host group, the way mitxonline
already is, because that single resource is shared with the legacy
`api-learn-ai.ol.mit.edu` host where a `.learn.mit.edu` cookie domain would be
rejected outright.

## Validation

pre-commit (ruff format, ruff check, mypy, detect-secrets) passes, as does
the full test suite (226 tests).

`tests/ol_infrastructure/components/services/test_apisix.py` gains coverage
for the new surface: `session.cookie_name` emitted as the flat
lua-resty-session 4.x key when set, absent when not, coexisting with
`cookie_domain`, and surviving into the plugin config a route actually
receives; the derived names including the unsuffixed-Production and
hyphen-normalising cases; and the cleanup plugin's generated Lua — its
`header_filter` phase, per-domain deletions and chunked-cookie matching.

The Lua was additionally exercised against a real interpreter with stubbed
openresty: no false positives on `mitlearn_apisix_session`, `presession`,
`session_extra` or `my_session`; correct deletions for `session` and its
chunked variants. `pulumi preview` on the CI stacks for mit_learn, mitxonline
and jupyterhub shows the intended diffs and nothing else.

https://claude.ai/code/session_0179yZeaewx1VYnUCpxy4sju
https://claude.ai/code/session_019SwYqX2HKKLW7oHdcERrrP

### DIFF
--- a/src/bridge/lib/constants.py
+++ b/src/bridge/lib/constants.py
@@ -14,3 +14,37 @@ FASTLY_A_TLS_1_3 = [
     IPv4Address("151.101.130.132"),
     IPv4Address("151.101.194.132"),
 ]
+
+
+def mit_learn_session_cookie_name(env_suffix: str) -> str:
+    """Return the APISIX OIDC session cookie name for a MIT Learn environment.
+
+    APISIX's openid-connect plugin (lua-resty-session 4.x) names its session
+    cookie ``session`` by default.  That is a problem for MIT Learn for two
+    reasons:
+
+    * It is generic enough to collide with any other ``session`` cookie a
+      *.mit.edu host sets, and it says nothing about which system owns it when
+      debugging an oversized/garbled Cookie header.
+    * MIT Learn's session cookie is deliberately scoped to the parent domain
+      (``.learn.mit.edu``) so it also reaches sibling subdomains, which means
+      the Production cookie is sent to ``api.rc.learn.mit.edu`` and
+      ``api.ci.learn.mit.edu`` as well.  Under a single shared name the three
+      environments' cookies stack up on those hosts and the gateway can be
+      handed a session envelope it cannot decrypt (each environment has its own
+      session secret).  Giving every environment its own name keeps them
+      distinct in the browser's cookie jar.
+
+    Every APISIX OIDC resource that participates in the shared MIT Learn login
+    session must use the same name for a given environment: mit-learn's own
+    routes, learn-ai's ``/ai/*`` routes and ol-analytics-api's Learn-scoped
+    host all share the ``sso/mitlearn`` Keycloak client and read the cookie
+    mit-learn set, so a mismatch silently turns their ``unauth_action="pass"``
+    routes into anonymous ones.
+
+    :param env_suffix: The lowercase stack environment suffix, e.g. ``"qa"``.
+
+    :returns: The session cookie name for that environment.
+    :rtype: str
+    """
+    return f"mitlearn_session_{env_suffix.lower()}"
--- a/src/bridge/lib/constants.py
+++ b/src/bridge/lib/constants.py
@@ -16,35 +16,78 @@ FASTLY_A_TLS_1_3 = [
 ]
 
 
-def mit_learn_session_cookie_name(env_suffix: str) -> str:
-    """Return the APISIX OIDC session cookie name for a MIT Learn environment.
+# The cookie name lua-resty-session 4.x -- and therefore APISIX's
+# openid-connect plugin -- uses when nothing sets ``session.cookie_name``.
+# Every application that has since moved to an explicit name left one of these
+# behind in its users' browsers; see ``stale_session_cookie_cleanup_plugin`` in
+# ol_infrastructure.components.services.apisix for how they get evicted.
+DEFAULT_OIDC_SESSION_COOKIE_NAME = "session"  # pragma: allowlist secret
+
+
+def apisix_oidc_session_cookie_name(application: str, env_suffix: str) -> str:
+    """Return the APISIX OIDC session cookie name for an application/environment.
 
     APISIX's openid-connect plugin (lua-resty-session 4.x) names its session
-    cookie ``session`` by default.  That is a problem for MIT Learn for two
+    cookie ``session`` by default.  That is a poor default for us for two
     reasons:
 
     * It is generic enough to collide with any other ``session`` cookie a
       *.mit.edu host sets, and it says nothing about which system owns it when
       debugging an oversized/garbled Cookie header.
-    * MIT Learn's session cookie is deliberately scoped to the parent domain
-      (``.learn.mit.edu``) so it also reaches sibling subdomains, which means
-      the Production cookie is sent to ``api.rc.learn.mit.edu`` and
-      ``api.ci.learn.mit.edu`` as well.  Under a single shared name the three
+    * Several of our session cookies are deliberately scoped to a parent domain
+      (``.learn.mit.edu``, ``.mitxonline.mit.edu``) so they reach sibling
+      subdomains, which means the Production cookie is also sent to the RC and
+      CI hosts under that parent.  Under a single shared name the three
       environments' cookies stack up on those hosts and the gateway can be
       handed a session envelope it cannot decrypt (each environment has its own
       session secret).  Giving every environment its own name keeps them
       distinct in the browser's cookie jar.
 
-    Every APISIX OIDC resource that participates in the shared MIT Learn login
-    session must use the same name for a given environment: mit-learn's own
-    routes, learn-ai's ``/ai/*`` routes and ol-analytics-api's Learn-scoped
-    host all share the ``sso/mitlearn`` Keycloak client and read the cookie
-    mit-learn set, so a mismatch silently turns their ``unauth_action="pass"``
-    routes into anonymous ones.
+    Production is unsuffixed (``mitlearn_session``) to match how the rest of
+    the codebase names user-visible, environment-scoped strings -- compare the
+    ``learn_csrftoken`` / ``learn_rc_csrftoken`` pair.  That is safe precisely
+    because the non-production names *are* suffixed: a Production
+    ``.learn.mit.edu`` cookie riding along to ``api.rc.learn.mit.edu`` is
+    called ``mitlearn_session`` there, which is not the name the RC gateway
+    reads (``mitlearn_session_qa``), so the two never contend.
 
-    :param env_suffix: The lowercase stack environment suffix, e.g. ``"qa"``.
+    :param application: Slug of the application owning the session, e.g.
+        ``"mitlearn"``.  Hyphens are normalised to underscores.
+    :param env_suffix: The stack environment suffix, e.g. ``"qa"``.
 
-    :returns: The session cookie name for that environment.
+    :returns: The session cookie name for that application and environment.
     :rtype: str
     """
-    return f"mitlearn_session_{env_suffix.lower()}"
+    base = f"{application.replace('-', '_')}_session"
+    env = env_suffix.lower()
+    return base if env == "production" else f"{base}_{env}"
+
+
+def mit_learn_session_cookie_name(env_suffix: str) -> str:
+    """Return the APISIX OIDC session cookie name for a MIT Learn environment.
+
+    Every APISIX OIDC resource that participates in the shared MIT Learn login
+    session must use the same name for a given environment, so they all call
+    this rather than building the name themselves:
+
+    * mit-learn's own prefixed and un-prefixed routes on
+      ``api.<env>.learn.mit.edu``, which perform the login,
+    * learn-ai's ``/ai/*`` routes, served from that same host,
+    * ol-analytics-api's Learn-scoped host, and
+    * mitxonline's ``/mitxonline/*`` routes, which are served from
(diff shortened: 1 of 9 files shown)


src/ol_infrastructure/components/services/apisix.py
```diff
@@ -209,6 +209,11 @@ class OLApisixOIDCConfig(BaseModel):
         "user": True,
     }
     oidc_session_cookie_domain: str | None = None
+    # None leaves lua-resty-session's default cookie name ("session") in
+    # place. Applications that share a parent domain with other environments
+    # or other applications should set an explicit, environment-scoped name so
+    # their session cookies stay distinct in the browser's cookie jar.
+    oidc_session_cookie_name: str | None = None
```
ChristopherChudzicki, line 288: should it just be required to be set?

src/bridge/lib/constants.py
```diff
@@ -14,3 +14,37 @@
...
+    host all share the ``sso/mitlearn`` Keycloak client and read the cookie
+    mit-learn set, so a mismatch silently turns their ``unauth_action="pass"``
+    routes into anonymous ones.
+
+    :param env_suffix: The lowercase stack environment suffix, e.g. ``"qa"``.
+
+    :returns: The session cookie name for that environment.
+    :rtype: str
+    """
+    return f"mitlearn_session_{env_suffix.lower()}"
```
ChristopherChudzicki: would `mitlearn_apisix_session_...` be too verbose?

ChristopherChudzicki (approved):

src/bridge/lib/constants.py
blarghmatey: Agreed, and done in 41226a961 — names are now `mitlearn_apisix_session` / `mitxonline_apisix_session` / `jupyterhub_apisix_session` (still unsuffixed in Production, suffixed elsewhere).

The verbosity buys something real: the application behind the gateway sets its own cookies on the same host — mit-learn's Django app being the obvious case — so when the question in front of you is *which component wrote this oversized Cookie header*, `mitlearn_apisix_session` answers it and `mitlearn_session` does not. Seven bytes on a cookie this size is not the constraint.

src/ol_infrastructure/components/services/apisix.py
blarghmatey: Worth doing, but not here — it is a bigger change than it looks.

There are 14 `OLApisixOIDCConfig` call sites across 12 stacks. Making the field required renames the cookie for all of them in one deploy: airbyte, dagster, gwarek, opik, leek (celery-monitoring), marimo, and ol-analytics-api's canonical host all get logged out, and each also wants its own `stale_session_cookie_cleanup_plugin()` attachment or it leaves a dead `session` cookie behind — which is the other half of what this PR is about. Folding that in would take the blast radius from three applications to ten.

The better end state is not *required* but *automatic*: have the component default the name from `application_name` + the stack environment, so no caller can forget it, and keep the field as an explicit override for the resources that must share a name they do not own (learn-ai, ol-analytics-api's Learn host and mitxonline's `/mitxonline/*` routes all need `mitlearn_apisix_session`, not their own).

```python
cookie_name = oidc_config.oidc_session_cookie_name or apisix_oidc_session_cookie_name(
    oidc_config.application_name, parse_stack().env_suffix
)
```

I have left it optional here and will open that as a follow-up so the rename and the cleanup plugins land together. Reasoning recorded in 41226a961's commit message.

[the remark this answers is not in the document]
blarghmatey: Good catch — added in 41226a961, in `tests/ol_infrastructure/components/services/test_apisix.py` (23 tests in the file now, all passing).

Covering exactly what you asked plus the rest of the new surface:

- `session.cookie_name` is emitted as the **flat** lua-resty-session 4.x key when `oidc_session_cookie_name` is set — the nested `session.cookie.name` form is a silent no-op on the pinned APISIX 3.17.0, so this is the regression most worth pinning down.
- It coexists with `cookie_domain` in one `session` block rather than either overwriting the other.
- The `session` block is **omitted entirely** when a caller opts into none of the session settings, so lua-resty-session's own defaults are not clobbered by an empty dict.
- The name survives into `get_full_oidc_plugin_config()`, i.e. the config a route actually receives, not just the component's intermediate dict.
- The derived names themselves: unsuffixed in Production, suffixed elsewhere, hyphens normalised (`jupyterhub-authoring` → `jupyterhub_authoring_apisix_session`), and `mit_learn_session_cookie_name()` not drifting from the generic helper.
- The cleanup plugin's generated Lua: `header_filter` phase, one deletion per (name, domain) pair, and the chunked-cookie match (`session_2`, `session_3`, …).

The Lua was separately exercised against a real interpreter with stubbed openresty before landing: no false positives on `mitlearn_apisix_session`, `presession`, `session_extra` or `my_session`, correct deletions for `session` and its chunks.

blarghmatey:
Feedback pass complete — all 3 review threads addressed and resolved, in 41226a961.

- **Cookie naming** (@ChristopherChudzicki): took the more verbose form. Names are now `mitlearn_apisix_session` / `mitxonline_apisix_session` / `jupyterhub_apisix_session`, still unsuffixed in Production. The gateway segment earns its bytes — the application behind it sets cookies on the same host, so this distinguishes them when you are staring at an oversized Cookie header.
- **Required field?** (@ChristopherChudzicki): kept optional here, with reasoning on the thread. Requiring it renames the cookie for all 14 OIDC call sites across 12 stacks at once and each needs its own cleanup plugin, so it goes in a follow-up that defaults the name from `application_name` + stack environment rather than merely requiring it.
- **Test coverage** (Copilot): added to `tests/ol_infrastructure/components/services/test_apisix.py` — the flat `session.cookie_name` rendering (set, unset, alongside `cookie_domain`, and through `get_full_oidc_plugin_config()`), the derived names, and the cleanup plugin's generated Lua including per-domain deletions and chunked-cookie matching. This also covers both of the suppressed comments on the latest Copilot review.

Full suite green (226 tests), pre-commit clean, and `pulumi preview` re-run on the mit_learn CI stack to confirm the renamed cookie renders. All 7 checks passing.

Issue #5004, opened 2026-07-15
Sentry DSN Usage Simplification


feoh:
### Description/Context
We currently have cargo-culted, hard-coded Sentry DSNs scattered across several of our projects. These should instead be regularly-named, defined outputs from the Sentry Pulumi stack, rather than values that get copy-pasted and stored as secrets in Vault.

Standardizing DSN provisioning as Pulumi stack outputs will let us delete the corresponding hard-coded DSN secrets from Vault, reducing secret sprawl and the risk of stale/incorrect DSNs being used.

### Plan/Design
- Audit projects for hard-coded Sentry DSNs and identify which Sentry projects/stacks they correspond to.
- Define a consistent naming convention for Sentry DSN outputs in the Sentry Pulumi stack.
- Update the Sentry Pulumi stack to expose DSNs as stack outputs using that convention.
- Migrate consuming projects to read the DSN from the Pulumi stack output instead of a hard-coded value or ad hoc Vault secret.
- Remove the now-redundant DSN secrets from Vault.

feoh:
## Progress update

All target Sentry projects/keys already existed in `infrastructure/sentry/__main__.py` (auto-imported from live Sentry) — only `ol_analytics_api` followed the desired stack-output pattern. 10 PRs are up implementing the rest of the plan; 2 items need a decision before they can be finished.

### PRs ready for review

1. #5222 — **Foundation**: adds `pulumi.export(...)` stack outputs for all remaining Sentry projects/keys. Must merge + deploy first; everything else depends on it.
2. #5223 — airbyte
3. #5224 — edxapp (all 4 deployments: mitx, mitx-staging, mitxonline, xpro)
4. #5225 — learn_ai
5. #5226 — mit_learn (Django backend)
6. #5227 — mit_learn_nextjs (frontend; confirmed it shares the `open-next` Sentry project with mit_learn's backend)
7. #5228 — mitxonline (Django backend) — first-time Pulumi-managed write to a previously externally-populated Vault path, flagged for extra review
8. #5229 — ocw_studio
9. #5230 — open_discussions (legacy Heroku `mit-open` app — confirmed still live, distinct from mit_learn) — same first-time-write caveat as #5228
10. #5231 — xpro (Django backend, distinct from the xpro edxapp instance in #5224)

Suggested merge/deploy order: #5222 first, then the rest in any order, then verify each app's Sentry error reporting still works before removing the corresponding hard-coded secret.

[...]

PR #5222, opened 2026-08-03
Add Sentry DSN stack outputs for all Sentry projects

feoh:
Part of https://github.com/mitodl/ol-infrastructure/issues/5004

Adds `pulumi.export(...)` calls for every remaining `SentryKey` resource in `infrastructure/sentry/__main__.py`, exposing each Sentry project's DSN as a stack output. This follows the pattern already established by `ol_analytics_api_sentry_dsn` (the only existing example of this pattern in the file).

Naming convention: `<project_slug>_sentry_dsn`, with a name-suffixed variant for projects that have more than one key (e.g. `odl_video_service_eternal_mink_sentry_dsn` alongside `odl_video_service_sentry_dsn`).

This is the foundational change for #5004: it does not remove any existing hard-coded/SOPS/Vault secrets by itself, but unblocks a series of follow-up PRs that migrate each consuming application to read its DSN via `sentry_stack.require_output(...)` instead.

Also documents this as a hand-authored generator exception in `IMPORT_SUMMARY.md`, matching the existing note for `ol_analytics_api`, since `infrastructure/sentry/__main__.py` is otherwise owned by `bin/import-sentry-config`.

### How can this be tested?
- `uv run python -m py_compile src/ol_infrastructure/infrastructure/sentry/__main__.py`
- `uv run ruff check src/ol_infrastructure/infrastructure/sentry/__main__.py`
- `pulumi preview` against the `ol-infrastructure-sentry` `Production` stack should show only new stack outputs being added, no resource changes.

### Additional Context
This PR must be merged and deployed (`pulumi up` on the `ol-infrastructure-sentry` `Production` stack) before any of the follow-up consumer-migration PRs from #5004 can be deployed, since they call `sentry_stack.require_output(...)` on these new outputs.

### DIFF
--- a/src/ol_infrastructure/infrastructure/sentry/__main__.py
+++ b/src/ol_infrastructure/infrastructure/sentry/__main__.py
@@ -3715,3 +3715,78 @@ key_dagster = sentry.SentryKey(
 )
 
 pulumi.export("dagster_sentry_dsn", key_dagster.dsn_public)
+
+# Hand-added stack outputs (not produced by bin/import-sentry-config) exposing
+# each project's DSN so consuming stacks can read it via
+# sentry_stack.require_output(...) instead of a hard-coded/SOPS/Vault secret.
+# See ol-infrastructure#5004. Naming convention: `<project_slug>_sentry_dsn`,
+# with a name-suffixed variant for projects that have more than one key.
+pulumi.export(
+    "airbyte_sentry_dsn",
+    key_airbyte_default_0e552f29a3594a6a972cf2d73dfa140b.dsn_public,
+)
+pulumi.export(
+    "learn_ai_sentry_dsn",
+    key_learn_ai_default_5e497d1f33127fc35dfc811b290747c7.dsn_public,
+)
+pulumi.export(
+    "micromasters_sentry_dsn",
+    key_micromasters_default_6134fa8f62f74ba49d3892f9eb8c2fb7.dsn_public,
+)
+pulumi.export(
+    "mitxonline_sentry_dsn",
+    key_mitxonline_default_8ba006dc817541a0865033722f8289ad.dsn_public,
+)
+pulumi.export(
+    "ocw_next_sentry_dsn",
+    key_ocw_next_default_eee58f41dda54d2b814296e12dced4b7.dsn_public,
+)
+pulumi.export(
+    "ocw_studio_sentry_dsn",
+    key_ocw_studio_default_ba06ef0080e342db9b9ae1db53944f1a.dsn_public,
+)
+pulumi.export(
+    "odl_video_service_sentry_dsn",
+    key_odl_video_service_default_f75d3b6abb8f4a30b95769f8f9644ee3.dsn_public,
+)
+pulumi.export(
+    "odl_video_service_eternal_mink_sentry_dsn",
+    key_odl_video_service_eternal_mink_7e70a5f1227743959ba1cb1e8db1240f.dsn_public,
+)
+pulumi.export(
+    "open_sentry_dsn", key_open_default_005a98ba45c3472dbb4c12405d814557.dsn_public
+)
+pulumi.export(
+    "open_next_sentry_dsn",
+    key_open_next_default_8131564a9b9e31ac3cbcf7952f0fea80.dsn_public,
+)
+pulumi.export(
+    "openedx_mitxonline_sentry_dsn",
+    key_openedx_mitxonline_default_8807f1e77ff44d91beeb2580507ad84a.dsn_public,
+)
+pulumi.export(
+    "openedx_mitxpro_sentry_dsn",
+    key_openedx_mitxpro_default_3b523a8e89c040f0a46ddc90a64f9d14.dsn_public,
+)
+pulumi.export(
+    "openedx_residential_sentry_dsn",
+    key_openedx_residential_default_3741f5e8334b429eaa048bec1e3daa30.dsn_public,
+)
+pulumi.export(
+    "python_sentry_dsn", key_python_default_7cc91e53ac70f26e91a2e788288be63d.dsn_public
+)
+pulumi.export(
+    "release_script_sentry_dsn",
+    key_release_script_default_67d74d4e5121a4ca332f1523ff8affdd.dsn_public,
+)
+pulumi.export(
+    "sandbox_sentry_dsn",
+    key_sandbox_default_8c6395a9422040c999373e204b15bd8b.dsn_public,
+)
+pulumi.export(
+    "unified_ecommerce_sentry_dsn",
+    key_unified_ecommerce_default_734c791eb3c83eb35db04c1b753ff67e.dsn_public,
+)
+pulumi.export(
+    "xpro_sentry_dsn", key_xpro_default_c9f9886a461647708cd0c0c551166866.dsn_public
+)
--- a/src/ol_infrastructure/infrastructure/sentry/__main__.py
+++ b/src/ol_infrastructure/infrastructure/sentry/__main__.py
@@ -252,8 +252,8 @@ project_open = sentry.SentryProject(
 project_open_next = sentry.SentryProject(
     "project_open_next",
     organization=ORGANIZATION,
-    name="open-next",
-    slug="open-next",
+    name="mit-learn",
+    slug="mit-learn",
     platform="python-django",
     teams=["mit-learn", "mit-office-of-digital-learning", "devops"],
     digests_min_delay=300,
@@ -301,19 +301,6 @@ project_openedx_residential = sentry.SentryProject(
     opts=sentry_opts,
 )
 
-project_python = sentry.SentryProject(
-    "project_python",
-    organization=ORGANIZATION,
-    name="python",
-    slug="python",
-    platform="python",
-    teams=["mit-office-of-digital-learning"],
-    digests_min_delay=300,
-    digests_max_delay=1800,
-    resolve_age=0,
-    opts=sentry_opts,
-)
-
 project_release_script = sentry.SentryProject(
     "project_release_script",
     organization=ORGANIZATION,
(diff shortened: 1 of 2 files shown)


src/ol_infrastructure/infrastructure/sentry/__main__.py
```diff
@@ -3680,3 +3680,78 @@
...
+    "odl_video_service_sentry_dsn",
+    key_odl_video_service_default_f75d3b6abb8f4a30b95769f8f9644ee3.dsn_public,
+)
+pulumi.export(
+    "odl_video_service_eternal_mink_sentry_dsn",
+    key_odl_video_service_eternal_mink_7e70a5f1227743959ba1cb1e8db1240f.dsn_public,
+)
+pulumi.export(
+    "open_sentry_dsn", key_open_default_005a98ba45c3472dbb4c12405d814557.dsn_public
+)
```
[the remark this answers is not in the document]
feoh: Fixed in 48704c6f: reformatted the 3 short exports (open_sentry_dsn, python_sentry_dsn, xpro_sentry_dsn) with a trailing comma so they explode to the same one-arg-per-line style as the rest of the block. Verified with `ruff format --check` -- both the collapsed and exploded forms were valid fixed points (ruff's magic-trailing-comma rule), so the trailing comma is what pins it to the consistent style going forward without any future format churn.

Ardiea (approved):

src/ol_infrastructure/infrastructure/sentry/__main__.py
```diff
@@ -3680,3 +3680,81 @@
...
+    key_release_script_default_67d74d4e5121a4ca332f1523ff8affdd.dsn_public,
+)
+pulumi.export(
+    "sandbox_sentry_dsn",
+    key_sandbox_default_8c6395a9422040c999373e204b15bd8b.dsn_public,
+)
+pulumi.export(
+    "unified_ecommerce_sentry_dsn",
+    key_unified_ecommerce_default_734c791eb3c83eb35db04c1b753ff67e.dsn_public,
+)
```
blarghmatey: This is a dead project, we should just remove it wholesale

src/ol_infrastructure/infrastructure/sentry/__main__.py
```diff
@@ -3680,3 +3680,81 @@
...
+    key_python_default_7cc91e53ac70f26e91a2e788288be63d.dsn_public,
+)
+pulumi.export(
+    "release_script_sentry_dsn",
+    key_release_script_default_67d74d4e5121a4ca332f1523ff8affdd.dsn_public,
+)
+pulumi.export(
+    "sandbox_sentry_dsn",
+    key_sandbox_default_8c6395a9422040c999373e204b15bd8b.dsn_public,
+)
```
blarghmatey: What's "Sandbox"?

src/ol_infrastructure/infrastructure/sentry/__main__.py
```diff
@@ -3680,3 +3680,81 @@
...
+    key_openedx_mitxpro_default_3b523a8e89c040f0a46ddc90a64f9d14.dsn_public,
+)
+pulumi.export(
+    "openedx_residential_sentry_dsn",
+    key_openedx_residential_default_3741f5e8334b429eaa048bec1e3daa30.dsn_public,
+)
+pulumi.export(
+    "python_sentry_dsn",
+    key_python_default_7cc91e53ac70f26e91a2e788288be63d.dsn_public,
+)
```
blarghmatey: What's this project?

src/ol_infrastructure/infrastructure/sentry/__main__.py
```diff
@@ -3680,3 +3680,81 @@
...
+pulumi.export(
+    "open_sentry_dsn",
+    key_open_default_005a98ba45c3472dbb4c12405d814557.dsn_public,
+)
+pulumi.export(
+    "open_next_sentry_dsn",
+    key_open_next_default_8131564a9b9e31ac3cbcf7952f0fea80.dsn_public,
+)
+pulumi.export(
```
blarghmatey, line 3676: Worth renaming the project and this export name so that it's more sensical match to what i suspect is actually mit-learn

src/ol_infrastructure/infrastructure/sentry/__main__.py
```diff
@@ -3680,3 +3680,81 @@
...
+    "micromasters_sentry_dsn",
+    key_micromasters_default_6134fa8f62f74ba49d3892f9eb8c2fb7.dsn_public,
+)
+pulumi.export(
+    "mitxonline_sentry_dsn",
+    key_mitxonline_default_8ba006dc817541a0865033722f8289ad.dsn_public,
+)
+pulumi.export(
+    "ocw_next_sentry_dsn",
```
blarghmatey:
```suggestion
    "ocw_site_sentry_dsn",
```

src/ol_infrastructure/infrastructure/sentry/__main__.py
```diff
@@ -3680,3 +3680,81 @@
...
+
+# Hand-added stack outputs (not produced by bin/import-sentry-config) exposing
+# each project's DSN so consuming stacks can read it via
+# sentry_stack.require_output(...) instead of a hard-coded/SOPS/Vault secret.
+# See ol-infrastructure#5004. Naming convention: `<project_slug>_sentry_dsn`,
+# with a name-suffixed variant for projects that have more than one key.
+pulumi.export(
+    "airbyte_sentry_dsn",
+    key_airbyte_default_0e552f29a3594a6a972cf2d73dfa140b.dsn_public,
+)
```
blarghmatey: This doesn't even actually work so we just just nuke it

src/ol_infrastructure/infrastructure/sentry/__main__.py
blarghmatey: Fixed in 22c7b6532: removed `project_unified_ecommerce`, its `SentryKey`, and the `unified_ecommerce_sentry_dsn` export entirely. Checked via the Sentry MCP -- zero issues/events for `unified-ecommerce` in the last 90 days, and nothing else in this file references it, so there's nothing else to clean up.

src/ol_infrastructure/infrastructure/sentry/__main__.py
blarghmatey: Fixed in 22c7b6532: removed `project_airbyte`, its `SentryKey`, and the `airbyte_sentry_dsn` export entirely. Checked via the Sentry MCP -- zero issues/events for `airbyte` in the last 90 days, confirming it's not actually receiving events.

src/ol_infrastructure/infrastructure/sentry/__main__.py
blarghmatey: Fixed in 22c7b6532: applied the suggestion, and also renamed the underlying Sentry project (`name`/`slug`) from `ocw-next` to `ocw-site` and updated the key's `project` reference to match, since the `ocw_site` app is what will actually consume this DSN. Left the Pulumi resource identifiers (`project_ocw_next`, `key_ocw_next_default_*`) unchanged so this is an in-place update, not a replace.

src/ol_infrastructure/infrastructure/sentry/__main__.py
blarghmatey: Looked this up via the Sentry MCP: `project_sandbox` (name="Python", slug="sandbox") has zero issues/events in the last 90 days, and no application in this repo references its DSN. Looks like an ad hoc scratch/testing project rather than something tied to a deployed service -- leaving it as-is rather than guessing at a rename; can you confirm what it's actually for?

src/ol_infrastructure/infrastructure/sentry/__main__.py
blarghmatey: Looked this up via the Sentry MCP: `project_python` (name=slug="python", team mit-office-of-digital-learning) has zero issues/events in the last 90 days, and no application in this repo references its DSN. Same read as the Sandbox project above -- looks like a generic scratch project, not something I can confidently rename without more context. Can you confirm what it's actually for?

src/ol_infrastructure/infrastructure/sentry/__main__.py
blarghmatey: Checked via the Sentry MCP: there ARE two separate live projects here -- `open` (name "Discussions") and `open-next` -- so this isn'\''t a case of the same app double-counted. But `open-next`'\''s own GitHub code mappings don'\''t point cleanly at a single "mit-learn-nextjs" identity: it's mapped to mitodl/mit-learn, mitodl/learn-ai, mitodl/ecommerce, mitodl/open-discussions (x3), and mitodl/edx-sga -- several of which overlap with `open`'\''s own mappings (which are dominated by mitodl/open-discussions + mitodl/mit-learn). That mix doesn'\''t match a single Next.js frontend app, so I don'\''t want to guess a name and rename a live Sentry project on a possibly-wrong assumption. Holding off on this one -- can you confirm what `open-next` is actually meant to represent (a real separate app, a "next"/staging environment for the same backend, or leftover from a past migration) so I rename it correctly?

src/ol_infrastructure/infrastructure/sentry/__main__.py
blarghmatey: Fixed in 1ba7054d7: deleted `project_sandbox`, its `SentryKey`, and the `sandbox_sentry_dsn` export entirely -- same zero-activity finding as before.

src/ol_infrastructure/infrastructure/sentry/__main__.py
blarghmatey: Fixed in 1ba7054d7: deleted `project_python`, its `SentryKey`, and the `python_sentry_dsn` export entirely -- same zero-activity finding as the Sandbox project.

src/ol_infrastructure/infrastructure/sentry/__main__.py
blarghmatey: Fixed in 1ba7054d7. Dug into the actual issue data via the Sentry MCP to confirm before renaming a live project: `open` (1M+ events, unresolved) is entirely legacy -- old `/api/v0/...` routes and old task/app names (`search.tasks.*`, `course_catalog.tasks.*`, `moira_lists.tasks.*`), matching the open-discussions/mit_open Heroku deployment. `open-next` is very much alive (25 unresolved issues in the last few days alone) and its culprits are a mix of the Next.js frontend (`/dashboard`, `/search`, `/video/:id/:slug`, ChunkLoadError) *and* the current backend (`/api/v1/learning_resources/...`, `vector_search.tasks.*`, SCIM) -- i.e. it'\''s the whole current MIT Learn stack sharing one project, not a separate frontend-only app. Renamed `open-next` -> `mit-learn` (name/slug + key ref + export `mit_learn_sentry_dsn`), same in-place-update approach as the ocw-site rename. Left `open` alone since it'\''s the legacy side of this, not part of what you flagged.

PR #5223, opened 2026-08-03
Migrate airbyte to stack-output Sentry DSN

feoh:
Part of https://github.com/mitodl/ol-infrastructure/issues/5004

`applications/airbyte/__main__.py` previously read its Sentry DSN from a Vault secret at `sentry-dsn`, itself populated from a SOPS-encrypted value in `bridge/secrets/airbyte/data.{ci,qa,production}.yaml`. This switches it to `sentry_stack.require_output("airbyte_sentry_dsn")`, reading the DSN from the `ol-infrastructure-sentry` Pulumi stack instead.

The now-unused `read_yaml_secrets` call and its import were removed since nothing else in the file referenced that dict.

### How can this be tested?
- `uv run python -m py_compile src/ol_infrastructure/applications/airbyte/__main__.py`
- `uv run ruff check src/ol_infrastructure/applications/airbyte/__main__.py`
- `pulumi preview` against an airbyte stack should show only the `airbyte-server-configuration-sentry-secrets` Vault secret's value changing (same shape: `{"value": <dsn>}`), no other resource changes.

### Additional Context
Depends on https://github.com/mitodl/ol-infrastructure/pull/5222 being merged and deployed to the `ol-infrastructure-sentry` `Production` stack first, since this PR calls `sentry_stack.require_output("airbyte_sentry_dsn")`.

Once this is deployed and verified, the `sentry-dsn` entry in `bridge/secrets/airbyte/data.{ci,qa,production}.yaml` becomes redundant and can be removed in a follow-up cleanup PR.

### DIFF
--- a/src/ol_infrastructure/applications/airbyte/__main__.py
+++ b/src/ol_infrastructure/applications/airbyte/__main__.py
@@ -23,7 +23,6 @@ from bridge.lib.magic_numbers import (
     DEFAULT_POSTGRES_PORT,
 )
 from bridge.lib.versions import AIRBYTE_CHART_VERSION
-from bridge.secrets.sops import read_yaml_secrets
 from ol_infrastructure.components.aws.database import OLAmazonDB, OLPostgresDBConfig
 from ol_infrastructure.components.aws.eks import (
     OLEKSTrustRole,
@@ -88,6 +87,7 @@ vault_stack = make_stack_reference(
     projects.VAULT_SERVER, f"operations.{stack_info.name}"
 )
 cluster_stack = make_stack_reference(projects.EKS, f"data.{stack_info.name}")
+sentry_stack = make_stack_reference(projects.SENTRY, "Production")
 
 mitodl_zone_id = dns_stack.require_output("odl_zone_id")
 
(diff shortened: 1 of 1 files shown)


feoh:
Closing — this migration cannot work as written, and there is nothing left to migrate to.

This PR reads `sentry_stack.require_output("airbyte_sentry_dsn")`, but no such output exists. Per `infrastructure/sentry/IMPORT_SUMMARY.md` (line 88), the airbyte Sentry project was deliberately removed from the stack after the config import:

> `project_airbyte` / `key_airbyte_default_*`, `project_unified_ecommerce` / `key_unified_ecommerce_default_*`, `project_python` / `key_python_default_*`, and `project_sandbox` / `key_sandbox_default_*`: removed by hand post-import (unused, zero Sentry events/issues in any of them).

Those resources were unprotected by hand and deleted via `pulumi up`, so the underlying Sentry project and key are gone live, not merely unexported. There is no DSN to expose as a stack output.

Two follow-ups worth considering separately, rather than reworking this PR:

1. The existing `sentry-dsn` value in `bridge/secrets/airbyte/data.{ci,qa,production}.yaml` and the `airbyte-server-configuration-sentry-secrets` Vault write now point at a deleted Sentry project, so airbyte error reporting is presumably already inert. Cleanup would mean removing that plumbing, not repointing it.
2. If airbyte *should* be reporting to Sentry, that is a new decision — it needs a Sentry project created and declared in the sentry stack first, at which point a fresh migration PR is straightforward.

Also note this PR changed the Vault payload shape from `json.dumps(airbyte_vault_secrets["sentry-dsn"])` to `json.dumps({"value": dsn})`, which would need verifying against whatever consumes that path if the work is revived.

Reopening is easy if the intent is to re-create the project rather than drop it.

PR #5242, opened 2026-08-04
fix(fastly): pin provider before TLS subscription regression

feoh:

N/A

Caps `pulumi-fastly` below 12.4 and restores the lockfile to 12.3.1.

`pulumi-fastly` 12.4.0 upgraded the underlying Fastly Terraform provider to 9.4.0, which included [fastly/terraform-provider-fastly#1353](https://github.com/fastly/terraform-provider-fastly/pull/1353). That change began sending API updates when only a TLS subscription's `configuration_id` appears to change.

The Fastly provider's read path can select a historical activation for a domain instead of its current activation. In the MITx Online Production stack this makes the checkpoint alternate between the old TLS v1.2 configuration and the current TLS v1.3 configuration. Version 12.4.0 then attempts to reconcile that false drift, and Fastly rejects the request:

```
Can't update subscription
Subscription has active domains
```

Version 12.3.1 still reads the inconsistent state, but does not issue an API request for a configuration-only diff. The upper bound is documented as temporary and should be removed after the upstream read-path bug is fixed.

### How can this be tested?

1. Run `uv lock --check` and `uv sync --frozen`.
2. Confirm `uv run python -c "from importlib.metadata import version; print(version('pulumi-fastly'))"` prints `12.3.1`.
3. From `src/ol_infrastructure/applications/mitxonline`, run `pulumi preview --stack Production --diff --refresh=false`.
4. Confirm the preview downgrades the Fastly provider from 12.4.0 to 12.3.1 and contains no replacements or deletions.

Validation performed:

- `uv lock --check` — passed
- `uv sync --frozen` — passed
- `uv run pre-commit run --files pyproject.toml uv.lock` — passed
- `uv run pytest tests/ol_infrastructure/ -q` — 200 passed
- Production Pulumi preview — completed; 3 updates, 139 unchanged, no replacements or deletions

### Additional Context

The preview continues to display the existing TLS subscription state diff. Under provider 12.3.1, its update handler does not call Fastly unless `domains` or `common_name` changed, so the false configuration-only diff no longer triggers the rejected production API operation.

A full repository pre-commit run passed all relevant checks but still reports existing environmental/repository findings: Packer is unavailable locally, and hadolint reports pre-existing Dockerfile warnings. A direct repository-wide mypy run also reports the repository's existing type-check backlog; this PR changes no Python source.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -21,7 +21,7 @@ dependencies = [
   "pulumi-aws>=7.1,<8",
   "pulumi-docker>=5,<6",
   # 12.4.0 turns historical TLS activation state drift into unsafe API updates.
-  # Remove this cap after the read-path regression exposed by upstream #1353 is fixed.
+  # Remove this cap after the fastly/terraform-provider-fastly read-path regression (PR #1353) is fixed.
   "pulumi-fastly>=12,<12.4",
   "pulumi-github>=6.0.0,<7",
   "pulumi-mailgun>=3.0.0,<4",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -20,7 +20,9 @@ dependencies = [
   "pulumi>=3.39.1,<4",
   "pulumi-aws>=7.1,<8",
   "pulumi-docker>=5,<6",
-  "pulumi-fastly>=12,<13",
+  # 12.4.0 turns historical TLS activation state drift into unsafe API updates.
+  # Remove this cap after the read-path regression exposed by upstream #1353 is fixed.
+  "pulumi-fastly>=12,<12.4",
   "pulumi-github>=6.0.0,<7",
   "pulumi-mailgun>=3.0.0,<4",
   "pulumi-tls>=5.0.0,<6",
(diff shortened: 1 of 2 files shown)


blarghmatey:
The real problem here was that we had 2 different TLS subscriptions in Fastly for the production MITx Online domain, one that was old and created manually and served TLS 1.2, and one created by Pulumi that served TLS 1.3. Manually removing the old one fixed the issue with this Pulumi up.

The way I figured it out was to install the Fastly CLI and then run `fastly tls-config describe --id <id from Pulumi diff>`

That yielded:
```
ID: <id>
Name: TLS v1.2
Bulk: false
Default: true
HTTP Protocol: http/1.1
HTTP Protocol: http/2
TLS Protocol: 1.2
Created at: 2020-05-13 12:43:09 +0000 UTC
Updated at: 2024-06-10 22:36:06 +0000 UTC
```

That caused me to go hunting in the Fastly UI which is when I found the culprit.

PR #4750, opened 2026-06-09
feat: add ol-python-base shared image and migrate local-dev to granian

blarghmatey:
### What are the relevant tickets?
N/A

### Description (What does it do?)
Adds a shared Python base image (`mitodl/ol-python-base:{3.11,3.12,3.13}`) that centralises the substrate all MIT OL Django services previously duplicated in every repo's Dockerfile: Python slim base, common-core apt packages (`build-essential git curl libjpeg-dev zlib1g-dev net-tools pkg-config libffi-dev libxmlsec1-dev libpq-dev postgresql-client`), the `uv` binary, a non-root `mitodl` user, `/src /opt/venv /var/media` directories, and environment variables (`PYTHONUNBUFFERED`, `PYTHONDONTWRITEBYTECODE`, `UV_PROJECT_ENVIRONMENT`, `UV_CACHE_DIR`, `PATH`). App Dockerfiles will adopt `FROM mitodl/ol-python-base:<ver>` in follow-up PRs (draft, blocked on this merging and the image publishing).

Also migrates both local-dev deployment manifests from `uwsgi` to `granian --interface wsgi`, standardising on granian across the stack.

**Files changed:**
- `dockerfiles/ol-python-base/Dockerfile` — parameterised base image (`ARG PYTHON_VERSION`)
- `dockerfiles/README.md` — documents the image and `fly set-pipeline` registration
- `pipelines/infrastructure/ol_python_base_docker.yaml` — Concourse pipeline that builds/pushes all three tags on Dockerfile changes to `main`
- `local-dev/apps/mitxonline/deployment.yaml` — uwsgi → granian (`main.wsgi:application`, port name `http-wsgi`)
- `local-dev/apps/odl-video-service/deployment.yaml` — uwsgi → granian (`odl_video.wsgi:application`, port name `http-wsgi`)

**Registering the Concourse pipeline** (required after merge, one-time per Concourse target):
```
fly -t <target> set-pipeline \
    -p ol-python-base-docker \
    -c pipelines/infrastructure/ol_python_base_docker.yaml
```
Credentials are resolved from Vault as `((dockerhub.username))` / `((dockerhub.password))`, matching the existing `ol_infrastructure_docker.yaml` pipeline convention.

### How can this be tested?
Build all three variants locally:
```bash
cd dockerfiles/ol-python-base
for v in 3.11 3.12 3.13; do
[...]

PR #5277, opened 2026-08-05
fix(local-dev): make kc-seed-users.sh compatible with bash 3.2

### LINKED ISSUE #5275: kc-seed-users.sh fails on macOS due to bash 3.2 incompatibility (declare -A unsupported)
This happens on every run (fresh cluster bring-up or Tilt re-trigger) on macOS, since the failing lines execute unconditionally before any idempotency check.

### Steps to Reproduce
1. On a Mac using the stock system `/bin/bash` (bash 3.2 — Apple hasn't shipped a newer bash since it moved to GPLv3 licensing), stand up the local-dev environment per `local-dev/README.md`.
2. Let Tilt run the `kc-seed-users` resource (or run `local-dev/scripts/kc-seed-users.sh` directly).
3. Observe the build fails with `line 81: admin: unbound variable`.

### Possible Solution
Root cause: [`kc-seed-users.sh` lines 81-84](https://github.com/mitodl/ol-infrastructure/blob/main/local-dev/scripts/kc-seed-users.sh#L81-L84) use `declare -A` (bash associative arrays), and later in the file an `ADMIN_USERS=("admin")` array is looped over. Bash 3.2 doesn't support associative arrays — under `set -u` (`nounset`), it misparses `declare -A arr=( [admin]=... )` as an indexed-array subscript referencing an unset variable `admin`, producing a confusing "unbound variable" error instead of a clear syntax error.

Suggested fix: replace the three `declare -A` maps (`USER_EMAILS`, `USER_FIRST`, `USER_LAST`) with `case`-statement lookup functions, and replace the `ADMIN_USERS` array + loop with a direct string comparison (`[ "$USERNAME" = "admin" ]`) — both are portable to bash 3.2 and preserve identical behavior. This has already been prototyped and verified working against real bash 3.2 locally.

### Additional Details
Introduced by commit [`fef2b977f`](https://github.com/mitodl/ol-infrastructure/commit/fef2b977f) ("fix(local-dev): auto-seed Keycloak test users and make the seed idempotent"). Affects any developer running local-dev on stock macOS, not just one machine — every fresh cluster bring-up requiring this script to run will hit this failure.

alexfigtree:
Closes #5275

Fixes `kc-seed-users.sh` to work on bash 3.2, which is the only `/bin/bash` macOS ships (Apple hasn't updated it since bash moved to GPLv3 licensing).

The script previously used `declare -A` associative arrays (`USER_EMAILS`, `USER_FIRST`, `USER_LAST`) and an array-based loop (`ADMIN_USERS`) to assign the admin realm role. Both require bash 4+. Under bash 3.2 with `set -u`, `declare -A` silently misparses and fails with a confusing `line 81: admin: unbound variable` error instead of a clear syntax error — so the `kc-seed-users` Tilt resource fails on every fresh local-dev bring-up on stock macOS.

Replaces the associative-array lookups with portable `case`-statement functions (`user_email`, `user_first`) and replaces the `ADMIN_USERS` array/loop with a direct string comparison (`[ "$USERNAME" = "admin" ]`). Behavior is unchanged — same three users created, same emails, same admin role assignment, same idempotency (skips users that already exist).

N/A — shell script change, no UI

### How can this be tested?
Tested directly against macOS's system bash (`/bin/bash`, bash 3.2.57):
- **Before the fix**: running `local-dev/scripts/kc-seed-users.sh` fails immediately with `line 81: admin: unbound variable`.
- **After the fix**: the script runs successfully, creating (or skipping, if already present) the three local-dev test users — `admin@odl.local`, `student@odl.local`, `prof@odl.local` — in the local `olapps` Keycloak realm, and correctly assigns the `admin` realm role to the admin user.

Also validated via `pre-commit` (`shellcheck` passes cleanly on the changed file).

**For reviewers**: on a machine where `/bin/bash` is bash 3.2 (any stock macOS), run `local-dev/scripts/kc-seed-users.sh` (or trigger the `kc-seed-users` Tilt resource) against a running local-dev Keycloak instance and confirm no `unbound variable` error occurs, and that the three test users are created/verified idempotently on repeat runs.

### Additional Context
Root cause was introduced by #4750 fka commit `fef2b977f` ("auto-seed Keycloak test users and make the seed idempotent"). This affects any developer running local-dev on stock macOS, not just one machine. No intentional behavior change beyond fixing the crash.

### DIFF
--- a/local-dev/scripts/kc-seed-users.sh
+++ b/local-dev/scripts/kc-seed-users.sh
@@ -92,16 +92,31 @@ ensure_admin_role
 
 # ---------------------------------------------------------------------------
 # User definitions: (username, email, firstName, lastName, assign_admin_role)
+#
+# Plain case statements instead of associative arrays: declare -A requires
+# bash 4+, but macOS ships bash 3.2 (GPLv3 licensing) as the only /bin/bash on
+# PATH, where declare -A silently misparses and fails with a confusing
+# "unbound variable" error under set -u.
 # ---------------------------------------------------------------------------
-declare -A USER_EMAILS=( [admin]="admin@odl.local" [student]="student@odl.local" [prof]="prof@odl.local" )
-declare -A USER_FIRST=( [admin]="Admin" [student]="Student" [prof]="Professor" )
-declare -A USER_LAST=( [admin]="User" [student]="User" [prof]="User" )
-ADMIN_USERS=("admin")
+user_email() {
+    case "$1" in
+        admin) echo "admin@odl.local" ;;
+        student) echo "student@odl.local" ;;
+        prof) echo "prof@odl.local" ;;
+    esac
+}
+user_first() {
+    case "$1" in
+        admin) echo "Admin" ;;
+        student) echo "Student" ;;
+        prof) echo "Professor" ;;
+    esac
+}
 
 for USERNAME in admin student prof; do
-    EMAIL="${USER_EMAILS[$USERNAME]}"
-    FIRST="${USER_FIRST[$USERNAME]}"
-    LAST="${USER_LAST[$USERNAME]}"
+    EMAIL="$(user_email "$USERNAME")"
+    FIRST="$(user_first "$USERNAME")"
+    LAST="User"
 
     # Check if user already exists. Look up by email, not username: the realm
     # has registrationEmailAsUsername=true, so Keycloak stores the username as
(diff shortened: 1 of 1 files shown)


ChristopherChudzicki (approved):
👍 

Aside: Apparently my first `bash` in PATH is `/opt/homebrew/bin/bash` 5.3, not macos system bash, which is probably why I didn't run into this.

Ardiea (approved):

PR #5283, opened 2026-08-05
Keep people out of the Pulumi estate; repo access comes from teams

blarghmatey:

N/A — follows PR #5233, tracked in witan under `wp-import-the-mitodl-github-organization-into-pulum-47211a`.

Two decisions that reinforce each other, plus the tooling and audit changes they imply. No GitHub state is modified.

**Pulumi manages teams, not individuals.** Neither `Membership` (who belongs to the org) nor `TeamMembership` (who belongs to a team) is imported. The organization payload drops **199 → 15** resources — 14 `Team` plus `OrganizationSettings`.

**Repo access comes from team membership only.** No individual holds a direct permission on a repository. `TeamRepository` is the sole path from a person to a repo, and all **169** of those stay managed.

The second rule is what makes the first one safe. Modelling people would only buy a security property if access flowed *through* those records — it doesn't. So what it actually bought was a declared roster, in exchange for a PR and a manual-approval apply on every hire, departure and team move. It also removes the highest-blast-radius resource in the estate: deleting a `Membership` evicts a human from the org, and not modelling something is a stronger guarantee than `protect=True`. The 39/39 seat ceiling stops being Pulumi's problem for the same reason — it can't fail on the 40th hire if it never adds one.

Given up, stated plainly: org membership and team rosters are **not declared in code**, so drift there is visible but not enforced. The crawl still records both — inventory feeding the audit, not Pulumi state.

#### This reverses an earlier conclusion

§9.3 concluded on 2026-08-03 to *keep* `Membership` and accept the onboarding cost. That section is marked superseded rather than rewritten, because the reasoning was sound on its own terms — what changed is a premise it treated as fixed. It also established things that still hold: the live seat count, the billing question, and audit rule DX-10, which matters **more** now that the audit is the only thing watching the ceiling.

#### The direct-collaborator backlog

The fleet does not satisfy rule 2 yet. Measured across all 316 repos:

| | |
|---|---:|
| Repos with ≥1 direct collaborator | **72** (43 active, 29 archived) |
| Total direct grants | **83** |
| — at `admin` | **73** |
| — at `write` | 10 |
| Distinct people | 20 |

**73 of 83 direct grants are `admin`.** The informal path to access isn't a lesser permission, it's the highest one. One person holds direct grants on 20 repos; a bot account (`odlbot`) holds admin on 7.

The crawl now records **who** holds each grant and at what level rather than counting them, so SEC-06 produces a migration list with a name and a target team instead of a number. The rule is fleet-wide — forks and archived included, no exemptions — since an `admin` grant on an archived repo is inert only until someone unarchives it.

Remediation is **phase 5**: it changes access, so it belongs downstream of a reviewed backlog rather than happening on the way past.

One convenient consequence: `RepositoryCollaborators` has no documented import in the provider schema (§5.2 flagged it as untrusted). Under rule 2 there is eventually nothing to import, so a provider gap stops mattering instead of needing a workaround.

### How can this be tested?

Requires the sops key for `src/bridge/secrets/pulumi/github_app.yaml`. Read-only.

```sh
uv run bin/github-org-inventory report --refresh
# -> "Direct collaborators (SEC-06)" section: 83 grants across 20 people, 73 admin / 10 write
# -> SEC-06 now reports "72 (23% of all repos)" rather than an active-only count

uv run bin/github-org-inventory imports
#    15 resources -> import-organization.json      (was 199)
#  1067 resources -> import-repositories.json      (unchanged)
```

Verified: the org payload contains only `Team` × 14 and `OrganizationSettings` × 1, with no `Membership` or `TeamMembership` entries; `import-repositories.json` still carries 169 `TeamRepository`; `crawl` regenerates the 316 per-repo files byte-identical, so none of this changes the fleet data.

### Additional Context

`_findings()` now has three scopes rather than two — active (default), archived (CON-07), and fleet-wide (SEC-06) — with each declared in a frozenset so the percentage denominator matches the population it is reporting on. Getting that wrong is how "94 of 138 archived" once rendered as "52% of active".

### DIFF
--- a/bin/github-org-inventory
+++ b/bin/github-org-inventory
@@ -123,6 +123,14 @@ DEFAULT_BRANCH_EXEMPT_ARCHETYPES = frozenset({"fork", "archived"})
 # share of active repos overstates them badly (92 of 138 archived reads as "52% of active").
 ARCHIVED_SCOPED_RULES = frozenset({"CON-07 archived repo still grants team write"})
 
+# Rules whose population is the WHOLE fleet, active and archived alike. SEC-06 is here
+# because the no-direct-collaborators policy (2026-08-05) admits no exceptions: an admin
+# grant on an archived repo is inert only until someone unarchives it, and the fork fleet
+# was explicitly included when the policy was set.
+FLEET_SCOPED_RULES = frozenset(
+    {"SEC-06 direct collaborator on repo (policy: repo access via teams only)"}
+)
+
 # The value GitHub reports for an active security-and-analysis feature. Hoisted to a
 # constant because comparing the scanning-status fields against a literal trips both
 # ruff S105 and detect-secrets as a hardcoded-credential false positive.
@@ -239,6 +247,13 @@ def _crawl_repo(token: str, listed: dict[str, Any]) -> dict[str, Any]:
         # IDs, not just counts: `imports` needs them to build the Pulumi import payload.
         hooks = _maybe(client, f"/repos/{ORG}/{name}/hooks?per_page=100") or []
         deploy_keys = _maybe(client, f"/repos/{ORG}/{name}/keys?per_page=100") or []
+        direct_collaborators = (
+            _maybe(
+                client,
+                f"/repos/{ORG}/{name}/collaborators?affiliation=direct&per_page=100",
+            )
+            or []
+        )
         # A SAMPLE of environment names, not the full set -- mitxonline has 412 and the
         # rest are ephemeral review apps. The count below stays authoritative; these
         # names exist to tell a real deploy target from publishing plumbing.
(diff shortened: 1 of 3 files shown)


blarghmatey:
Added two audit rules for team grants (9edad76c0), since with direct grants going away these **are** the access control rather than one input to it.

### CON-04 — team grants vs archetype

**174 of 176 active repos diverge. Two conform.**

`report` now loads `archetypes.yaml` to evaluate it, and skips the rule with a note when that file is absent rather than scoring it zero — an empty finding and an unmeasured one look identical in a report, and only one of them is good news.

### SEC-15 — permission ceiling on broad teams

`odl-engineering` may hold `push` or `maintain`, never `admin`. It is the broad engineering team, and `admin` carries repo deletion, settings changes and ruleset bypass. **It currently holds admin on 116 repos** (61 active, 55 archived). The team list is a constant, so the policy can grow without touching the rule.

### The spread underneath both is the real finding

```
## Team permission spread — 266/323 grants are admin
       266  admin
        35  push
        17  maintain
         4  pull
         1  triage
  teams holding admin, by repo count:
       116  odl-engineering  <-- SEC-15
        60  odl-engineering-owners
        57  arbisoft-contractors
        26  devops
```

**Admin is not the exception in this fleet, it is the default** — 82% of all team grants. Which is the thing worth knowing before making teams the sole access path: routing everything through teams only improves matters if the team grants themselves are least-privilege. Right now they mostly aren't.

Combined with SEC-06 (83 direct grants, 73 of them admin), the pattern is consistent across both mechanisms: whatever the path to access, it tends to arrive at admin.

### One self-inflicted bug worth noting

SEC-15 first rendered as "66% of active" while scanning all repos — the same denominator-versus-population mismatch the scope frozensets were built to prevent, reintroduced by adding a rule without registering its scope. Caught by cross-checking the count against the raw data. Fixed by adding it to `FLEET_SCOPED_RULES`; it now reads 37% of all repos.

---

Also confirmed for the earlier ask: **no membership data files were ever created.** All 319 tracked files under `substructure/github/` were checked against the 39 org members and 39 team-roster entries — zero mentions. The `teams:` key holds team slugs and permission levels (`TeamRepository` grants), not people. Members and rosters exist only in the gitignored crawl cache, feeding the audit rather than Pulumi state.

PR #5318, opened 2026-08-07
feat(sentry): drive project display names from ol_types Services

### LINKED ISSUE #4883: Sentry project naming Reality Realignment
### Description/Context
Our Sentry projects currently use names that don't match the actual product names we use internally and publicly. This creates friction when triaging errors, investigating incidents, and onboarding new engineers — you need to know the undocumented mapping between informal Sentry project names and real product names.

### Plan/Design
- Audit all current Sentry project names across our organization
- Document the discrepancies between Sentry project names and actual product/application names
- Rename Sentry projects to align with the names we actually use
- Update any infrastructure configuration that references old project DSNs, alert rules, or team assignments
- Communicate changes to affected teams so they can update local environment and secrets references

feoh:
Closes https://github.com/mitodl/ol-infrastructure/issues/4883

Sentry project display names were literal strings that `bin/import-sentry-config` copied out of live Sentry, so they had drifted from the service names used everywhere else in this repo. This makes `Services` in [`src/ol_infrastructure/lib/ol_types.py`](https://github.com/mitodl/ol-infrastructure/blob/main/src/ol_infrastructure/lib/ol_types.py) the single source of truth for them.

Six display names change:

| Sentry slug (unchanged) | Old name | New name |
|---|---|---|
| `micromasters` | `MicroMasters` | `micromasters` |
| `odl-video-service` | `ODL Video Service` | `ovs` |
| `open` | `Discussions` | `open` |
| `openedx-mitxonline` | `openedx-mitxonline` | `mitxonline-edx` |
| `openedx-mitxpro` | `openedx-xpro` | `xpro-edx` |
| `openedx-residential` | `openedx-residential` | `mitx-edx` |

The other nine projects (`dagster`, `learn-ai`, `mit-learn`, `mitxonline`, `ocw-site`, `ocw-studio`, `ol-analytics-api`, `xpro`, and the `release-script` exception) already matched their canonical names, so they now read from `Services` without a visible rename.

**Slugs are deliberately untouched.** Renaming them would break the `*_sentry_dsn` consumers, the Sentry-app issue alerts that pulumiverse-sentry cannot manage, and the Rootly alert route that matches on the `openedx-residential` project name. Every Sentry project URL stays put, which does mean a display name can differ from its slug — that is expected and documented.

Two supporting changes:

- **`Services.ocw_site`** is added. It was the only active Sentry project with no canonical service name. `release-script` deliberately does *not* get one, since that service is being phased out; it stays a literal via a documented exception.
- **The generator now fails closed.** `bin/import-sentry-config` maps each live Sentry slug to a `Services` member and raises on a project it has never seen. A newly created Sentry project now forces an explicit decision about its canonical name instead of silently reintroducing a literal one.

### How can this be tested?

`pulumi preview` on the `ol-infrastructure-sentry` `default` stack shows exactly the six display-name updates and nothing else:

```
Resources:
    ~ 6 to update
    339 unchanged
```

with each step being a lone `name` property change, e.g.

```
~ sentry:index/sentryProject:SentryProject: (update)
    [id=openedx-residential]
  ~ name: "openedx-residential" => "mitx-edx"
```

No replacements, creates, deletes, slug changes, or key/DSN changes.

Repo validation:

```bash
uv run pytest tests/                       # 458 passed
uv run mypy src/ol_infrastructure/lib/ol_types.py \
            src/ol_infrastructure/infrastructure/sentry/__main__.py   # clean
uv run ruff format --check <changed files> # clean
uv run pre-commit run --all-files          # all Python hooks pass
```

`pre-commit` also reports two failures that are pre-existing on `main` and touch no file in this PR: `packer_fmt` (the `packer` binary is not installed locally) and `hadolint` warnings on `dockerfiles/`.

Three new tests in `tests/test_import_sentry_config.py` cover the generator contract: a mapped project renders `name=Services.<member>` while keeping its literal slug, `release-script` keeps its literal name and emits a phase-out warning, and an unmapped project raises.

### Additional Context

- **Deploy step:** the renames only reach live Sentry on `pulumi up` against the `ol-infrastructure-sentry` `default` stack. Nothing else in this PR needs deploying, and nothing depends on it landing first.
- **`Services.odl_video_service` is `ovs`**, not `odl-video-service`. Using the enum literally is intentional here — changing the enum value itself would ripple into Kubernetes labels, Vault role names, and application resource naming, which is well outside this issue. The tradeoff is a terse Sentry display name.
- **The Open edX names** (`mitxonline-edx`, `xpro-edx`, `mitx-edx`) come from the existing service ontology rather than from the legacy `openedx-*` Sentry slugs, which is why those three names diverge most visibly from their slugs.
- **One drive-by fix:** `Services.learn_ai` had a stray trailing comma making it the 1-tuple `("learn-ai",)`. `StrEnum` unpacked that as constructor args to the identical string, so **the value is unchanged** and this is cosmetic — it is included because the tuple form would have broken the `Services.<member>` mapping work here if anyone had relied on it.
- **Preview guidance:** `IMPORT_SUMMARY.md` previously recommended `pulumi preview --refresh --diff` unconditionally. `--refresh` re-reads all 345 managed resources through pulumiverse-sentry and reliably trips Sentry's rate limits (HTTP 429; 40 req/s and 25 concurrent per endpoint), failing the preview outright on repeat runs. It is now scoped to the post-import drift check, with plain `pulumi preview --diff` documented as the default for code review.

### DIFF
--- a/bin/import-sentry-config
+++ b/bin/import-sentry-config
@@ -18,6 +18,7 @@ import cyclopts
 import httpx
 
 from bridge.secrets.sops import read_yaml_secrets
+from ol_infrastructure.lib.ol_types import Services
 
 app = cyclopts.App(help=__doc__)
 
@@ -44,6 +45,25 @@ RESOURCE_TOKENS = {
     ),
     "plugin": "sentry:index/sentryPlugin:SentryPlugin",
 }
+SENTRY_PROJECT_SERVICES: dict[str, Services] = {
+    "dagster": Services.dagster,
+    "learn-ai": Services.learn_ai,
+    "micromasters": Services.micromasters,
+    "mit-learn": Services.mit_learn,
+    "mitxonline": Services.mitxonline,
+    "ocw-site": Services.ocw_site,
+    "ocw-studio": Services.ocw_studio,
+    "odl-video-service": Services.odl_video_service,
+    "ol-analytics-api": Services.ol_analytics_api,
+    "open": Services.mit_open,
+    "openedx-mitxonline": Services.mitxonline_edx,
+    "openedx-mitxpro": Services.mitxpro_edx,
+    "openedx-residential": Services.mitx_edx,
+    "xpro": Services.xpro,
+}
+SENTRY_PROJECT_NAME_EXCEPTIONS = {
+    "release-script": "the release-script service is being phased out",
+}
 
 
 class PyExpr(str):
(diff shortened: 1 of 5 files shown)


feoh:
The cost is way higher than the value provided by this change so I'm abandoning this.

PR #5376, opened 2026-08-11
fix issues with LOCAL_DEV_ROOT_DOMAIN not being respected

gumaerc:
N/A

This PR makes a few changes to the local-dev setup based on some testing setting `LOCAL_DEV_ROOT_DOMAIN` to a custom value. There were some issues, primarily with Keycloak that this PR aims to fix.

### How can this be tested?
Spin up local-dev with `LOCAL_DEV_ROOT_DOMAIN` set to a custom value and ensure that you can log in to any of the sites that use Keycloak.

### DIFF
--- a/Tiltfile
+++ b/Tiltfile
@@ -11,7 +11,6 @@ config.define_bool("per_app_databases", usage="Deploy isolated DB/Valkey per app
 # local-dev/apps/openedx/Tiltfile is not yet included by the APPS loop below.
 config.define_string("openedx_mode", usage="qa (default) or local (Tutor)")
 config.define_string_list("prebuilt_tags", usage="Prebuilt image tag overrides per app, e.g. mit-learn=0.62.0 learn-ai=0.28.3")
-config.define_string("root_domain", usage="Root DNS domain for all local-dev services (default: mit.dev). Overrides LOCAL_DEV_ROOT_DOMAIN env var.")
 config.define_string("disk_keep_tags", usage="Newest tilt-built image tags kept per repo by the disk janitor (default: 3). Overrides LOCAL_DEV_DISK_KEEP_TAGS env var.")
 config.define_string("disk_buildcache_max_gb", usage="Docker build-cache size cap in GB (default: 10% of total disk; 0 disables). Overrides LOCAL_DEV_BUILDCACHE_MAX_GB env var.")
 cfg = config.parse()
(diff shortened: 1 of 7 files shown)


local-dev/infra/apps_infra/__main__.py
```diff
@@ -38,6 +39,25 @@
 keycloak_hostname = config.get("keycloak_hostname") or f"sso.ol.{root_domain}"
 keycloak_url = config.get("keycloak_url") or f"https://{keycloak_hostname}"
 
+# Client redirect URIs in modules/keycloak.py are written against the default
+# mit.dev domain. Rewriting them here rather than there keeps this patch off a
+# high-churn file and picks up any clients added upstream later.
```
ChristopherChudzicki: **Opinion:** Just update `create_olapps_dev_realm` with a kwarg and construct the domains there. It seems simpler.

The rationale
> and picks up any clients added upstream later
is only true for the two fields targeted here `("validRedirectUris", "validPostLogoutRedirectUris")`

E.g., Claude pointed out to me that the keycloak configs could have been easily rwritten with `root_url="https://learn.mit.dev", valid_redirect_uris=["/*"], web_origins=["/*"]` which this would replace.

Generally, constructing by an f-string in `create_olapps_dev_realm` seems more explicit 🤷

ChristopherChudzicki: IF YOU WANT TO KEEP IT HERE (not `create_olapps_dev_realm`): Then omit the `f root_domain != "mit.dev":` check. If you omit the check, the coe runs and is a no-op. That's better, because it makes any future issues more discoverable.

ChristopherChudzicki (changes requested):
Right now the root domain is comes from two places: `tilt_config.json`'s `root_domain`, and env value `LOCAL_DEV_ROOT_DOMAIN`. This split isn't new to this PR.
- This approach root domain when it comes from `LOCAL_DEV_ROOT_DOMAIN`.
- But not if it comes from `tilt_config.json`

The difficulty with `tilt_config.json` is passing it through the `include(...)` scripts in Tiltfile.

**Suggestion:** Drop `root_domain` from `til_config.json` entirely. If someone wants to set it, they need to make sure it's in their host env.

MITOL_WORKSPACE_ROOT behaves this way right now, i believe, though I've never set it.

**Additionally:** Claude suggested adding a check in Tiltfile along the lines of "check if /etc/hosts has the root_domain; if it doesn't, error loudly and instruct user to run `./setup.sh` (which would set /etc/hosts line)"

Tiltfile
```diff
@@ -20,12 +19,28 @@ enabled_apps = cfg.get("enabled_apps", ["mit-learn", "learn-ai", "mitxonline", "
...
+root_domain = os.environ.get("LOCAL_DEV_ROOT_DOMAIN", "mit.dev")
+
+# The TLS certificate covers only this domain, so an unresolvable hostname means
+# nothing the stack serves is reachable. DNS and /etc/hosts both satisfy this.
+_probe_host = "sso.ol." + root_domain
+if str(local(
+    "getent hosts %s >/dev/null 2>&1 && echo ok || echo missing" % _probe_host,
+    quiet=True,
+    echo_off=True,
+)).strip() != "ok":
```
ChristopherChudzicki, line 33: macOS doesnt have `getent`

Claude suggested

```python
if str(local(
    "python3 -c 'import socket,sys; socket.getaddrinfo(sys.argv[1], None)' %s >/dev/null 2>&1 && echo ok || echo missing" % _probe_host,
    quiet=True, echo_off=True,
)).strip() != "ok":
```

ChristopherChudzicki (approved):
👍 One request, but no need for re review

PR #5392, opened 2026-08-12
fix(release-bot): report the real release version, not a stale one

blarghmatey:
N/A

Companion to https://github.com/mitodl/ol-concourse/pull/88, which fixes the release resource itself. **That PR must merge and publish first** — see Checklist.

`mitodl/ol-analytics-api` has not cut a new release since **2026-08-03**. Every deployment since — RC and Production — has shipped `ref=2026.8.3.1`.

Three defects here each produced that same symptom independently, so fixing any one alone still leaves the version wrong.

#### 1. The calver tag pattern could never match a real tag

`github_client.py` matched release tags with `^\d{4}\.\d{2}\.\d{2}\.\d+$`, requiring zero-padded month and day. The release resource emits PEP 440 calver — `2026.8.3.1`, not `2026.08.03.1` — because uv and other modern Python tooling reject leading zeros.

So `_latest_release_tag()` matched **no tag any current release has ever produced**, returned `None`, and `/doof release-notes` silently fell back to listing the last 50 commits of all history. The existing fixtures only ever used padded tags (`_FakeTag("2026.07.01.1")`), which is why the suite stayed green.

#### 2. The latest tag was whichever matched first

`_latest_release_tag()` returned the first regex match from `repo.get_tags()`. That ordering is not documented to be chronological, so the release baseline depended on it. Now selected by parsed version — which also fixes `2026.12.1.1` sorting below `2026.8.3.1` lexically.

#### 3. `/doof release` never refreshed the version

The release resource is configured `check_every: never` with no `webhook_token`, so Concourse checks it **only** when explicitly asked. `_cmd_release` called `trigger_job` directly, so the build job bound whatever version Concourse last recorded — the *previous* release — and reused its version number and commit list.

It now checks `<app>-release` first and **refuses to trigger** if that check fails, rather than cutting a release from stale state.

#### Release-machinery commits excluded

`Release X` and `Merge releases/X` are dropped from release notes, for the same reason the resource drops them: the tag sits on the pre-bump HEAD and `finish` lands both on the default branch afterwards, so a finished release otherwise reads as two commits waiting to be released.

#### New visibility and control

- **`/doof preview <app>`** — the version the next release would get and the commits it would contain, with **no side effects at all** (no resource check, no job trigger). Tested explicitly.
- **In-flight releases surfaced.** A `releases/X` branch that outlived its production deploy means `finish` never completed. `release-status` now reports it (with age and checklist progress), and `/doof release` warns what it is about to supersede instead of deleting refs silently.
- **`/doof wait-for-checkboxes <app>`** — ports the [release-script bot's command](https://github.com/mitodl/release-script/blob/master/bot.py). It names *who* is still outstanding rather than reporting a bare count, and thanks people as they check off. The final ready-to-promote message and button are left to the existing `_notify_ready_to_promote` poll loop, which already covers every open release and dedupes on the `promote-ready` label — posting them here too would double up.

#### Pipeline: the `try` that hid the failure

`_build_release_resource_app_pipeline` wrapped the production job's `action: finish` put in a `TryStep`. That was added for a real reason: the job can be retriggered by an infra-only merge with no new app release, and `finish` then failed trying to fetch a release branch a prior finish had already deleted.

The release resource is now idempotent in that case, so the `try` buys nothing — and it equally swallowed **genuine** finish failures. That is how ol-analytics-api sat with an unmerged `releases/2026.8.3.1` branch, and a frozen release version, from 2026-08-03 onward with no red build to show for it. A failure there now means a real release that never finished, and stops the job.

#### Tests for `bot.py`

`bot.py` had no tests at all because `slack_bolt` is a dependency of the release_bot **container image** (`requirements.txt`), not of `ol-infrastructure`, so it is absent from the repo's test environment. Rather than pull a runtime dep into the repo, the test conftest registers stand-ins for the two names `bot.py` binds at import time. Every handler takes its Slack callables as arguments, so they drive directly.

### How can this be tested?

**Automated:** 224 tests pass (`uv run pytest tests/ol_infrastructure/applications/release_bot/ tests/ol_concourse/ -q`) — 43 in `test_github_client.py` (up from 19) and 13 new in `test_bot.py`. `pre-commit run` passes including mypy and detect-secrets.

Highest-value regression tests, each pinned to one of the three defects:

- `test_latest_release_tag_finds_unpadded_calver` — the padding bug the old fixtures hid.
- `test_latest_release_tag_picks_highest_not_first` / `..._compares_numerically_not_lexically`.
- `test_release_checks_the_resource_before_triggering` — asserts strict call *order*: `check_resource("my-app-pipeline", "my-app-release")` then `trigger_job(..., "build-my-app-release-image")`.
- `test_release_refuses_to_trigger_when_the_check_fails` — a failed check must not fall through to a release.
- `test_preview_triggers_nothing` — asserts `check_resource` and `trigger_job` are both never called.

**Pipeline generation:** confirmed the production job's plan now ends in a bare put rather than a `try`:

```json
{
  "params": {
    "action": "finish",
    "repo_dir": "ol-analytics-api-main",
    "version_file": "ol-analytics-api-release/version"
  },
  "put": "ol-analytics-api-release"
}
```

Reproduce with:

```bash
uv run python -c "
from ol_concourse.pipelines.infrastructure.k8s_apps.pipeline import build_app_pipeline
import json
d = build_app_pipeline('ol-analytics-api').model_dump(exclude_none=True)
print(json.dumps([j for j in d['jobs'] if 'production' in j['name'].lower()], indent=1))
"
```

**Reviewer validation of the premise** — all public:

```bash
# the release branch that never merged
gh api repos/mitodl/ol-analytics-api/compare/main...releases/2026.8.3.1 \
  --jq '{status, ahead_by, behind_by}'
# -> {"status":"diverged","ahead_by":1,"behind_by":1}

# every deployment stuck on the same ref
gh api "repos/mitodl/ol-analytics-api/deployments?per_page=10" \
  --jq '.[] | "\(.environment) ref=\(.ref) \(.created_at)"'
# -> Production ref=2026.8.3.1 2026-08-12T15:27:52Z  (and RC back to 2026-08-06)
```

The unpadded-tag bug is checkable directly:

```python
import re
re.match(r"^\d{4}\.\d{2}\.\d{2}\.\d+$", "2026.8.3.1")   # -> None   (the bug)
re.match(r"^(\d{4})\.(\d{1,2})\.(\d{1,2})\.(\d+)$", "2026.8.3.1")  # -> matches
```

**After deploy:** `/doof preview ol-analytics-api` should report the real next version and only genuinely unreleased commits, and flag `2026.8.3.1` as in flight. `/doof release-status` should show the same in-flight warning.

### Additional Context

- The release bot runs a **single replica** — Socket Mode connections are not multiplexed. `_slack_app` is therefore a module global, set by `create_app()`; the checkbox watcher needs it because a watch outlives the 30-minute window a `respond` URL stays valid.
- Duplicate `/doof wait-for-checkboxes` invocations for the same app are rejected rather than starting a second watcher, which would post every progress message twice.
- The in-flight lookup is decorative on `/doof release` — if GitHub is unreachable the release still proceeds, it just loses the supersede warning. Covered by `test_release_still_triggers_when_the_in_flight_lookup_fails`.

### Checklist:
- [ ] Merge https://github.com/mitodl/ol-concourse/pull/88 and let the `mitodl/concourse-release-resource` image publish **before** merging this PR — removing the `TryStep` is only safe once `action: finish` is idempotent.
- [ ] Re-set the affected app pipelines (`fly sp`) so the regenerated plan takes effect.
- [ ] Redeploy the release bot so the new commands and tag matching are live.


blarghmatey:
Thanks — most of these were real. Fixed in cb05468d0.

**`check_resource` didn't wait** (`bot.py:74`) — confirmed, and the most important finding here: it defeated the central fix of this PR. The check endpoint answers `201` as soon as the build is *created*, so `trigger_job` could still run before the new version was recorded and bind the stale one. `check_resource` now polls the returned build to a terminal status and raises on anything but `succeeded`, with a 180s timeout so a wedged check surfaces instead of hanging the Slack handler. A bodyless response (older Concourse) keeps the previous fire-and-forget behaviour, since there is nothing to wait on. `concourse_client.py` had no tests at all; it now has five covering wait, failure, error, timeout, and the bodyless case.

**100-tag truncation** (`github_client.py:116`, `:225`) — confirmed, and it directly contradicted the fix immediately above it: selection sorts by parsed version *precisely because* `get_tags()` ordering isn't guaranteed, so truncating an unordered listing can hide the highest tag and silently restore the stale baseline. Both call sites now share `_release_tags()`, the cap is well past what release repositories reach, and reaching it logs a warning instead of quietly returning a wrong answer. `test_release_tags_scans_the_whole_listing` uses 141 tags with the maximum last.

**Watcher double-posted "ready to promote"** (suppressed, `bot.py:367`) — confirmed, and it contradicted the docstring I wrote saying it delegates. `_notify_ready_to_promote` already posts one *with the promote button*, deduped by label, so the channel got two notifications and the first had no button. The watcher now stops silently.

**Watcher registry TOCTOU** (suppressed, `bot.py:442`) — confirmed. The slot is now reserved in `_checkbox_watch_preflight` before any `await` (nothing suspends between the check and the add, so it's atomic on the event loop), and released on every path that then declines to start a watcher. Two tests: one asserts the slot is already claimed *during* the GitHub lookup, one asserts it's released when no watcher starts.

**Refresh `continue` without sleep** (`bot.py:381`) — partially. I've added the sleeps so every path has the same cadence, but the severity was overstated: after that `continue` the loop top either sleeps in its own exception handler or falls through to the main sleep, so the worst case was one extra immediate API call, not the tight loop described.

Test count 56 → 67.

PR #5397, opened 2026-08-13
Update local-dev to print hosts for --skip-hosts

rhysyngsun:

This adds output to `local-dev/scripts/setup.sh` so that if you specify `--skip-hosts` it prints out the hosts that it otherwise would've added and informs you you're responsible for manually setting them up as otherwise tilt will fail.

### How can this be tested?

You can run `./local-dev/scripts/setup.sh --skip-hosts` and observe the logging. I can't personally test the path without `--skip-hosts` because it doesn't work on my system, so someone else should probably test that or at least heavily scrutinize the code.


gumaerc (approved):
👍 LGTM

PR #5417, opened 2026-08-14
fix(apisix): restart the login flow instead of 500ing on an expired auth session

blarghmatey:

Part of https://github.com/mitodl/hq/issues/12178 — specifically part (1a). Parts (1b) and (2) stay open.

Follows https://github.com/mitodl/ol-infrastructure/pull/5413, which added the alerting for these failures.

Adds `oidc_error_callback_recovery_plugin()` and attaches it to the shared plugin configs for `api.learn.mit.edu`, `mitxonline.mit.edu`, and `nb.learn.mit.edu`.

- When a Keycloak authentication session expires — user left the login tab open, or followed a stale bookmark — Keycloak redirects back with `error=temporarily_unavailable` and no `code`, marking its own event `restart_after_timeout="true"`. Its intent is that the client start over.
- `lua-resty-openidc` treats any `error` parameter as fatal, so APISIX serves a 21KB HTTP 500 where the user expected a login form. Nothing retries.
- The new plugin runs as a `serverless-pre-function` in the **rewrite** phase, where its priority (10000) outranks `openid-connect`'s (2599) — confirmed by their relative order in [`APISIX_DEFAULT_PLUGINS_3_17`](https://github.com/mitodl/ol-infrastructure/blob/main/src/ol_infrastructure/infrastructure/aws/eks/apisix_official.py#L46-L56). It redirects back into the authorization flow before the plugin can fail.
- The redirect target is derived from the callback URI rather than configured. The callback always sits at `<login prefix>/.apisix/redirect` and the route serving it is by construction the `unauth_action="auth"` one, so one attachment per host covers every route group on it — including mit-learn's `/login` and `/learn/login` prefixes.
- Only transient errors are recovered. `access_denied` (user pressed Cancel) and `invalid_request` (a real misconfiguration) keep failing, otherwise the browser spins between the gateway and Keycloak.
- A guard cookie caps recovery at one attempt per browser per minute, so a persistently broken IdP surfaces as an error rather than a redirect loop.

**Why this cause and not the others.** #5413 shipped with the caveat that its 500 rate is a rate over *requests*, not users, and that real impact was unknown. Keycloak's event stream answers it — unlike the gateway access log, it identifies users and sessions. The 1,251 callback 500s/day split three ways:

| Cause | Volume/day | Blocks anyone? |
|---|---|---|
| `error=temporarily_unavailable` (expired auth session) | 614 | **Yes** — 530 distinct client addresses, 181 of which never reached a successful callback in the same 24h |
| Stale-code replay | ~465 | **No** — the first callback already succeeded |
| Genuine code-exchange failure | ~171 | Yes, but small: 2.0% of api.learn first attempts, ≤6.5% of mitxonline's |

Keycloak settles the replay question: **4,762 successful `code_to_token` exchanges against only 7 `invalid_code` errors** in 24h. If replayed callbacks reached the token endpoint we would see thousands. They die at APISIX's state check instead. That is why the 27.4% / 14.4% callback failure rates in the original audit are not login failure rates, and why only the first row here is worth fixing now.

Per-host the split is exact — for api.learn, 327 + 191 + 55 = 573, the measured total.

### How can this be tested?

Three layers, each covering something the one below it cannot. All run locally on this branch; the first two are wired as CI jobs alongside the existing `test` job.

**1. Unit — `uv run pytest tests/ol_infrastructure/components/services/` (132 pass)**
Config rendering, plus `test_apisix_lua.py` executing the shipped Lua in a `lupa` runtime against a stubbed `ngx`/`apisix.core`. Fast feedback on the Lua's logic; it does not prove APISIX accepts anything.

**2. Integration — `uv run pytest tests/apisix_integration -m integration` (11 pass, ~2s)**
A real APISIX 3.17.0 container in standalone mode, fed the route config built by the actual plugin helper. This is what proves the `oidc_error_recovery` block survives schema validation and is readable on `conf`, and that `Set-Cookie` survives `ngx.redirect` on the real runtime. Marked `integration`, so the repo's existing `-m 'not integration'` default keeps it out of the fast suite.

**3. test-nginx — `tests/apisix_testnginx/run.sh` (19 subtests pass)**
APISIX's own harness (`t/APISIX.pm`) on the same OpenResty build the gateway runs. Notes for anyone touching it:
- It needs **etcd**. `APISIX.pm` boots APISIX's real init path, so without etcd every test fails test-nginx's default `no [error] in error.log` check on connection-refused noise. etcd is in the image rather than filtering those lines out, because filtering would also hide errors worth seeing.
- It builds from the **repo root**, so the Dockerfile copies the Lua from `src/`. A second copy beside the `.t` would be free to drift, and the suite would then pass against Lua we do not ship.
- `t/APISIX.pm` and the eight cert fixtures it opens come from the APISIX source tarball at build time — one request, not nine from `raw.githubusercontent.com`, which reliably trips its rate limiter (HTTP 429) and made the build a coin flip.

**Mutation-tested, not assumed.** Each layer was verified to fail when the behaviour it covers is broken:

| Mutation | Effect |
|---|---|
| Remove the guard-cookie early return | 1 integration test fails |
| Simulate the settings block being stripped | 6 integration tests fail |
| Remove the repeated-`error` table handling | 2 test-nginx subtests fail |

**Lint:** `pre-commit` clean on the changed set, including hadolint on the new Dockerfile (verified separately with `hadolint` directly — the repo-wide `Lint Dockerfiles` failure is pre-existing across other Dockerfiles and is not from this branch).

**Config-passing chain**, checked at every hop against the pinned versions rather than assumed:

| Layer | Why an unknown key survives |
|---|---|
| `ApisixPluginConfig` v2 / `PluginConfig` v1alpha1 CRDs | `config` marked `x-kubernetes-preserve-unknown-fields: true` |
| apisix-ingress-controller | `Config apiextensionsv1.JSON` — raw bytes |
| ADC | `type Plugins map[string]any` |
| APISIX 3.17.0 `serverless/init.lua` | schema declares only `phase`/`functions`, no `additionalProperties` |
| runtime | invoked as `func(conf, ctx)` |

### Additional Context

- **Correction — the chronic alert will NOT go quiet, and an earlier version of this description wrongly said it would.** `APISIXOIDCCallbackFailureRateChronic` fires at >5% over 6h, and its own annotation says it stays up "until the two known causes are fixed". This PR fixes one. Working the measured numbers through:

  | host | callbacks/24h | 500 rate now | after this PR | chronic >5%? |
  |---|---|---|---|---|
  | api.learn.mit.edu | 3,297 | 17.4% | 7.5% | **still fires** |
  | mitxonline.mit.edu | 2,633 | 25.4% | 14.6% | **still fires** |

  The replay 500s alone keep both hosts above the threshold. Alert silence is therefore *not* the deployment check. Verify instead that the `temporarily_unavailable` callbacks stop returning 500 — they should go to ~0 while the count of callback requests carrying `error=` stays flat:

  ```logql
  sum by (host, status) (count_over_time(
    {namespace="operations", container="apisix"}
    |= ".apisix/redirect" |= "time_local" |= "error=" | logfmt | __error__="" [24h]))
  ```

  and that the overall rates land near the "after" column above. No rule change is included here — the rule behaves as designed, it is the claim about it that was wrong. Adding a `temporarily_unavailable`-specific expression so this cause is independently trackable is a reasonable follow-up if you want it.
- **On prioritising the rest of #12178:** on this evidence part (1b) (tolerating replayed codes) is log-noise and alert-inflation cleanup, not a user-facing fix, and is better done after 1a has been observed in production since removing 1a changes the remaining mix. Part (2) (per-issuer route scoping, Rootly INC-10) is untouched and still needs design.
- **Unrelated finding, filed separately:** Keycloak rejects ~12/day of `redirect_uri="http://nb.learn.mit.edu/.apisix/redirect"` with `invalid_redirect_uri` — note `http`, not `https`. Those fail at the authorization endpoint and never reach the callback, so they are invisible in the numbers above.
- **On the dismissed Seer finding:** it predicted (HIGH) that `ngx.header["Set-Cookie"]` would be dropped by `ngx.redirect()`, breaking the loop guard. It was dismissed last round on source reading ([`ngx_http_lua_ngx_redirect`](https://github.com/openresty/lua-nginx-module/blob/master/src/ngx_http_lua_control.c) pushes Location onto the *existing* `r->headers_out.headers` list and clears nothing). That is now settled empirically instead: `test_guard_cookie_survives_ngx_redirect` asserts the header on a response a real APISIX generates, and `test_second_failure_falls_through_instead_of_looping` asserts the guard actually engages. Its suggested `header_filter` fix would not have worked — `ngx.redirect()` terminates the request in `rewrite`, so that phase never runs for this response.

### Checklist:

Not merge blockers. The three test layers now cover the plugin's behaviour and APISIX's acceptance of its config; what they cannot reach is the real cluster's CRD → ingress-controller → ADC path and a live Keycloak. Worth confirming against `applications.mitlearn.CI` after this deploys there, before promoting to Production:

- [ ] The settings block survives the real config pipeline end to end: `curl -sSi 'https://api.ci.learn.mit.edu/login/.apisix/redirect?error=temporarily_unavailable&state=x'` returns **302** with `Location: /login/` and `Set-Cookie: apisix_oidc_recovery=1; ...`. A 500 here means the `oidc_error_recovery` key was dropped somewhere between the CRD and APISIX — the one hop no test covers.
- [ ] A normal login through https://ci.learn.mit.edu still completes end to end against real Keycloak — this plugin runs in the rewrite phase of every route on the host.

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -6,7 +6,11 @@ on:
     branches: [main]
   pull_request:
 
-permissions: {}
+# Every job here only reads the checkout: nothing publishes, comments, or writes
+# back. Set at the workflow level rather than per job so a job added later
+# inherits the restriction instead of silently getting the default token.
+permissions:
+  contents: read
 
 jobs:
   test:
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -34,6 +34,52 @@ jobs:
     - name: Run tests
       run: uv run pytest
 
+  # The unit suite stubs ngx and apisix.core, so it proves the Lua does what we
+  # meant but not that APISIX accepts the plugin config or that OpenResty
+  # behaves as assumed. These two jobs cover that, and are separate so a
+  # gateway-level failure is distinguishable from an ordinary test failure.
+  apisix-integration:
+    name: APISIX integration (real gateway)
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    steps:
+    - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
+      with:
+        persist-credentials: false
+
+    - name: Install uv
+      uses: astral-sh/setup-uv@20cfd1bf945f4377ade1205e4dbc17946fc9a30d # v10.0.1
+      with:
+        enable-cache: true
+
+    - name: Set up Python
+      uses: actions/setup-python@5fda3b95a4ea91299a34e894583c3862153e4b97 # v7
+      with:
+        python-version-file: pyproject.toml
+
+    - name: Install dependencies
+      run: uv sync --frozen
+
+    # Excluded from the default run by `-m 'not integration'` in pyproject.
+    - name: Run APISIX integration tests
+      run: uv run pytest tests/apisix_integration -m integration
+
+  apisix-testnginx:
+    name: APISIX test-nginx
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    steps:
+    - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
+      with:
+        persist-credentials: false
+
+    # Builds an image from the pinned APISIX version, adds Test::Nginx and
+    # etcd, and runs APISIX's own harness against src/.../files/*.lua.
+    - name: Run test-nginx suite
+      run: tests/apisix_testnginx/run.sh
+
   # ONE required status check standing for this whole workflow.
   #
   # Today this wraps a single job, so it changes nothing on its own. It exists for the
@@ -50,7 +96,7 @@ jobs:
   # required directly -- a workflow that does not trigger reports nothing at all and
   # would leave every unrelated PR pending forever.
   ci-gate:
-    needs: [test]
+    needs: [test, apisix-integration, apisix-testnginx]
     # Without this the gate is skipped when a dependency fails, and GitHub treats a
     # skipped required job as successful, allowing the PR to merge despite the failure.
     if: always()
--- a/src/ol_infrastructure/components/services/apisix.py
+++ b/src/ol_infrastructure/components/services/apisix.py
@@ -155,7 +155,15 @@ def oidc_error_callback_recovery_plugin(
     :returns: A ``serverless-pre-function`` plugin config to attach to routes.
     :rtype: OLApisixPluginConfig
     """
-    errors = recoverable_errors or ["temporarily_unavailable"]
+    # `is None`, not `or`: an explicit empty list means "recover nothing",
+    # which is how a caller turns the plugin into a no-op without detaching it
+    # from every route that references the shared config.  `or` would quietly
+    # turn that back into the default.
+    errors = (
+        ["temporarily_unavailable"]
+        if recoverable_errors is None
+        else recoverable_errors
+    )
     recoverable_table = ", ".join(f'["{error}"] = true' for error in errors)
     recovery_function = f"""return function(conf, ctx)
     local uri = ngx.var.uri
--- a/src/ol_infrastructure/components/services/apisix.py
+++ b/src/ol_infrastructure/components/services/apisix.py
@@ -1,6 +1,7 @@
 # ruff: noqa: E501
 """APISIX ingress controller components for Kubernetes."""
 
+from pathlib import Path
 from typing import Any, Literal
 
 import pulumi_kubernetes as kubernetes
@@ -13,6 +14,14 @@ from ol_infrastructure.components.services.vault import (
     OLVaultK8SStaticSecretConfig,
 )
 from ol_infrastructure.lib.pulumi_helper import parse_stack
+
+# Read once at import: the file is shipped verbatim as the serverless function
+# body, with configuration passed separately on the plugin config.
+OIDC_ERROR_RECOVERY_LUA = (
+    Path(__file__)
+    .parent.joinpath("files", "oidc_error_callback_recovery.lua")
+    .read_text()
+)
 
 
 class OLApisixPluginConfig(BaseModel):
@@ -145,9 +154,20 @@ def oidc_error_callback_recovery_plugin(
     looping: a persistently broken IdP should surface as an error, not as an
     infinite redirect.
 
+    The Lua itself lives in ``files/oidc_error_callback_recovery.lua`` and is
+    shipped verbatim -- nothing is interpolated into it.  Tunables travel as an
+    ``oidc_error_recovery`` block on the plugin config, which the function reads
+    off ``conf``: ``serverless/init.lua`` invokes each function as
+    ``func(conf, ctx)``, and its schema does not set ``additionalProperties``,
+    so extra keys validate.  Keeping it a real ``.lua`` file means it is
+    syntax-highlighted, reviewable, and testable under APISIX's own test-nginx
+    harness (``t/oidc_error_callback_recovery.t``).
+
     :param recoverable_errors: OAuth 2.0 ``error`` codes to restart the flow
         for.  Defaults to ``temporarily_unavailable``, which is 100% of what
-        production emits today.
+        production emits today.  An explicit empty list makes the plugin a
+        no-op, for turning it off without detaching it from every route that
+        references a shared plugin config.
     :param guard_cookie_name: Name of the loop-breaker cookie.
     :param guard_max_age: Seconds the guard cookie lives, bounding how often one
         browser can be sent back through login.
@@ -155,47 +175,6 @@ def oidc_error_callback_recovery_plugin(
     :returns: A ``serverless-pre-function`` plugin config to attach to routes.
     :rtype: OLApisixPluginConfig
     """
-    # `is None`, not `or`: an explicit empty list means "recover nothing",
-    # which is how a caller turns the plugin into a no-op without detaching it
-    # from every route that references the shared config.  `or` would quietly
-    # turn that back into the default.
-    errors = (
-        ["temporarily_unavailable"]
-        if recoverable_errors is None
-        else recoverable_errors
-    )
-    recoverable_table = ", ".join(f'["{error}"] = true' for error in errors)
-    recovery_function = f"""return function(conf, ctx)
-    local uri = ngx.var.uri
-    if not uri or not uri:match("%.apisix/redirect$") then
-        return
-    end
-    local core = require("apisix.core")
-    local args = core.request.get_uri_args(ctx)
-    if not args then
-        return
-    end
-    local err = args["error"]
-    if type(err) == "table" then
-        err = err[1]
-    end
-    local recoverable = {{{recoverable_table}}}
-    if not err or not recoverable[err] then
-        return
-    end
-    local cookie = ngx.var.http_cookie
-    if cookie then
-        for pair in cookie:gmatch("[^;]+") do
-            if pair:match("^%s*([^=%s]+)=") == "{guard_cookie_name}" then
-                return
-            end
-        end
-    end
-    core.log.warn("oidc callback error=", err, " uri=", uri, " restarting auth")
-    ngx.header["Set-Cookie"] = "{guard_cookie_name}=1; Path=/; Max-Age="
-        .. "{guard_max_age}" .. "; Secure; HttpOnly; SameSite=Lax"
-    return ngx.redirect((uri:gsub("%.apisix/redirect$", "")), 302)
-end"""
     return OLApisixPluginConfig(
         name="serverless-pre-function",
         secretRef=None,
@@ -204,7 +183,22 @@ end"""
         # outranks it (2599), so this gets to inspect the callback and bail out
         # before the plugin turns the error parameter into a 500.  In the access
         # phase it would run after openid-connect had already failed.
-        config={"phase": "rewrite", "functions": [recovery_function]},
+        config={
+            "phase": "rewrite",
+            "functions": [OIDC_ERROR_RECOVERY_LUA],
+            "oidc_error_recovery": {
+                # `is None`, not `or`: an explicit empty list means "recover
+                # nothing", and `or` would quietly turn that back into the
+                # default.
+                "recoverable_errors": (
+                    ["temporarily_unavailable"]
+                    if recoverable_errors is None
+                    else recoverable_errors
+                ),
+                "guard_cookie_name": guard_cookie_name,
+                "guard_max_age": guard_max_age,
+            },
+        },
     )
 
 
--- a/src/ol_infrastructure/components/services/apisix.py
+++ b/src/ol_infrastructure/components/services/apisix.py
@@ -15,13 +15,20 @@ from ol_infrastructure.components.services.vault import (
 )
 from ol_infrastructure.lib.pulumi_helper import parse_stack
 
-# Read once at import: the file is shipped verbatim as the serverless function
-# body, with configuration passed separately on the plugin config.
+# Read once at import: the files are shipped verbatim as serverless function
+# bodies, with configuration passed separately on the plugin config.
 OIDC_ERROR_RECOVERY_LUA = (
     Path(__file__)
     .parent.joinpath("files", "oidc_error_callback_recovery.lua")
     .read_text()
 )
+CANONICAL_HTTPS_REDIRECT_LUA = (
+    Path(__file__).parent.joinpath("files", "canonical_https_redirect.lua").read_text()
+)
+
+# The only statuses ngx.redirect accepts; anything else is a Lua error at
+# request time (ngx_http_lua_control.c:209-219).
+NGX_REDIRECT_STATUSES = (301, 302, 303, 307, 308)
 
 
 class OLApisixPluginConfig(BaseModel):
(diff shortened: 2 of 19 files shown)


tests/ol_infrastructure/components/services/test_apisix.py
```diff
@@ -347,6 +348,61 @@ def test_cleanup_plugin_honours_a_custom_stale_name():
...
+
+    assert '["temporarily_unavailable"] = true' in lua
+    assert '["server_error"] = true' in lua
+
+
+def test_recovery_plugin_ignores_non_callback_requests():
+    """Attached to a host's shared plugin config, this sees every route."""
+    (lua,) = oidc_error_callback_recovery_plugin().config["functions"]
+
+    assert 'uri:match("%.apisix/redirect$")' in lua
```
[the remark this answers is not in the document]
blarghmatey: Fair catch, and the harness existed — it caught a real `%%`-escaping bug during development — but I ran it manually and deleted it instead of committing, so the repo only had the substring assertions you are describing. Committed in e84547871 as `tests/ol_infrastructure/components/services/test_apisix_lua.py`: 13 cases loading the generated function into a `lupa` Lua runtime with a stubbed `ngx`/`apisix.core`, asserting on the actual redirect URI, status, and Set-Cookie. Covers pass-through on a `code` callback, `access_denied` not recovering, non-callback URIs, exact-cookie-name matching, the repeated-parameter table, and the loop guard. `lupa` is in-process so CI needs no interpreter. Confirmed it has teeth by mutation: deleting the guard-cookie early return fails 1 test, reintroducing the `%%` bug fails 7.

src/ol_infrastructure/components/services/apisix.py
```diff
@@ -103,6 +103,103 @@ def stale_session_cookie_cleanup_plugin(
...
+        for.  Defaults to ``temporarily_unavailable``, which is 100% of what
+        production emits today.
+    :param guard_cookie_name: Name of the loop-breaker cookie.
+    :param guard_max_age: Seconds the guard cookie lives, bounding how often one
+        browser can be sent back through login.
+
+    :returns: A ``serverless-pre-function`` plugin config to attach to routes.
+    :rtype: OLApisixPluginConfig
+    """
+    errors = recoverable_errors or ["temporarily_unavailable"]
```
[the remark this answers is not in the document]
blarghmatey: Correct — fixed in e84547871, now `if recoverable_errors is None`. An empty list is a meaningful value here (it makes the plugin a no-op without having to detach it from every route referencing the shared config), so it needs to be honoured rather than folded into the default. Covered both statically and at runtime by `test_an_empty_recoverable_list_disables_recovery`.

src/ol_infrastructure/components/services/apisix.py
```diff
@@ -103,6 +103,111 @@ def stale_session_cookie_cleanup_plugin(
...
+        for pair in cookie:gmatch("[^;]+") do
+            if pair:match("^%s*([^=%s]+)=") == "{guard_cookie_name}" then
+                return
+            end
+        end
+    end
+    core.log.warn("oidc callback error=", err, " uri=", uri, " restarting auth")
+    ngx.header["Set-Cookie"] = "{guard_cookie_name}=1; Path=/; Max-Age="
+        .. "{guard_max_age}" .. "; Secure; HttpOnly; SameSite=Lax"
+    return ngx.redirect((uri:gsub("%.apisix/redirect$", "")), 302)
```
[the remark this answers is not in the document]
blarghmatey: Verified against the implementation — this is a false positive, no change made.

`ngx_http_lua_ngx_redirect` ([src/ngx_http_lua_control.c](https://github.com/openresty/lua-nginx-module/blob/master/src/ngx_http_lua_control.c)) pushes the Location header onto the *existing* `r->headers_out.headers` list and sets `r->headers_out.status`. It clears nothing else. `ngx.header["Set-Cookie"]` writes into that same `r->headers_out`, so the cookie is still in the list when the response is sent.

The suggested fix cannot work here: `ngx.redirect()` terminates the request during `rewrite`, so `header_filter` never runs for this response. That phase is the right call for `stale_session_cookie_cleanup_plugin` — which is why the codebase uses it there — because that plugin annotates a response the upstream produces, rather than generating one itself.

Worth noting the query-string alternative is also unworkable: the guard has to survive the round trip through Keycloak back to `/.apisix/redirect`, and the openid-connect plugin rebuilds `redirect_uri` from its own config, so any parameter added to the recovery redirect is gone by the second callback. A host-only `Path=/` cookie is the only carrier that persists across the flow — and `SameSite=Lax` is correct for it, since the callback arrives as a top-level navigation.

src/ol_infrastructure/components/services/apisix.py
```diff
@@ -103,6 +103,111 @@ def stale_session_cookie_cleanup_plugin(
...
+        return
+    end
+    local cookie = ngx.var.http_cookie
+    if cookie then
+        for pair in cookie:gmatch("[^;]+") do
+            if pair:match("^%s*([^=%s]+)=") == "{guard_cookie_name}" then
+                return
+            end
+        end
+    end
```
rhysyngsun: Can't we just use `ctx.var["cookie_{guard_cookie_name}"]` to look this up rather than running several regexes over the cookie header? I can't imagine this to be performant and possibly a good vector for a DOS attack.

src/ol_infrastructure/components/services/apisix.py
```diff
@@ -103,6 +103,111 @@ def stale_session_cookie_cleanup_plugin(
...
+    # which is how a caller turns the plugin into a no-op without detaching it
+    # from every route that references the shared config.  `or` would quietly
+    # turn that back into the default.
+    errors = (
+        ["temporarily_unavailable"]
+        if recoverable_errors is None
+        else recoverable_errors
+    )
+    recoverable_table = ", ".join(f'["{error}"] = true' for error in errors)
+    recovery_function = f"""return function(conf, ctx)
```
rhysyngsun: I'd prefer to have this live in an actual lua source file and then pass configuration in through the normal apisix configuration mechanisms rather than a python fstring with no syntax highlighting which makes it harder to review.

tests/ol_infrastructure/components/services/test_apisix_lua.py
```diff
@@ -0,0 +1,218 @@
...
+    fn({}, {})
+    local redirect = captured.redirect
+    return redirect and redirect.uri or nil,
+           redirect and redirect.code or nil,
+           captured.headers["Set-Cookie"]
+end
+"""
+
+
+class Harness:
```
rhysyngsun, line 56: I'm questioning what this testing is actually providing us, once these tests are green, they're always going to be green because we're testing in a vacuum and not actually against apisix itself, so we won't catch any possible regressions at best until they hit RC but possibly not until they hit production. 

APISIX has a testing framework for this: https://apisix.apache.org/docs/apisix/internal/testing-framework/

This plugin template appears to have the setup for running these kinds of tests: https://github.com/api7/apisix-plugin-template

rhysyngsun (changes requested):

src/ol_infrastructure/components/services/apisix.py
blarghmatey: Done in 1ae83432b — now a single `ctx.var["cookie_" .. guard]` read, so nginx does the parsing and the name match is exact rather than pattern-based.

One correction on the DoS framing, for the record rather than to argue the change: the URI check returns before any cookie work, so the scan only ever ran on requests to `*/.apisix/redirect` carrying a recoverable `error=` — a few hundred a day, not every request — and Lua patterns are not regex, so `[^;]+` cannot backtrack. The exposure was not there. `ctx.var` is still the better lookup and I would not go back to the loop.

src/ol_infrastructure/components/services/apisix.py
blarghmatey: Done in 1ae83432b. The Lua is now `src/ol_infrastructure/components/services/files/oidc_error_callback_recovery.lua`, shipped verbatim with nothing interpolated, and the tunables travel as an `oidc_error_recovery` block on the plugin config that the function reads off `conf`.

I checked that mechanism against the pinned 3.17.0 image rather than master before relying on it: `apisix/plugins/serverless/init.lua` invokes each function as `func(conf, ctx)`, and its schema declares only `phase`/`functions` with no `additionalProperties`, so extra keys pass `core.schema.check`. The rest of the path is opaque to unknown keys too — both CRDs mark `config` `x-kubernetes-preserve-unknown-fields`, the controller carries it as `apiextensionsv1.JSON`, and ADC as `Plugins map[string]any`. There is now a test that fails if any of those layers ever tightens.

Note I left `stale_session_cookie_cleanup_plugin` on the f-string pattern — same file, but out of scope for this PR. Happy to follow up if you want them consistent.

tests/ol_infrastructure/components/services/test_apisix_lua.py
blarghmatey: Agreed, and both layers are in 1ae83432b.

`tests/apisix_testnginx/` is the harness you pointed at — `t/APISIX.pm` on the same OpenResty build the gateway runs, 19 subtests, run via `tests/apisix_testnginx/run.sh`. Two things worth knowing if you touch it: it needs etcd (APISIX.pm boots the real init path, so without etcd every test fails the default `no [error] in error.log` check on connection-refused noise — I put etcd in the image rather than filtering those lines, since filtering would also hide errors we want), and it builds from the repo root so the Dockerfile copies the Lua out of `src/`. A copy sitting next to the `.t` would be free to drift and the suite would then pass against Lua we do not ship.

I also added `tests/apisix_integration/` — 11 tests against a real APISIX 3.17.0 container in standalone mode, fed the route config the actual plugin helper produces. That covers something test-nginx does not reach: whether APISIX accepts the config block at all. Both run as separate CI jobs.

On keeping the lupa tests: they are fast unit feedback on the generated Lua, not integration coverage, and I have stopped describing them as more than that. All three layers are mutation-tested rather than assumed — dropping the loop guard fails 1 integration test, simulating the config block being stripped fails 6, and removing the repeated-error table handling fails 2 test-nginx subtests. Happy to drop the lupa layer if you would rather not carry the dependency now that the other two exist.

blarghmatey: Agreed, and both layers are in 1ae83432b.

`tests/apisix_testnginx/` is the harness you pointed at — `t/APISIX.pm` on the same OpenResty build the gateway runs, 19 subtests, run via `tests/apisix_testnginx/run.sh`. Two things worth knowing if you touch it: it needs etcd (APISIX.pm boots the real init path, so without etcd every test fails the default `no [error] in error.log` check on connection-refused noise — I put etcd in the image rather than filtering those lines, since filtering would also hide errors we want), and it builds from the repo root so the Dockerfile copies the Lua out of `src/`. A copy sitting next to the `.t` would be free to drift and the suite would then pass against Lua we do not ship.

I also added `tests/apisix_integration/` — 11 tests against a real APISIX 3.17.0 container in standalone mode, fed the route config the actual plugin helper produces. That covers something test-nginx does not reach: whether APISIX accepts the config block at all. Both run as separate CI jobs.

On keeping the lupa tests: they are fast unit feedback on the generated Lua, not integration coverage, and I have stopped describing them as more than that. All three layers are mutation-tested rather than assumed — dropping the loop guard fails 1 integration test, simulating the config block being stripped fails 6, and removing the repeated-error table handling fails 2 test-nginx subtests. Happy to drop the lupa layer if you would rather not carry the dependency now that the other two exist.

.github/workflows/pytest.yml
```diff
@@ -29,3 +29,45 @@
...
+        python-version-file: pyproject.toml
+
+    - name: Install dependencies
+      run: uv sync --frozen
+
+    # Excluded from the default run by `-m 'not integration'` in pyproject.
+    - name: Run APISIX integration tests
+      run: uv run pytest tests/apisix_integration -m integration
+
+  apisix-testnginx:
```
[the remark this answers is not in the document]
blarghmatey: Fixed in 85be91910 — `permissions: contents: read` set at the workflow level rather than on these two jobs, so the existing `test` job is covered too and any job added later inherits the restriction instead of silently getting the default token. Nothing in this workflow writes back: no comments, no publishing, no artifacts.

.github/workflows/pytest.yml
```diff
@@ -29,3 +29,45 @@
...
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
+      with:
+        persist-credentials: false
+
+    # Builds an image from the pinned APISIX version, adds Test::Nginx and
+    # etcd, and runs APISIX's own harness against src/.../files/*.lua.
+    - name: Run test-nginx suite
+      run: tests/apisix_testnginx/run.sh
```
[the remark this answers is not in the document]
blarghmatey: Same fix as the sibling finding — 85be91910 sets `permissions: contents: read` for the whole workflow.

blarghmatey:
Picked up the three suppressed findings from the latest Copilot review — they had no inline threads, so replying here.

**1. `%.apisix/redirect$` matched near-miss paths — correct, fixed in 09b730a9e.** It also matched application paths ending in the same characters, e.g. `/login/foo.apisix/redirect`, and since this is attached host-wide such a request carrying `error=temporarily_unavailable` would have been redirected. Now anchored on the leading slash. The `gsub` still strips only `.apisix/redirect` so the target keeps its trailing slash (`/login/`, not `/login`). Regression case added to all three test layers; reverting the anchor fails one test in each.

**2. `APISIX_IT_IMAGE` documented but not read — correct, fixed in 09b730a9e.** The docstring advertised an override the constant never honoured, so setting it silently kept testing 3.17.0. That is exactly the false confidence you would not want when rehearsing a version bump. It reads the environment variable now.

**3. The alert-silence claim was wrong — thank you, this one mattered.** I checked the arithmetic and Copilot is right. `APISIXOIDCCallbackFailureRateChronic` fires on a ratio (>5% over 6h) and its own annotation says it stays up until *both* causes are fixed. This PR fixes one:

| host | callbacks/24h | 500 rate now | after this PR | chronic >5%? |
|---|---|---|---|---|
| api.learn.mit.edu | 3,297 | 17.4% | 7.5% | **still fires** |
| mitxonline.mit.edu | 2,633 | 25.4% | 14.6% | **still fires** |

The replay 500s alone hold both hosts above the threshold — removing 49% of the absolute count is not the same as clearing a 5% ratio. I had stated the opposite confidently in the PR description, the tracking task, and a team memory; all three are corrected. The deployment check is now the `temporarily_unavailable`-specific 500 count going to ~0 plus the overall rates landing near the "after" column, not alert silence.

No rule change included: the rule behaves as designed, it was the claim about it that was wrong. Adding a `temporarily_unavailable`-specific expression so this cause is independently trackable is a reasonable follow-up if wanted.

blarghmatey:
## Verified on CI

Deployed this branch's shared-plugin config to the CI stack and ran the probes. The hop no test layer reaches — CRD → ingress-controller → ADC, against real Keycloak — works: the `oidc_error_recovery` block survives to APISIX, so the config-drop failure mode the unit/integration suites can't see is ruled out.

Deploy was scoped with `--target` to the two `mitlearn-ol-shared-plugins` resources, so it left #4759's browser resources and the pre-existing Fastly drift alone. `pulumi preview` on the branch vs. on `main` differ by exactly those two resources — this PR introduces no other change to CI.

| # | probe | result |
|---|---|---|
| 1 | `error=temporarily_unavailable` | **302**, `location: /login/`, `set-cookie: apisix_oidc_recovery=1; Path=/; Max-Age=60; Secure; HttpOnly; SameSite=Lax` |
| 2 | `error=access_denied` | **500** — correctly not recovered |
| 3 | retry carrying the guard cookie | **500** — guard engages, no redirect loop |
| 4 | `/login/foo.apisix/redirect` | falls through to a normal Keycloak auth redirect, not intercepted — 09b730a9e's slash anchoring holds |
| 5 | `/login` | authorization redirect well-formed, `redirect_uri` https; Keycloak serves a real login form (200) |
| 6 | `/api/v1/courses/` | 200 — unauthenticated API paths unaffected by the rewrite-phase plugin |

Probe 1 matches the expected `Set-Cookie` byte for byte, which also settles Seer's HIGH prediction that `ngx.redirect()` would drop a header set via `ngx.header[...]` — it does not, on a real gateway.

Not covered: a full credentialed login. Probe 5 gets as far as Keycloak rendering the form; completing the code exchange needs credentials I don't have. Worth one manual browser login through https://ci.learn.mit.edu before production.

## Merge-order interaction with #4759

#4759 adds a second shared plugin config, `mitlearn-ol-browser-shared-plugins`, and points the `browser-*` route rules at it. It carries its own copy of `stale_session_cookie_cleanup_plugin` but would not carry this PR's recovery plugin, so the two configs drift apart on the `/login/*` path. On CI right now, with both deployed, that is directly observable:

```
Origin: https://ci.learn.mit.edu  →  500   (browser-reqauth → browser-shared-plugins, no recovery plugin)
no Origin header                  →  302   (reqauth → ol-shared-plugins, recovered)
```

In practice this probably does not hit the real callback: the callback is a cross-origin **GET** top-level navigation from Keycloak, and browsers don't attach `Origin` to those, so it should keep landing on the non-`Origin` rules. I'd rather not rest a login path on that reasoning without a real browser confirming it.

Either way it's a latent trap — whichever of these two PRs merges second should attach `oidc_error_callback_recovery_plugin()` to the browser config as well, so the two configs can't silently disagree about whether an expired auth session is recoverable.

## Unrelated: CI stack has pending operations

The CI stack state carries 4 pending `creating` ops on `mitlearn-ci-pre-deploy-job`, stamped `2026-08-19T19:30:55Z` — from the #4759 deploys, not from this run (my `--target`ed update never touched that resource, and no such Jobs exist in the cluster now). Clearing them needs an interactive `pulumi refresh` on the CI stack.

blarghmatey:
Rebased onto main and resolved the pytest.yml conflict (this branch's apisix-integration/apisix-testnginx jobs vs. main's ci-gate + workflow-scoped permissions in #5567/#5571): kept both, added the two jobs to ci-gate's needs list, applied the contents:read restriction. No review threads were open — the only blocker was the merge-order conflict causing pre-commit.ci's mergeable-check error.

blarghmatey:
Checked whether this is still needed given upstream shipped a similar fix: [apache/apisix#13825](https://github.com/apache/apisix/pull/13825) adds native `temporarily_unavailable`/state-mismatch recovery to the `openid-connect` plugin itself, released in APISIX 3.18.0. `main` is already on that version (chart 2.17.0, #5641).

Confirmed the fix is actually running in production, not just merged: `kube_pod_container_info` shows the `operations` namespace APISIX pods on `apisix:3.18.0-ubuntu`.

It doesn't fix the problem in practice. Live LogQL against the last 24h:

```logql
sum by (host, status) (count_over_time(
  {namespace="operations", container="apisix"}
  |= ".apisix/redirect" |= "error=temporarily_unavailable" | logfmt | __error__="" [24h]))
```

- `api.learn.mit.edu`: 485 still 500, 63 recovered as 302 (88% still failing)
- `mitxonline.mit.edu`: 58 still 500, 4 recovered (94% still failing)

Sample lines from the last few minutes show real OIDC session cookies present (`cookie_count` 13-21, `oidc_session_bytes≈3.3KB`) still hitting 500 on `error=temporarily_unavailable`.

Root cause of the gap: upstream's recovery branch in `openid-connect.lua` only fires when `session and core.string.has_prefix(err, UNHANDLED_REDIRECT_URI_ERR)`, and further requires `authorization_states[uri_args.state]` (or the legacy single `state` slot) to still be present in APISIX's own pending-auth session. If that entry has already been pruned or evicted — plausible when APISIX's own session TTL is shorter than how long a user leaves the Keycloak tab idle, or under concurrent tabs/flows — it falls straight through to a plain 500 instead of restarting.

This PR's `oidc_error_callback_recovery.lua` doesn't have that dependency: it acts on the `error=` query param and a standalone guard cookie, ahead of `openid-connect` and its session/state matching, so it still recovers what the native fix misses. Keeping this open, and given the evidence above it looks worth merging soon rather than assuming 3.18.0 already covers it.

PR #5456, opened 2026-08-17
fix(dagster): size run concurrency per environment and stop the pool deadlocking forever

blarghmatey:

N/A — falls out of the Dagster/PgBouncer observability work in `docs/plans/dagster-pgbouncer-observability.md`. Companion to #5454 (the alert that surfaces this condition) and follows #5426 (the connection cap).

QA's Dagster had been frozen for days. Found while calibrating the alert thresholds in #5454, not by anything that was watching.

State when found, on both PgBouncer replicas: every one of the 764 backends pinned (`sv_active` 764, `sv_idle` 0), 132 clients queued, the oldest waiting 23,797s and climbing 1:1 with wall clock — the same client, never served, having reached 210,619s (2.4 days) before a pod restart reset the counter. All 382 active clients per replica had last issued a request at the same instant, 08:50:46, and never made another. The daemon's own log stops at 08:50:46 too, part-way through cancelling backfill `lupslfla`.

Two independent settings made that state both reachable and permanent.

**`max_concurrent_runs` was 100 for every environment.** Dagster's storage uses `NullPool`, so a run worker opens a connection per query and holds one per concurrent step. During the incident 32 run workers held ~700 client connections between them. That is comfortable against production's aggregate cap of 4248 and 2.9× QA's 764. It now comes from stack config — production keeps 100, QA takes 20.

**`query_wait_timeout` was 0.** It was set that way for a good reason: the 120s default disconnected clients during RDS checkpoint I/O, which clears in 1–2 minutes. But disabling it entirely traded a transient failure for a permanent one. A timeout is the only thing that breaks a pool deadlock — when every server connection is held by a client that is itself blocked waiting for a server connection, nothing is released until something gives up, and with 0 nothing ever does. Now 600s: 5× the margin over the pressure it was meant to absorb, while restoring the property that a wedged pool eventually unwedges itself.

`max_concurrent_runs` becomes a placeholder in `dagster_instance.yaml` substituted in `__main__.py`. The existing `checksum/ol-dagster-instance` annotation hashes the substituted text, so a change to the value rolls the daemon and webserver automatically.

### How can this be tested?

Already applied to QA by hand and observed. Recovery sequence was: scale `dagster-daemon` to 0 (so nothing dequeues when connections free), delete the 32 stuck run Jobs, then restore the daemon at the new concurrency.

- On deleting the Jobs, the pool drained immediately — `cl_waiting` 764 → 0, `maxwait` → 0, ~140 idle servers per replica. The webserver that had been crash-looping 86 times went 1/1 without intervention.
- Queue drained on restart: `QUEUED` 683 → 347 within minutes, `CANCELING`/`STARTING`/`NOT_STARTED` all → 0.
- `STARTED` settled at exactly **20**, confirming the new limit binds.
- Pool stayed healthy throughout: peak 221 active servers on the busier replica against its 382 per-pod cap, `cl_waiting` 0 and `maxwait` 0 at every check.

The in-cluster ConfigMap was patched directly to `max_concurrent_runs: 20` to get QA running before this merges, so QA is currently **drifted from Pulumi in the same direction as this PR** — deploying the dagster stack reconciles it. Nothing was changed by hand in production.

### Additional Context

**The ~22-connections-per-worker figure is a deadlock artifact, not steady state.** With the deadlock cleared, 47 run workers were using 11 server connections between them. QA's 20 is therefore deliberately conservative, and the comment in `Pulumi.QA.yaml` says so — it should be revisited once the exporter from #5426 has enough history to show what a run worker really holds under normal load.

**Separate finding worth a look, not addressed here:** load across the two PgBouncer replicas is very uneven — 221 active servers on one against 4 on the other, both capped at 382. `max_db_connections` is derived by dividing the budget by replica count, which assumes even distribution; in practice the effective ceiling is whichever replica saturates first, not the aggregate. Worth folding into the pool re-tuning work.

https://claude.ai/code/session_013x721suvDpTPEbYXn9jZQD

### DIFF
--- a/src/ol_infrastructure/applications/dagster/Pulumi.QA.yaml
+++ b/src/ol_infrastructure/applications/dagster/Pulumi.QA.yaml
@@ -23,5 +23,13 @@ config:
   dagster:edx_pipeline_xpro_edx_bucket_name: dagster-data-qa
   dagster:edx_pipeline_xpro_edx_course_bucket_name: xpro-qa-edxapp-courses
   dagster:edx_pipeline_xpro_edx_xpro_purpose: mitx-etl-xpro-qa
+  # QA's aggregate PgBouncer connection cap is 764 (db.m7g.large, 2 replicas)
+  # against production's 4248. At the global default of 100 concurrent runs,
+  # QA deadlocked its own pool on 2026-08-17 and stayed wedged for days. 20
+  # runs at the ~22 connections apiece measured during that incident is ~440,
+  # leaving room for the daemon, webserver and ten code location servers.
+  # Revisit once the pgbouncer exporter has enough history to show what a run
+  # worker really holds outside a deadlock.
+  dagster:max_concurrent_runs: 20
   vault:address: https://vault-qa.odl.mit.edu
   vault_server:env_namespace: operations.qa
--- a/src/ol_infrastructure/applications/dagster/__main__.py
+++ b/src/ol_infrastructure/applications/dagster/__main__.py
@@ -701,16 +701,33 @@ pgbouncer_config = kubernetes.core.v1.ConfigMap(
                     # terminates them, preventing zombie connections from being
                     # assigned to clients.
                     "server_idle_timeout = 120",
-                    # Disable the default 120s query_wait_timeout. During heavy RDS
-                    # checkpoint I/O, queries slow from milliseconds to seconds, which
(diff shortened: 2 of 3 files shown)


Ardiea (approved):

blarghmatey:
Consolidating into #5454, which already carries both commits. Same changes, one review.

PR #5488, opened 2026-08-18
fix(k8s): split webapp liveness to TCP, keep readiness on HTTP

blarghmatey:

Part of the Granian configuration overhaul epic. Closes the deferred probe item from `docs/plans/granian-configuration-overhaul.md` (witan `tk-split-webapp-liveness-probe-to-tcp-keep-readines-4f58d9`).

## Description (What does this PR do?)

Changes the webapp **liveness** probe from an HTTP GET to a TCP connect, on the component default and on every Granian app that overrides `probe_configs` wholesale. **Readiness and startup stay on HTTP** against the same django-health-check / `/heartbeat` endpoints they use today.

An HTTP liveness request has to be served by the same bounded pool of Granian blocking threads that is serving user traffic. Now that this overhaul is tightening backpressure (64 → 16), a pod that is merely *saturated* starts failing liveness, gets restarted, and takes its capacity out of a service that was already overloaded. Granian's maintainer reports exactly that pod-churn spiral in [granian discussion #663](https://github.com/emmett-framework/granian/discussions/663).

A TCP connect is answered by the listener rather than by a worker thread, so it distinguishes "the process is gone" from "the process is busy".

**The trade-off is deliberate and worth stating plainly:** liveness no longer restarts a pod whose Granian workers are wedged but whose listener still accepts. Readiness *does* still fail in that case, so the pod is pulled from the Service endpoints and stops receiving traffic — it just is not killed automatically. We give up auto-recovery from a true wedge in exchange for never restarting a pod for the crime of being busy.

Files touched:

| file | shape |
| --- | --- |
| `components/services/k8s.py` | component default (`default_probe_configs`) — covers ocw_studio, xpro, mit_learn, learn_ai, mitxonline |
| `applications/micromasters/__main__.py` | override, had a Host header |
| `applications/edxapp/k8s_resources.py` | override x2 (LMS + CMS), had a Host header |
| `applications/odl_video_service/__main__.py` | override, no header |
| `applications/ol_analytics_api/__main__.py` | override, no header |

The Host header is dropped from the liveness entries only — it existed to satisfy Django's `ALLOWED_HOSTS`, which a TCP connect never reaches. Readiness and startup keep theirs.

`edx_notes` and `xqueue` also override `probe_configs`, but they are not Granian workloads, so they are deliberately left alone rather than swept along.

## How can this be tested?

`pulumi preview` against CI for all four code shapes. Every stack shows a single Deployment update touching only `livenessProbe`, no replacements:

- **ocw_studio.CI** (component default): `- httpGet {path: /health/liveness/, port: 8073}` → `+ tcpSocket {port: 8073}`, 1 Deployment
- **micromasters.CI** (override + header): `- httpGet {httpHeaders: [...backend.odl.mit.edu]}` → `+ tcpSocket {port: 8071}`, 1 Deployment
- **edxapp mitx.CI** (override + header, x2): both LMS and CMS Deployments → `+ tcpSocket {port: 8000}`
- **odl_video_service.CI** (override, no header): → `+ tcpSocket {port: 8073}`, 1 Deployment

Test suite: 816 passed. Updated the two existing `default_probe_configs` assertions and the `_probe_ports` helper (which assumed every probe carries `http_get`), and added `test_liveness_is_tcp_while_readiness_stays_http` as the regression guard.

`pre-commit run --all-files`: ruff, ruff-format, and mypy all pass. (hadolint reports pre-existing findings on Dockerfiles this PR does not touch.)

## Additional Context

Sequenced deliberately: this was held back until after stage 2 so the probe change and the concurrency changes land in separate deploys and stay separable if either one moves a latency or restart metric.

https://claude.ai/code/session_0166K1vrjtPKzK49L45UMyDe

### DIFF
--- a/src/ol_infrastructure/applications/edxapp/k8s_resources.py
+++ b/src/ol_infrastructure/applications/edxapp/k8s_resources.py
@@ -794,17 +794,14 @@ def create_k8s_resources(  # noqa: C901
                 limit_workers_max_rss=True,
             ),
             probe_configs={
+                # TCP rather than HTTP so a saturated Granian worker pool
+                # cannot fail liveness and trigger a restart that removes
+                # capacity from an already overloaded service. Readiness stays
+                # on HTTP, and keeps the Host header because it still reaches
+                # Django's ALLOWED_HOSTS check. See default_probe_configs in
+                # components/services/k8s.py.
                 "liveness_probe": kubernetes.core.v1.ProbeArgs(
-                    http_get=kubernetes.core.v1.HTTPGetActionArgs(
-                        path="/heartbeat",
-                        port=8000,
-                        http_headers=[
-                            kubernetes.core.v1.HTTPHeaderArgs(
-                                name="Host",
-                                value=edxapp_config.require_object("domains")["lms"],
-                            ),
-                        ],
-                    ),
+                    tcp_socket=kubernetes.core.v1.TCPSocketActionArgs(port=8000),
                     initial_delay_seconds=30,
                     period_seconds=30,
                     failure_threshold=3,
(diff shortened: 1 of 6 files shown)


src/ol_infrastructure/applications/edxapp/k8s_resources.py
```diff
@@ -794,17 +794,14 @@ def _make_vector_sidecar(service: str) -> kubernetes.core.v1.ContainerArgs:
...
-                        path="/heartbeat",
-                        port=8000,
-                        http_headers=[
-                            kubernetes.core.v1.HTTPHeaderArgs(
-                                name="Host",
-                                value=edxapp_config.require_object("domains")["lms"],
-                            ),
-                        ],
-                    ),
+                    tcp_socket=kubernetes.core.v1.TCPSocketActionArgs(port=8000),
```
Ardiea, line 804: This is smart. Saves us from requesting an application change for a second /healthcheck endpoint or something similar.

Ardiea (approved):

src/ol_infrastructure/applications/edxapp/k8s_resources.py
blarghmatey: Worth noting the cost, since it is not free: a TCP liveness probe no longer restarts a pod whose Granian workers are wedged but whose listener still accepts. Readiness still fails there, so the pod leaves the Service endpoints and stops taking traffic — it just is not killed. That trade is deliberate, and it is recorded in the default_probe_configs docstring.

PR #5511, opened 2026-08-18
feat(local-dev): MIT Learn account actions via a dedicated Keycloak client

rhysyngsun:
Part of https://github.com/mitodl/hq/issues/10815

Makes MIT Learn's change-email / change-password account actions work in local-dev.

- Adds `ol-mitlearn-account-client`, a confidential Keycloak client dedicated to Keycloak's application-initiated actions (AIA). It is separate from `ol-mitlearn-client`, which APISIX uses for the session: Django sends the user to Keycloak with `kc_action`, Keycloak returns them to `/account/action/complete`, and Django exchanges the code server-side to read the updated email, since the gateway's cached userinfo still holds the old address.
- Its `valid_redirect_uris` ends in `*` so Keycloak's exact-match check still passes once Django appends `?next=...&kc_action=...`, mirroring the production mitxonline entry in `substructure/keycloak/Pulumi.Production.yaml`.
- Enables the `UPDATE_EMAIL` required action in the `olapps` realm. It was registered but left `enabled=False`, and an AIA for a disabled required action redirects back with `kc_action_status=error` — which is what the change-email flow was hitting.

The comment being replaced claimed `UPDATE_EMAIL` was removed in Keycloak 26. It was not. It is gated behind the `update-email` preview feature, which the `mitodl/keycloak` image bakes in at build time (`--features=update-email` in `ol-keycloak`'s `Dockerfile.hosted`). On the pinned image, `kc.sh show-config` reports:

```
kc.features =  update-email (Persisted)
```

So a provider "unknown alias" validation error on that alias would mean the image lost the feature flag, not that the action is obsolete — noted in the new comment so the next person does not re-derive it.

### How can this be tested?

Against local-dev (`tilt up`), signed in to Learn as a non-SSO user:

1. Go to `https://learn.mit.dev/dashboard/settings` and start the change-email action.
2. Keycloak should present its update-email form and return you to `https://api.learn.mit.dev/account/action/complete?...&kc_action_status=success`. Before this change the same URL came back with `kc_action_status=error`.
3. Confirm the new address is reflected in Learn.

Confirmed working end-to-end on local-dev.

To verify the realm state directly (local-dev admin credentials are the committed `admin`/`admin` defaults used by `local-dev/scripts/kc-fix-flow-bindings.sh`):

```bash
TOKEN=$(curl -s --cacert local-dev/certs/rootCA.pem \
  -X POST "https://sso.ol.mit.dev/realms/master/protocol/openid-connect/token" \
  -d "client_id=admin-cli&username=admin&password=admin&grant_type=password" \
  | jq -r .access_token)

curl -s --cacert local-dev/certs/rootCA.pem -H "Authorization: Bearer $TOKEN" \
  "https://sso.ol.mit.dev/admin/realms/olapps/authentication/required-actions/UPDATE_EMAIL"
```

`UPDATE_PASSWORD` was already enabled, so the change-password action is a useful control that the client and its redirect URI are sound independently of the required-action change.

### Additional Context

Note for reviewers on how the required action reaches the realm: `keycloak.RequiredAction` for `UPDATE_EMAIL` updates the built-in action that Keycloak already registered rather than creating a new one, which is the same thing the three existing entries (`CONFIGURE_TOTP`, `VERIFY_EMAIL`, `UPDATE_PASSWORD`) do. No Pulumi import is needed.

Because the action was flipped live on one local-dev cluster while debugging, that cluster was already in the fixed state before this commit existed. The declaration here is what makes it survive a realm rebuild — anyone else needs the `pulumi up` to pick it up.

A `keycloak_log_level` knob for local-dev is deliberately not in this PR and will follow separately. It is worth having: the `olapps` realm has `events_enabled` with the `jboss-logging` listener, but Keycloak emits those events at DEBUG, so this failure produced nothing in the pod log at the default level and had to be diagnosed from the realm's required-action state instead.

### DIFF
--- a/local-dev/infra/modules/keycloak.py
+++ b/local-dev/infra/modules/keycloak.py
@@ -145,7 +145,14 @@ def create_olapps_dev_realm(  # noqa: PLR0913
     for alias, default in [
         ("CONFIGURE_TOTP", False),
         ("VERIFY_EMAIL", verify_email),
-        # UPDATE_EMAIL was removed in Keycloak 26 — omit to avoid validation error.
+        # UPDATE_EMAIL is gated behind Keycloak's `update-email` preview feature,
+        # which the mitodl/keycloak image bakes in at build time (see
+        # Dockerfile.hosted in ol-keycloak). Declaring it here keeps the action
+        # enabled for MIT Learn's application-initiated update-email flow -- an
+        # AIA request for an action the realm has disabled comes back as
+        # kc_action_status=error. If the provider rejects this alias as unknown,
+        # the image lost the feature flag rather than the action being obsolete.
+        ("UPDATE_EMAIL", False),
         ("UPDATE_PASSWORD", False),
     ]:
         keycloak.RequiredAction(
--- a/local-dev/infra/modules/keycloak.py
+++ b/local-dev/infra/modules/keycloak.py
@@ -30,6 +30,7 @@ def create_olapps_dev_realm(  # noqa: PLR0913
     keycloak_url: str,
     k8s_provider: k8s.Provider,
     mitlearn_client_secret: Output,
+    mitlearn_account_client_secret: Output,
     learn_ai_client_secret: Output,
     mitxonline_client_secret: Output,
     unified_ecommerce_client_secret: Output,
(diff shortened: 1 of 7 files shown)


local-dev/infra/modules/keycloak.py
```diff
@@ -494,6 +502,50 @@ def _make_oidc_secret(
         mitlearn_client_secret,
     )
 
+    # --- MIT Learn account actions ---
+    # Separate from ol-mitlearn-client (which APISIX uses for the session)
+    # because its only job is Keycloak's application-initiated actions — update
+    # email / update password. Django sends the user here with `kc_action`,
+    # Keycloak returns them to the callback below, and Django exchanges the code
+    # server-side to read the updated email, since the gateway's cached userinfo
```
ChristopherChudzicki, line 510: > and Django exchanges the code server-side to read the updated email
@rhysyngsun Am I misunderstanding this PR? Didn't we remove the token exchange?

In that case, can't Learn just stay on `ol-mitlearn-client` and `local-dev/apps/mit-learn/configmaps/app-env.yaml` can point `KEYCLOAK_CLIENT_ID: ol-mitlearn-client`? Isn't that what prod does:

https://github.com/mitodl/ol-infrastructure/blob/6edd20cca9efe75f1a04f8831af4e3a203b8c8d7/src/ol_infrastructure/substructure/keycloak/olapps.py#L446-L452

?

ChristopherChudzicki (changes requested):
As mentioned in Slack, I didn't have SCIM set up, so I didn't get the email updated in Learn. Keycloak did get the update, though.

It seems like some of this PR is built for the time when https://github.com/mitodl/mit-learn/pull/3726 included token exchange. But that got removed, right? So I think we can remove `mitlearn_account_client`

PR #5558, opened 2026-08-21
fix(starrocks): harden the FE MySQL NLB security group for the new MIT Learn consumer

blarghmatey:
Witan task: tk-sg-harden-starrocks-fe-mysql-nlb-for-vault-mit-l-fb8043

StarRocks' internal NLB (FE MySQL port 9030) had no explicit security group -- the AWS Load Balancer Controller auto-manages a default one with no source restriction, gated only by internal scheme + VPC-peering routability. MIT Learn is now going to read from StarRocks directly (replacing a broken Hightouch-based sync and drawing down Trino usage), so this closes the SG gap before adding that new consumer.

- `applications/starrocks/__main__.py`: adds `fe_mysql_nlb_security_group`, a `SecurityGroup` in the data VPC that allows only TCP/9030 from three known consumers, then attaches it to the NLB `Service` via the `service.beta.kubernetes.io/aws-load-balancer-security-groups` annotation. This makes the new SG the sole authority on the NLB's ingress (the LBC stops managing a default one), so all real consumers had to be identified up front:
  - Vault's server SG (`operations.{env}` stack) -- dynamic DB credential management, existing consumer.
  - MIT Learn's app pod SG (`{env}` stack) -- the new consumer.
  - This environment's Concourse worker SG (`{env}` stack) -- runs the `substructure/starrocks` SQL setup `local.Command` resources, which need mysql-client access to the same NLB endpoint (see that stack's module docstring). Confirmed via `src/ol_concourse/pipelines/infrastructure/simple_pulumi/meta.py`, which routes `starrocks-substructure[-qa]` to the env-matched Concourse instance -- so this is an env-matched dependency, same as Vault and MIT Learn.
- `applications/mit_learn/__main__.py`: exports `app_security_group_id` (the existing `mitlearn_app_security_group`, already bound to the app pods) so starrocks can consume it.
- `applications/concourse/__main__.py`: exports `concourse_worker_security_group_id` (the existing `concourse_worker_security_group`) for the same reason -- it wasn't exported before this.

### How can this be tested?
Ran `pulumi preview` for both `QA` and `Production` across all three touched projects (S3 backend `s3://mitol-pulumi-state`):

- `applications/mit_learn` (stacks `QA`, `Production`): additive-only diff, `+ app_security_group_id: "sg-..."` in outputs, no resource changes from this PR. (Preview required `MIT_LEARN_DOCKER_TAG=latest` to get past an unrelated missing-env-var error; the resulting container-image-tag diffs are an artifact of that placeholder, not this change.)
- `applications/concourse` (stacks `QA`, `Production`): additive-only diff, `+ concourse_worker_security_group_id: "sg-..."` in outputs, no resource changes from this PR.
- `applications/starrocks` (stacks `lakehouse.QA`, `lakehouse.Production`): preview currently fails with `KeyError: 'app_security_group_id'` because the *deployed* `mit_learn` and `concourse` stacks don't yet export the new keys this code reads -- expected, since stack references resolve against already-applied state. **Deploy order matters: `mit_learn` and `concourse` must ship (both are additive, output-only, zero resource changes) before `starrocks`.** Once deployed, `starrocks` preview is expected to show one new `SecurityGroup` (with the three ingress source SGs above) and an additive annotation update on the existing `fe_mysql_nlb_service` -- no replacement of the NLB `Service` itself, so no StarRocks-access outage.
- `pre-commit run --files <changed files>` and `mypy` both pass clean on all three files.

### Additional Context
- The Concourse-coverage question was the main open risk in this task: Vault's SG and Concourse's worker SG are *not* the same thing (Vault runs on its own EC2 ASG; Concourse workers are a separate EC2 fleet, both in the operations VPC), so Concourse needed its own explicit ingress entry -- confirmed via `src/ol_infrastructure/applications/concourse/__main__.py` and the pipeline routing in `src/ol_concourse/pipelines/infrastructure/simple_pulumi/meta.py`.
- Not merging/deploying this PR myself per the driving task's instructions -- flagging the two-phase rollout (mit_learn + concourse first, then starrocks) for whoever applies it.

### DIFF
--- a/src/ol_infrastructure/applications/concourse/__main__.py
+++ b/src/ol_infrastructure/applications/concourse/__main__.py
@@ -408,6 +408,11 @@ concourse_worker_security_group = ec2.SecurityGroup(
     egress=default_egress_args,
     vpc_id=ops_vpc_id,
 )
+# Exported so other stacks' local.Command resources (e.g.
+# substructure/starrocks's SQL setup, which runs on this Concourse instance's
+# workers per src/ol_concourse/pipelines/infrastructure/simple_pulumi/meta.py)
+# can admit these workers into security groups gating VPC-internal endpoints.
+export("concourse_worker_security_group_id", concourse_worker_security_group.id)
 
 # Create web node security group
 concourse_web_security_group = ec2.SecurityGroup(
--- a/src/ol_infrastructure/applications/mit_learn/__main__.py
+++ b/src/ol_infrastructure/applications/mit_learn/__main__.py
@@ -1895,5 +1895,6 @@ export(
         "vault_iam_role": Output.all(
             mitlearn_vault_iam_role.backend, mitlearn_vault_iam_role.name
         ).apply(lambda role: f"{role[0]}/roles/{role[1]}"),
+        "app_security_group_id": mitlearn_app_security_group.id,
     },
 )
--- a/src/ol_infrastructure/applications/starrocks/__main__.py
+++ b/src/ol_infrastructure/applications/starrocks/__main__.py
@@ -7,7 +7,7 @@ from typing import Any, cast
 import pulumi_kubernetes as kubernetes
 import pulumi_vault
 from pulumi import Config, InvokeOptions, Output, ResourceOptions, export
-from pulumi_aws import iam
+from pulumi_aws import ec2, iam
 
 from bridge.lib.versions import STARROCKS_CHART_VERSION, STARROCKS_VERSION
 from ol_infrastructure.components.applications.eks import (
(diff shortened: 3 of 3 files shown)


blarghmatey:
Duplicate of #5556, which a concurrent session opened for the same witan task (tk-sg-harden-starrocks-fe-mysql-nlb-for-vault-mit-l-fb8043) minutes earlier with the same approach (same 3 files, same 3 admitted security groups, same export additions). Closing this one in favor of #5556 to avoid two competing PRs for the same change.

PR #5560, opened 2026-08-21
feat(edxapp): wrap LMS/CMS/celery with opentelemetry-instrument

blarghmatey:
N/A -- addresses a copilot review comment on mitodl/lehrer#177.

Stacked on #5557 (dead uwsgi cleanup) since both touch the same LMS/CMS config dicts.

lehrer#177 installs OTel packages in the edxapp image, but installing them doesn't activate anything: LMS/CMS run under `granian` and workers under a plain `celery` invocation, and `edx_django_utils.monitoring.OpenTelemetryBackend` only decorates whatever span is already active -- it never installs a TracerProvider itself. Copilot flagged exactly this on lehrer#177. This PR wraps every process with the `opentelemetry-instrument` auto-instrumentation agent.

- Adds `command_prefix: list[str]` to `OLApplicationK8sConfig` in the shared `k8s.py` component -- empty by default, so every other `granian_config` consumer (mit_learn, mitxonline, etc.) is unaffected. Set to `["opentelemetry-instrument"]` for edxapp's LMS/CMS only.
- Prefixes the command on all 5 hand-rolled celery-family deployments (lms-celery, lms-high-mem-celery, lms-beat, cms-celery, process-scheduled-emails).
- Moves `OTEL_SERVICE_NAME` out of the shared interpolated-config env dict (`env_name + "-edxapp"` for every workload, which collapsed LMS/CMS/all five celery deployments into one Tempo service) into each workload's own env, distinct per deployment.

Two things I confirmed by actually running `opentelemetry-instrument` against `opentelemetry-distro` + `opentelemetry-exporter-otlp-proto-http` locally (the exact package set lehrer#177 installs), not assumed from docs:

1. `OTEL_TRACES_EXPORTER` defaults to `"otlp"`, which needs an entry point named exactly `"otlp"` -- only `"otlp_proto_http"` is registered here (grpc isn't installed), so it must be set explicitly or the exporter fails to resolve.
2. More importantly: `OTEL_METRICS_EXPORTER`/`OTEL_LOGS_EXPORTER` also default to `"otlp"` (grpc), and the SDK resolves traces/metrics/logs exporters together in one call that raises on the *first* one it can't find -- so an unset metrics/logs exporter aborts trace configuration too. Confirmed via `RuntimeError: Requested component 'otlp_proto_grpc' not found in entry point 'opentelemetry_metrics_exporter'`. Worse: this failure is completely silent in the actual wrapped-process path -- `opentelemetry-instrument`'s sitecustomize swallows the exception by default -- so without this fix the wrap would produce zero traces with no error anywhere. Set both to `"none"` since neither signal is part of this rollout.

### How can this be tested?
Ran `pulumi preview --diff` against `mitxonline.CI`. All 5 celery-family containers and both LMS/CMS granian containers show `command` prefixed with `opentelemetry-instrument`; `OTEL_SERVICE_NAME` is distinct per workload; no unexpected resource changes beyond the normal rolling updates.

### Additional Context
**Ordering hazard: do not merge/deploy ahead of mitodl/lehrer#177 shipping and a new edxapp image being built with it.** `opentelemetry-instrument` must actually be present in the image or granian/celery fail to start (missing executable).

Per the project's staged-rollout guidance, consider deploying to QA and a low-volume workload first rather than flipping LMS/CMS in production immediately -- LMS/CMS are the highest-volume workloads in the cluster, and an under-sized tail sampler previously caused a ~98% trace-drop incident earlier this project.


Ardiea (approved):

blarghmatey:

blarghmatey:
@copilot address the pr feedback, move the otel settings into the container environment

PR #5612, opened 2026-08-27
feat(xpro): wire CyberSource REST credentials for export compliance

Anas12091101:

Refs https://github.com/mitodl/hq/issues/13003
Pairs with mitodl/mitxpro#4067

xPRO is moving its export-compliance check off CyberSource's retired SOAP UsernameToken auth onto the REST API, which authenticates with HTTP Signature. This adds the four settings that change needs:

- `CYBERSOURCE_REST_KEY_ID` / `CYBERSOURCE_REST_SECRET` — templated from the existing `secret-xpro/cybersource` Vault path
- `CYBERSOURCE_REST_MERCHANT_ID` — the same `mit_odl_xpro` merchant as the SOAP integration
- `CYBERSOURCE_REST_ENVIRONMENT` — `apitest.cybersource.com` on CI/QA, `api.cybersource.com` on production

**Deliberately additive.** The SOAP settings stay in place so this can deploy in any order relative to the application change. Removing them while the old code is still running would blank a required setting and silently disable export screening — which is a legal requirement in production. Cleanup is a follow-up once mitodl/mitxpro#4067 is deployed everywhere.

### ⚠️ Before merging

`rest_key_id` and `rest_secret` must be added to the `cybersource` block of `src/bridge/secrets/xpro/secrets.{ci,qa,production}.yaml`. Those files are SOPS-encrypted, so I can't add them. Without them the `{{ get .Secrets "..." }}` template won't render and the app's secret will fail to sync.

The values come from the Test/Live Business Center for `mit_odl_xpro`: *Payment Configuration → Key Management → + Generate key → REST APIs → REST – Shared Secret*. The **Key** is `rest_key_id`, the **Shared Secret** is `rest_secret`.

### How can this be tested?

`pulumi preview` on the CI stack should show only the four new env vars on the xpro app. After deploy, `CYBERSOURCE_REST_*` are present in the pod environment, and once mitodl/mitxpro#4067 is deployed there, registration and login succeed with export compliance enabled.

### Additional Context

#5596 disabled the check on CI/RC by blanking `CYBERSOURCE_MERCHANT_ID`, a setting mitodl/mitxpro#4067 deletes. Once that app change deploys, the override is inert and screening is governed solely by whether these new settings are populated — unset means off, which is the safe default while credentials are pending. That override and its comment should be removed in the same follow-up that drops the SOAP settings.


arslanashraf7 (approved):
👍

PR #5667, opened 2026-08-31
chore(sentry): remove redundant SOPS Sentry DSN entries and dead airbyte plumbing

feoh:

Follow-up cleanup for https://github.com/mitodl/ol-infrastructure/issues/5004 (already closed — this is the final step from that issue's plan, not a reopen).

Deletes the hard-coded Sentry DSN entries that became dead weight once every consumer moved to `sentry_stack.require_output(...)` in #5222–#5231.

- **27 SOPS entries removed** across `edxapp` (12), `learn_ai`, `mitlearn`, `ocw_studio`, `mitxonline`, and `xpro` (3 each). Every one of these stacks already overwrites the SOPS value with the stack output before it reaches Vault, so the stored copy was unreachable.
- **`xpro`**: dropped the `{**xpro_vault_secrets["sentry"], "dsn": dsn}` spread. That block held only `dsn`, which the override immediately replaced, so the read was a no-op that pinned the SOPS entry in place. Now writes `{"dsn": dsn}` — identical payload; `k8s_secrets.py` reads only `dsn`.
- **`airbyte`**: removed the `secret-airbyte/sentry-dsn` Vault write and its 3 SOPS entries. This one is *not* a migration — the airbyte Sentry project and key were deleted from Sentry after the config import (see `IMPORT_SUMMARY.md`), so this path has been publishing a DSN for a project that no longer exists, and airbyte error reporting is already inert. The `secret-airbyte` mount stays, since `secret-airbyte/dagster` is still live for APISIX basic auth. Its README documented only the removed path, so it now documents the real one.

Not touched: `odl_video_service` and `micromasters` were never actually migrated and still read their DSNs live, so their entries stay.

The SOPS diff is large only because `sops` re-encrypts every value with a fresh IV on write. The semantic change is 30 removed keys.

### How can this be tested?

Verification run on this branch:

- `sops -d` on all 30 touched files — all decrypt cleanly (a hand-edited deletion would fail here with `MAC mismatch`).
- Decrypted-tree diff against the pre-change files: each differs by **exactly** the removed key, with every other value hashing identically.
- Recipient audit: KMS, PGP, and Vault-transit `enc` blobs are byte-identical before/after. `sops unset` reuses the existing data key, so no recipient was re-wrapped or dropped.
- Simulated each consumer's merge to confirm the resulting payload shape is unchanged (e.g. `{**edxapp_static_secrets, "sentry_dsn": dsn}` still yields `sentry_dsn`; the `xpro` loop no longer sees a `sentry` key; `mitxonline`'s collected-secrets blob no longer carries one).
- Full `pre-commit run` passes, including yamllint, detect-secrets, ruff, and mypy.

No `pulumi preview` was run — worth doing against one edxapp stack before merge to confirm the only Vault diff is the removed `secret-airbyte/sentry-dsn`.

### Additional Context

- Merging this makes Pulumi **delete** `secret-airbyte/sentry-dsn` in ci/qa/production on the next `up`. That's intended, and the DSN it holds points at a deleted Sentry project.
- Worth knowing for future secrets work: these files carry four PGP recipients that no longer appear in `.sops.yaml`'s creation rules. `sops unset` leaves them alone, but any edit path that re-encrypts from scratch would silently drop all four.


feoh:
### `pulumi preview` verification — `mitx.CI`

Ran an identical preview on this branch and on `main` as a baseline, same stack, same image digest (`sha256:536d890e…`, pulled from live stack state), `OPENEDX_RELEASE` left at its default `master` which is what mitx CI is on.

| | `main` (baseline) | this branch |
|---|---|---|
| To update | 2 | 2 |
| Unchanged | 223 | 223 |

The set of changed resource **types and URNs is byte-identical** between the two runs. The only textual difference in the raw output is the ordering of parallel `StackReference` reads, which is nondeterministic.

Both pending updates are pre-existing drift, present on `main` and untouched by this PR:

- `vault:index/mount:Mount` `secret-mitx` — dropping a stale `options: {version: "1"}`
- `fastly:index/tlsSubscription:TlsSubscription` `fastly-mitx-ci-tls-subscription`

**No diff on any Vault secret resource**, which is the expected result: removing `sentry_dsn` from SOPS can't change the payload, because `__main__.py` already overwrote that key with the sentry stack output before writing to Vault.

Note this covers the edxapp deployments only. The `secret-airbyte/sentry-dsn` deletion is in the airbyte stack and isn't exercised by this preview — that one is an intentional removal and will show as a delete when the airbyte stack next runs.

blarghmatey (approved):

PR #5723, opened 2026-09-02
feat(mit-learn-nextjs): set NEXT_PUBLIC_ORG_LEARNING_HUBSPOT_FORM_ID

daniellefrappier18:
Part of https://github.com/mitodl/hq/issues/13016 (closed by the mit-learn PR below, not by this one)

Pairs with https://github.com/mitodl/mit-learn/pull/3846

Sets `NEXT_PUBLIC_ORG_LEARNING_HUBSPOT_FORM_ID` in `raw_env_vars` for the mit-learn Next.js app, so the lead form on `/organizational-learning` can reach its HubSpot form in CI, QA and production.

Without it the var is unset in all three environments and the form renders a "contact form is unavailable" fallback instead. `NEXT_PUBLIC_STAY_UPDATED_HUBSPOT_FORM_ID` was already wired this way; this is the equivalent entry for the new page.

<details>
<br>

**Why an env var is required rather than a build-time value.** mit-learn resolves `NEXT_PUBLIC_*` at runtime, not through bundler inlining — the image is built in CI with no per-environment values, so the server reads `process.env` via dynamic bracket access and ships the values to the browser in an `x-public-env` meta tag. See [`env.ts`](https://github.com/mitodl/mit-learn/blob/main/frontends/main/src/env.ts). The pod environment is therefore the only source for this value in a deployed environment.

**Why this failed silently.** The app's yup schema declares the var optional ([`validateEnv.js`](https://github.com/mitodl/mit-learn/blob/main/frontends/main/validateEnv.js)), so `validateEnv()` does not fail at startup. Pods came up healthy with only the form area degraded, and `getOrgLearningHubspotFormId()` returned `""`.

**One ID across all three environments**, set as a flat constant rather than the `stack_info.env_suffix` dict used by `stay_updated_hubspot_form_ids` above. The dict plus its `try`/`except KeyError` only earns its keep when the values differ per environment; here they don't.

**The form is recycled from the Universal AI landing page**, which submits to the same ID ([`UAILandingPage.tsx`](https://github.com/mitodl/mit-learn/blob/main/frontends/main/src/app-pages/UAILandingPage/UAILandingPage.tsx)). Lead attribution between the two pages is preserved: `ol_hubspot`'s submit view falls back to the `Referer` header for `page_uri` when the client sends none, and forwards it to HubSpot in the submission `context`, so HubSpot records which page each lead came from.

</details>

### How can this be tested?
Nothing was deployed or previewed as part of this PR — `pulumi preview` needs stack credentials I don't have, so CI is the first real check. What was run locally: `pre-commit run --files src/ol_infrastructure/applications/mit_learn_nextjs/__main__.py`, all hooks passing (`ruff format`, `ruff check`, `mypy`, `detect-secrets`).

To validate after apply:

1. Confirm the var reached the pod:
   ```
   kubectl exec -n learn deploy/mitlearn-nextjs -c app -- printenv NEXT_PUBLIC_ORG_LEARNING_HUBSPOT_FORM_ID
   ```
   Expect `60a1983b-361a-4e80-a0db-d24ff636d7bf`.
2. Confirm it reaches the browser — load any page and check the `x-public-env` meta tag contains the key.
3. With the `organizational-learning` PostHog flag enabled for your user, visit `/organizational-learning`, choose **My organization**, and confirm the HubSpot form loads rather than showing the "contact form is unavailable" fallback.

### Additional Context
- **No merge-order constraint.** Nothing reads this var until mitodl/mit-learn#3846 ships, so this can merge first with no effect.
- **This does not expose the page.** `/organizational-learning` and both of its nav entries are gated client-side on the `organizational-learning` PostHog flag, which is not managed here. This PR only ensures the form works once that flag is turned on.
- The value is a HubSpot form ID, not a credential — it is sent to the browser by design, matching the existing plaintext `stay_updated_hubspot_form_ids` above.


src/ol_infrastructure/applications/mit_learn_nextjs/__main__.py
```diff
@@ -123,6 +123,18 @@
...
+    "ci": "e6910657-d832-4df8-be27-d8f6f9ecdfcc",
+    "qa": "e6910657-d832-4df8-be27-d8f6f9ecdfcc",
+    "production": "60a1983b-361a-4e80-a0db-d24ff636d7bf",
+}
+
+try:
+    org_learning_hubspot_form_id = org_learning_hubspot_form_ids[stack_info.env_suffix]
+except KeyError as exc:
+    msg = f"Unsupported MIT Learn Next.js environment: {stack_info.env_suffix}"
+    raise ValueError(msg) from exc
```
gumaerc: I think that if we raise an error here, it will stop the deploy. I don't think that's really a good course of action here. Realistically if whatever env is rendering isn't in the list and it can't get a form ID, it should just log an error and continue. Something like this maybe:

```python
org_learning_hubspot_form_id = org_learning_hubspot_form_ids.get(
    stack_info.env_suffix, ""
)
if not org_learning_hubspot_form_id:
    pulumi.log.warn(
        "No org learning HubSpot form ID for environment "
        f"{stack_info.env_suffix}; the form will render its unavailable fallback."
    )
```

gumaerc (changes requested):
The form IDs look correct, but I think there's one thing we should address:

src/ol_infrastructure/applications/mit_learn_nextjs/__main__.py
daniellefrappier18: @gumaerc is this what you mean?

gumaerc (approved):
👍

daniellefrappier18:
@zamanafzal - this PR is touching the same code you just edited. Do you mind taking a look to see if you agree with what I've done here? The hubspot_portal_id went from fail-fast to warn-and-continue.

zamanafzal:
Thanks for the heads-up @daniellefrappier18. One note on `NEXT_PUBLIC_STAY_UPDATED_HUBSPOT_FORM_ID`: per the decision in ([mitodl/hq#13041 (comment)](https://github.com/mitodl/hq/issues/13041#issuecomment-5438280791)) we went with "pass form ID per-page, retire the site-wide form." That shipped in [#3844](https://github.com/mitodl/mit-learn/issues/3844) — the product-page button is now gated only on the per-page hubspot_form_id, and the frontend no longer reads NEXT_PUBLIC_STAY_UPDATED_HUBSPOT_FORM_ID (no consumers left in frontends/). So that var + the stay_updated_hubspot_form_ids dict are orphaned now — worth dropping here or in a follow-up.

daniellefrappier18:
> Thanks for the heads-up @daniellefrappier18. One note on `NEXT_PUBLIC_STAY_UPDATED_HUBSPOT_FORM_ID`: per the decision in ([mitodl/hq#13041 (comment)](https://github.com/mitodl/hq/issues/13041#issuecomment-5438280791)) we went with "pass form ID per-page, retire the site-wide form." That shipped in [#3844](https://github.com/mitodl/mit-learn/issues/3844) — the product-page button is now gated only on the per-page hubspot_form_id, and the frontend no longer reads NEXT_PUBLIC_STAY_UPDATED_HUBSPOT_FORM_ID (no consumers left in frontends/). So that var + the stay_updated_hubspot_form_ids dict are orphaned now — worth dropping here or in a follow-up.

@zamanafzal @gumaerc 
I removed NEXT_PUBLIC_STAY_UPDATED_HUBSPOT_FORM_ID since it's no longer being used but now I'm questioning if the implementation of NEXT_PUBLIC_ORG_LEARNING_HUBSPOT_FORM_ID is correct because I was following the NEXT_PUBLIC_STAY_UPDATED_HUBSPOT_FORM_ID implementation as a guide.

zamanafzal:
> @zamanafzal @gumaerc I removed NEXT_PUBLIC_STAY_UPDATED_HUBSPOT_FORM_ID since it's no longer being used but now I'm questioning if the implementation of NEXT_PUBLIC_ORG_LEARNING_HUBSPOT_FORM_ID is correct because I was following the NEXT_PUBLIC_STAY_UPDATED_HUBSPOT_FORM_ID implementation as a guide.


`NEXT_PUBLIC_ORG_LEARNING_HUBSPOT_FORM_ID` looks correct to me. One static page, one form, so a global env var read by `getOrgLearningHubspotFormId()` in [`[OrgLeadForm.tsx](https://github.com/mitodl/mit-learn/blob/af811999bc792f39ece3c63d3c7cc37035c9c8e7/frontends/main/src/app-pages/OrganizationalLearningPage/OrgLeadForm.tsx)`](https://github.com/mitodl/mit-learn/blob/af811999bc792f39ece3c63d3c7cc37035c9c8e7/frontends/main/src/app-pages/OrganizationalLearningPage/OrgLeadForm.tsx) is the right pattern (unset just falls back to "unavailable"). Stay-updated only moved off env vars because it spans many CMS-managed pages — doesn't apply here. Removing the orphaned `STAY_UPDATED_HUBSPOT_FORM_ID` is right too. LGTM. 

@ChristopherChudzicki — tagging you in case you want to weigh in, since this ties back to the [hq#13041](https://github.com/mitodl/hq/issues/13041#issuecomment-5438280791) decision.

ChristopherChudzicki (commented):
@daniellefrappier18 You modeled `org_learning_hubspot_form_ids` after  `stay_updated_hubspot_form_ids` which was reasonable. However, that group of env vars was done somewhat differently than all the others, and I don't see a reason why.

In general, we have three types of env vars:
- **Same value across envs:** (e.g., `NEXT_PUBLIC_SITE_NAME`, `NEXT_PUBLIC_MITOL_SUPPORT_EMAIL`) These are in `__main__.py`
- **Interpolated value across envs:** (e.g., `NEXT_PUBLIC_LEARN_AI_CSRF_COOKIE_NAME`) These are in `__main__.py`
- **Distinct, arbitrary values across envs:** (e.g., `NEXT_PUBLIC_ORIGIN`, `NEXT_PUBLIC_ANALYTICS_API_BASE_URL`) These are stored in the sibling YAML files and retrieved either by `nextjs_config.get` (optional) or `nextjs_config.require` (required; errors if unset)

This variable falls in the third category. So:

**Suggestion:**
- Handle it like all the other varies-across-env variables and read it from the YAML files.
- Drop the warning altogether. We don't do that for other env vars.

ChristopherChudzicki:
As for the global env var vs per-page settings, I agree with @zamanafzal 's assessment:

> . One static page, one form, so a global env var read by getOrgLearningHubspotFormId()

👍

daniellefrappier18:
> @daniellefrappier18 You modeled `org_learning_hubspot_form_ids` after `stay_updated_hubspot_form_ids` which was reasonable. However, that group of env vars was done somewhat differently than all the others, and I don't see a reason why.
> 
> In general, we have three types of env vars:
> 
> * **Same value across envs:** (e.g., `NEXT_PUBLIC_SITE_NAME`, `NEXT_PUBLIC_MITOL_SUPPORT_EMAIL`) These are in `__main__.py`
> * **Interpolated value across envs:** (e.g., `NEXT_PUBLIC_LEARN_AI_CSRF_COOKIE_NAME`) These are in `__main__.py`
> * **Distinct, arbitrary values across envs:** (e.g., `NEXT_PUBLIC_ORIGIN`, `NEXT_PUBLIC_ANALYTICS_API_BASE_URL`) These are stored in the sibling YAML files and retrieved either by `nextjs_config.get` (optional) or `nextjs_config.require` (required; errors if unset)
> 
> This variable falls in the third category. So:
> 
> **Suggestion:**
> 
> * Handle it like all the other varies-across-env variables and read it from the YAML files.
> * Drop the warning altogether. We don't do that for other env vars.

@ChristopherChudzicki I add those ids to the YAML files and dropped the warning.  Does this look correct now?

ChristopherChudzicki (approved):
👍 Yes, that's how I would expect the env vars to be handled.

I guess `hubspot_portal_ids` was added to the `__main__` file since this was opened. That could be cleaned up here or separately. (One thing that's awkward is the comment ... i'd just put it in the prod yaml file).

PR #5724, opened 2026-09-02
fix(mitxonline): redirect home page to MIT Learn

alexfigtree:

Fixes mitodl/hq#12506
Subissue of mitodl/hq#12027

## Description (What does it do?)

The bare `https://mitxonline.mit.edu/` currently serves the mitxonline home page. Per mitodl/hq#12506, this should redirect to MIT Learn's MITx channel page instead: `https://learn.mit.edu/c/unit/mitx`.

This PR adds a new Fastly VCL rule matching only the exact root path (`req.url.path == "/"`), so it can't shadow `/cms`, `/staff-dashboard`, or any other route. As noted in the issue thread, this applies to **all** users regardless of login state — investigation confirmed there's no distinct logged-in experience at the bare `/` today (the dashboard lives at a separate `/dashboard/` route), so there's nothing to disrupt.

This PR introduces a new error code (605), following the same pattern as the existing UAI B2C (602), course/program (603), and catalog (604) redirects.

## Screenshots

N/A — infra-level redirect, no UI change.

## How can this be tested?
These have already been used to test:
- `uv run pre-commit run --files src/ol_infrastructure/applications/mitxonline/__main__.py` — ruff format/check, mypy, secret detection all pass.
- `uv run pytest tests/ol_infrastructure/lib/test_fastly.py` — 17/17 pass, including the repo-wide scan that every VCL snippet is built through the validating helper.
- Manually verified the exact-match condition against realistic and edge-case paths in Python: matches only `/`; does not match `/cms`, `/cms/login`, `/staff-dashboard`, `/staff-dashboard/foo`, `/courses/`, `/programs/`, `/catalog`, `/dashboard/`, etc.
- Checked this repo's source for any other use of custom error code 605 — none found (602/603/604 are the only other codes in use, sequential with no gaps). **This only confirms no collision in source; it does not confirm live Fastly state.**

Needs to get done after deploy:
- After deploy, confirm with `curl -I https://mitxonline.mit.edu/` — expect `301` with `Location: https://learn.mit.edu/c/unit/mitx`. Also confirm `/cms/login` and `/staff-dashboard/` still resolve normally (not redirected).

This also still needs to get done, but by devops:
- `pulumi preview` not run against live state in this session — **devops should confirm no existing/manual use of custom error code 605 on this Fastly service as part of the preview**, in addition to the general plan review (same as mitodl/hq#13100 / ol-infrastructure#5693).

## Additional Context

This work is very similar to other subissues in the Redirect epic. Same pattern as the existing `/catalog/*` redirect (mitodl/hq#12504), course/program redirect (mitodl/hq#12449), and bare `/courses/`/`/programs/` redirect (mitodl/hq#13100) — this repo's Fastly VCL snippet approach in `src/ol_infrastructure/applications/mitxonline/__main__.py`.

### DIFF
--- a/src/ol_infrastructure/applications/mitxonline/__main__.py
+++ b/src/ol_infrastructure/applications/mitxonline/__main__.py
@@ -1118,6 +1118,20 @@ catalog_redirect_vcl = (
     f"}}"
 )
 
+# Home page redirects to MIT Learn's unit catalog view too - per hq#12506,
+# matches only the exact site root, not any deeper path, so /cms and
+# /staff-dashboard (and everything else) are unaffected by construction.
+# Applies to all requests regardless of login state - there's no distinct
+# logged-in experience at bare "/" today for this to disrupt. Uses a
+# distinct error code (605) from the catalog (604), course/program (603),
+# and UAI B2C (602) redirects above so this is fully additive.
+home_redirect_vcl = (
+    f'if (req.url.path == "/") {{\n'
+    f'  set req.http.x-redir-location = "https://{learn_frontend_domain}/c/unit/mitx";\n'
+    f"  error 605;\n"
+    f"}}"
+)
+
 gzip_settings: dict[str, set[str]] = {"extensions": set(), "content_types": set()}
 for k, v in mimetypes.types_map.items():
     if k in (
@@ -1229,6 +1243,15 @@ mitxonline_service = fastly.ServiceVcl(
             type="recv",
         ),
         vcl_snippet(
+            content=home_redirect_vcl,
+            name="Redirect home page to MIT Learn",
+            # Exact match on "/" only, so it can't shadow /cms, /staff-dashboard,
+            # or any other path - ordering relative to the other recv snippets
+            # isn't load-bearing, placed after them for readability only.
+            priority=170,
+            type="recv",
+        ),
+        vcl_snippet(
             content=textwrap.dedent("""\
             if (obj.status == 602) {
               set obj.status = 301;
@@ -1282,6 +1305,25 @@ mitxonline_service = fastly.ServiceVcl(
               return(deliver);
             }"""),
             name="Handle catalog redirects to MIT Learn",
+            type="error",
+        ),
+        vcl_snippet(
+            content=textwrap.dedent("""\
+            if (obj.status == 605) {
+              set obj.status = 301;
+              set obj.response = "Moved Permanently";
+              set obj.http.Location = req.http.x-redir-location;
+              set obj.http.Cache-Control = "no-store";
+              if (req.url.qs != "") {
+                if (obj.http.Location !~ "\\?") {
+                  set obj.http.Location = obj.http.Location "?" req.url.qs;
+                } else {
+                  set obj.http.Location = obj.http.Location "&" req.url.qs;
+                }
+              }
+              return(deliver);
+            }"""),
+            name="Handle home page redirects to MIT Learn",
             type="error",
         ),
         vcl_snippet(
--- a/src/ol_infrastructure/applications/mitxonline/__main__.py
+++ b/src/ol_infrastructure/applications/mitxonline/__main__.py
@@ -1315,11 +1315,9 @@ mitxonline_service = fastly.ServiceVcl(
               set obj.http.Location = req.http.x-redir-location;
               set obj.http.Cache-Control = "no-store";
               if (req.url.qs != "") {
-                if (obj.http.Location !~ "\\?") {
-                  set obj.http.Location = obj.http.Location "?" req.url.qs;
-                } else {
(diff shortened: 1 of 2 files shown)


feoh:
Validated this against live Fastly state via the API. **The VCL is correct and 605 is genuinely free** — but the change silently blinds two production Pingdom checks, which I think should be fixed before merge.

### ✅ Error code 605 is unused (confirmed live, not just in source)

Scanned the **active** versions of both services this stack manages:

| Service | ID | Active version | Custom codes found |
|---|---|---|---|
| MITx Online Application Production | `54YRiE18QJhCPQRlVmFHFm` | v209 | 602, 603, 604 |
| MITx Online Application QA | `20oLGuxlN8Gj7qnRhHyred` | v565 | 602, 603, 604 |

Both have exactly 9 snippets; I checked every one for `error NNN` and `obj.status == NNN`. Also confirmed **no custom VCL files, no response objects, and no status-referencing request conditions** on either service, so there's no non-snippet hiding place. 605 is clear on both. The CI stack has no deployed Fastly service yet, so nothing to check there.

This covers the "devops should confirm no existing/manual use of 605" item in the description — consider it confirmed.

One correction to the PR body: `error 605` **does** already appear in this repo, at `src/ol_infrastructure/applications/ocw_site/snippets/redirects.vcl:62`. It's harmless (error codes are scoped per Fastly service, and that's the OCW service), but the "no other use of 605 in this repo" claim isn't accurate and will trip up anyone who greps.

### ✅ Other things that checked out

- **Escaping is right.** I pulled the deployed 604 error snippet from Fastly and AST-parsed the new 605 literal: the generated VCL is byte-for-byte identical to the live, proven 604 except the status code. Worth stating explicitly, since `"\\?"` inside a non-raw `textwrap.dedent` is exactly where this silently breaks.
- **Match scoping is as claimed.** Simulated all four recv rules; only `/` hits 605. Confirmed live that `/cms/login` and `/staff-dashboard/` currently 302 to login and `/dashboard/` is 200 — all unaffected. The "ordering isn't load-bearing" comment on priority 170 is accurate.
- **Env-awareness is correct.** `learn_frontend_domain` derives per stack (prod → `learn.mit.edu`, QA → `rc.learn.mit.edu`, CI → `ci.learn.mit.edu`), and each service has exactly one domain, so there's no host-scoping leak onto `api.mitxonline.mit.edu`.
- **Login flow is safe.** `/login/` goes to Keycloak with `redirect_uri=.../login/.apisix/redirect`, and `oidc_post_logout_redirect_uri` targets `/logout/`, not `/`. Nothing lands on bare `/`.

### ⚠️ Blocking-ish: this blinds two Pingdom checks

`src/ol_infrastructure/infrastructure/grafana_alerting/pingdom_checks.py:346` targets `https://mitxonline.mit.edu/` with `alert_sensitivity="high"`, and `:354` targets `https://rc.mitxonline.mit.edu/`. Both are plain HTTP checks with no content assertion.

Because `error 605` is raised in `vcl_recv`, the 301 is synthesized **at the Fastly edge and never reaches the K8s backend**. So once this deploys, the production check reports green off the redirect even if MITx Online is completely down — it stops testing MITx Online and starts transitively testing MIT Learn. That's a silent loss of production monitoring rather than a visible failure, which is what makes it easy to miss.

Suggest repointing both checks at a path that actually exercises the backend. `/dashboard/` returns 200 and is not redirected. Note there's no health endpoint on the Fastly-fronted domain — `/heartbeat/`, `/health/`, and `/healthz/` all 404 — unlike the edX checks above them, which use `/heartbeat`.

### Minor

Query strings survive the redirect, so `/?utm_source=x` → `https://learn.mit.edu/c/unit/mitx?utm_source=x`. Same as 603/604, and probably what you want for the homepage.

---

Also ran on the branch: `ruff format --check` and `ruff check` clean, `tests/ol_infrastructure/lib/test_fastly.py` 17/17. I did **not** run `pulumi preview` against live state.

Ardiea:
Reviewed and pushed af51acca2 with two small fixes:

- **Pingdom**: the prod/RC uptime checks hit bare `/`, which Fastly now answers with a synthetic 301 without touching the origin — they'd stay green through a real outage. Repointed them at `/health/readiness/` (verified 200 on both hosts).
- **605 handler**: removed the unreachable `"&" qs` branch — the redirect target is a fixed URL, so it can never already contain `?`.

Two things we noted and are fine with: logout redirects to `SITE_BASE_URL` (`/`), so logging out now lands users on the MIT Learn catalog — accepted. And the 602 handler is the only redirect handler without `Cache-Control: no-store`; pre-existing, left as-is.

LGTM otherwise. :shipit:

Ardiea (approved):

feoh (approved):

Issue #5690, opened 2026-09-01
mitxonline production Meilisearch is memory-starved: 4Gi limit against a 9.31Gi live index


AhtishamShahid:
## Ask

Raise the memory allocation for `meilisearch-0` in `mitxonline-openedx`:

| knob | current | requested |
|---|---|---|
| `meilisearch:memory_limit` | `4Gi` | **`16Gi`** (minimum viable `14Gi`) |
| `meilisearch:memory_request` | `1Gi` | **`12Gi`** |
| `meilisearch:cpu_request` | `250m` | **`2`** (no CPU limit, per house practice) |

Config lives at `src/ol_infrastructure/applications/edxapp/Pulumi.mitxonline.Production.yaml`;
`meilisearch.py:136-145` already plumbs all three keys through. **No code change required.**

### Where 16Gi comes from

```
usedIndexSize (must be resident)                 9.31 GiB
MEILI_MAX_INDEXING_MEMORY (already pinned)       2.00 GiB
process heap / base                             ~1.00 GiB
                                                ---------
working set                                     ~12.31 GiB
+ headroom for growth and transients            -> 16 GiB limit
```

Request is sized to resident index + heap (~10.3 GiB) plus slack -> 12Gi. Kubernetes schedules on the
**request**, so a token request lets the node be packed until node-level reclaim evicts the page cache,
which reproduces the same thrashing the limit was raised to prevent.

Note `indexSize` is **11.91 GiB** but that includes LMDB free pages which never need to be resident.
`usedIndexSize` (9.31 GiB) is the correct basis. CPU is raised because once paging stops, indexing
becomes CPU-bound and 250m guaranteed would just relocate the ceiling.

### This is explicitly an interim measure

987 of 1,323,394 documents in `studio_content` are Libraries V2 - **99.925% is course XBlock content**.
If course indexing is gated (tracked from mitodl/hq#13014), the index drops to ~1,000 documents and 4Gi
becomes generous, at which point this allocation can be reverted. That work needs an upstream
edx-platform change and a decision about losing Studio course search, so it is not a near-term option.

---

### Config
- [ ] `Pulumi.mitxonline.Production.yaml` sets `memory_limit: 16Gi` (>= `14Gi`), `memory_request: 12Gi`,
      `cpu_request: 2`, with no CPU limit
- [ ] **The Pulumi stack is actually applied to `applications-production`.** Merged is not deployed -
[...]

PR #5734, opened 2026-09-03
fix(edxapp): raise mitxonline Meilisearch volume throughput to 500 MB/s

feoh:

Relates to https://github.com/mitodl/ol-infrastructure/issues/5690 and https://github.com/mitodl/hq/issues/13014. Closes neither — this raises the IO ceiling, it does not fix the undersized memory limit that causes the re-reading.

mitxonline production Meilisearch holds a 12.02 GiB `studio_content` index in a pod with a 4Gi cgroup limit. Only ~2.26 GB of it can stay in page cache (`memory.stat` `file`), so the rest is re-read from EBS on every commit, and the volume runs at its throughput ceiling.

- Adds an `ebs-gp3-throughput-500` VolumeAttributesClass next to the existing `ebs-gp3-sc` StorageClass, plus `ebs-gp3-throughput-125` as an explicit rollback target. Declared for every cluster; they cost nothing until a PVC names one.
- Adds a `meilisearch:volume_attributes_class` config key. When set, the Helm-created PVC is patched to reference the class. Emitted only where configured, so the other eleven edxapp stacks render unchanged.
- Sets it on mitxonline Production only.

**Why a VolumeAttributesClass and not StorageClass parameters:** StorageClass parameters apply at provisioning time and cannot revise a volume that already exists. A VolumeAttributesClass drives `ec2:ModifyVolume` against the live volume instead. That matters beyond convenience here — the PVC has required node affinity on `us-east-1a` and `meilisearch.py` pins the pod to `ol.mit.edu/core_node=true`, so anything that forces a reschedule risks leaving `meilisearch-0` `Pending`. That risk is what closed https://github.com/mitodl/ol-infrastructure/pull/5611.

**Why a PVC patch and not a chart value:** the meilisearch-kubernetes 0.38.0 `templates/pvc.yaml` has no field for `volumeAttributesClassName`.

#### Measurements

15 consecutive minutes, 17:20–17:34 UTC on 2026-09-03, `vol-09030d59932897b45`, CloudWatch at 60s granularity:

| | measured | provisioned |
|---|---|---|
| read throughput | 110 MB/s mean, **122 MB/s peak** | 125 MB/s |
| IOPS | 1,464/s mean, 1,904/s peak | 5,000 |

98% of the throughput ceiling at peak against 38% of the IOPS budget, at a mean read size of 77 KiB. Total read over the window was 91.8 GiB against a 12.02 GiB index — the whole index 7.6 times over.

Meilisearch's own batch trace attributes 44.26s of a 47.45s single-document commit to `post processing facets > facet search`, and that cost is flat against batch size (1 document: 52.9s; 1,741 documents: 47.5s), which is what makes per-XBlock indexing during a course import so expensive.

### How can this be tested?

`pulumi preview --diff`:

- `infrastructure/aws/eks` @ `applications.Production` → `+ 2 to create, 175 unchanged`; the two VolumeAttributesClass objects and nothing else.
- `applications/edxapp` @ `mitxonline.Production` → adds one resource from this change, `kubernetes:core/v1:PersistentVolumeClaimPatch` with `volumeAttributesClassName: "ebs-gp3-throughput-500"`.

`pre-commit run --from-ref origin/main --to-ref HEAD` passes (ruff, ruff format, mypy, yamllint, detect-secrets).

**Apply order matters:** the EKS stack has to go first. There are currently no VolumeAttributesClass objects in `applications-production`, and per `kubectl explain pvc.spec.volumeAttributesClassName`, a claim naming a class that does not exist is held `Pending` under `status.modifyVolumeStatus` until it appears. The claim stays bound and the pod keeps running, so this delays the modification rather than causing an outage — but it is a silent no-op until the class exists.

**After applying**, verify against the volume rather than the PVC, since the PVC field reflects intent and the volume reflects the result:

```bash
kubectl get pvc meilisearch -n mitxonline-openedx \
  -o jsonpath='{.status.currentVolumeAttributesClassName}{"\n"}'

aws ec2 describe-volumes --volume-ids vol-09030d59932897b45 \
  --query 'Volumes[0].[Iops,Throughput]' --output text   # expect: 5000   500
```

Then re-read a single-document commit trace and check that the `post processing facets > facet search` step falls roughly in proportion to the throughput increase:

```bash
kubectl exec -n mitxonline-openedx meilisearch-0 -- sh -c \
  'curl -s -H "Authorization: Bearer $MEILI_MASTER_KEY" localhost:7700/batches?limit=1'
```

I have not applied this. The predicted improvement is inferred from the measured saturation, not observed.

### Additional Context

- EBS permits one modification per volume per six hours, so a correction here is slow to land — which is also why the rollback path is a named class rather than a revert.

**Rollback.** Reverting the config key does not undo this. Clearing `volumeAttributesClassName` means "no class applies"; it does not call `ec2:ModifyVolume`, so the volume keeps its last geometry and the revert silently leaves production at 500 MB/s. Roll back by pointing the claim at `ebs-gp3-throughput-125`, confirming `Throughput` reads `125` on the volume, and only then dropping the key. Documented in MEILISEARCH.md.
- `iops` is pinned to `5000` in the class rather than omitted, matching what `iopsPerGB: 50` yields for a 100Gi volume, so the class states the full target geometry rather than inheriting whatever the volume happens to carry.
- The durable fix is still the memory limit. https://github.com/mitodl/ol-infrastructure/issues/5690 was closed as completed on 2026-09-01, but nothing was applied — the repo and the live pod both still read `memory_limit: 4Gi` / `memory_request: 1Gi`. This PR is deliberately independent of that work: it needs no scheduling headroom and no pod restart, so it does not have to win the argument that closed #5611 first.
- Meilisearch QA cannot be used to validate the performance effect. Its index is 2.2 GiB and fits inside the same 4Gi limit, so the facet step there runs in 53ms rather than 44s and there is no saturation to relieve.

### DIFF
--- a/src/ol_infrastructure/applications/edxapp/meilisearch.py
+++ b/src/ol_infrastructure/applications/edxapp/meilisearch.py
@@ -164,7 +164,7 @@ def create_meilisearch_resources(
             max_indexing_memory
         )
 
-    return kubernetes.helm.v3.Release(
+    meilisearch_release = kubernetes.helm.v3.Release(
         f"ol-{stack_info.env_prefix}-edxapp-meilisearch-helm-release-{stack_info.env_suffix}",
         kubernetes.helm.v3.ReleaseArgs(
             name="meilisearch",
@@ -180,3 +180,31 @@ def create_meilisearch_resources(
         ),
         opts=ResourceOptions(delete_before_replace=True),
     )
+
+    # Raise the throughput of an already-provisioned volume. The storageclass sets
+    # gp3 throughput at provisioning time and cannot revise it afterwards, so a
+    # volume whose index has outgrown the pod's page cache stays pinned at whatever
+    # ceiling it was born with. Pointing the PVC at a VolumeAttributesClass drives
+    # ec2:ModifyVolume against the live volume instead -- online, no pod restart, no
+    # rebind, so it does not disturb the zone-locked PVC.
+    #
+    # The chart's PVC template has no field for this (meilisearch-kubernetes 0.38.0),
+    # hence a patch against the Helm-created claim rather than a values entry.
+    # Emitted only where configured, so the stacks that have not been sized for it
+    # render unchanged.
+    if volume_attributes_class := meilisearch_config.get("volume_attributes_class"):
+        kubernetes.core.v1.PersistentVolumeClaimPatch(
+            f"ol-{stack_info.env_prefix}-edxapp-meilisearch-pvc-attributes-{stack_info.env_suffix}",
+            metadata=kubernetes.meta.v1.ObjectMetaPatchArgs(
+                name="meilisearch",
+                namespace=namespace,
+                labels=k8s_global_labels,
+                annotations={"pulumi.com/patchForce": "true"},
+            ),
+            spec=kubernetes.core.v1.PersistentVolumeClaimSpecPatchArgs(
+                volume_attributes_class_name=volume_attributes_class,
+            ),
+            opts=ResourceOptions(depends_on=[meilisearch_release]),
+        )
+
+    return meilisearch_release
--- a/src/ol_infrastructure/infrastructure/aws/eks/__main__.py
+++ b/src/ol_infrastructure/infrastructure/aws/eks/__main__.py
@@ -842,6 +842,41 @@ if eks_config.get_bool("ebs_csi_provisioner"):
             depends_on=[ebs_csi_driver_role, *node_groups],
         ),
     )
+    # gp3 volumes provisioned by the storageclass above get 125 MB/s, the AWS
+    # default. That is a throughput ceiling, not a starting point: a volume whose
+    # working set does not fit in its consumer's page cache re-reads from EBS
+    # continuously and saturates it. Over fifteen consecutive minutes on 2026-09-03,
+    # mitxonline production Meilisearch averaged 110 MB/s against that ceiling and
+    # peaked at 122 MB/s (98% of it), while peak IOPS reached 1,904/s -- 38% of the
+    # 5,000 the storageclass had already provisioned. Throughput bound it; IOPS did
+    # not. Mean read size was 77 KiB, which is why.
+    #
+    # StorageClass parameters only apply at provisioning time, so they cannot fix an
+    # existing volume. A VolumeAttributesClass can: setting it on a live PVC drives
+    # ec2:ModifyVolume in place, with no pod restart and no rebind. This is declared
+    # for every cluster because it costs nothing until a PVC references it by name.
+    #
+    # iops is pinned to 5,000 rather than omitted so the class states the full target
+    # geometry instead of silently inheriting whatever the volume happens to carry.
+    # That matches what `iopsPerGB: 50` yields for the 100Gi volumes this is aimed at.
+    kubernetes.storage.v1.VolumeAttributesClass(
+        resource_name=f"{cluster_name}-ebs-gp3-throughput-500-volumeattributesclass",
+        metadata=kubernetes.meta.v1.ObjectMetaArgs(
+            name="ebs-gp3-throughput-500",
+            labels=k8s_global_labels,
+        ),
+        driver_name="ebs.csi.aws.com",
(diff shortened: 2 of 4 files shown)


src/ol_infrastructure/infrastructure/aws/eks/__main__.py
```diff
@@ -842,6 +842,41 @@ def get_node_role_arn_pattern(cluster_name: str):
...
+    # StorageClass parameters only apply at provisioning time, so they cannot fix an
+    # existing volume. A VolumeAttributesClass can: setting it on a live PVC drives
+    # ec2:ModifyVolume in place, with no pod restart and no rebind. This is declared
+    # for every cluster because it costs nothing until a PVC references it by name.
+    #
+    # iops is pinned to 5,000 rather than omitted so the class states the full target
+    # geometry instead of silently inheriting whatever the volume happens to carry.
+    # That matches what `iopsPerGB: 50` yields for the 100Gi volumes this is aimed at.
+    kubernetes.storage.v1.VolumeAttributesClass(
+        resource_name=f"{cluster_name}-ebs-gp3-throughput-500-volumeattributesclass",
```
blarghmatey, line 863: Scoping this to the cluster will error out on a duplicate resource across xPRO and MITx Online, as well as MITx and MITx-Staging

Ardiea (approved):

src/ol_infrastructure/applications/edxapp/meilisearch.py
```diff
@@ -180,3 +180,31 @@ def create_meilisearch_resources(
...
+    # volume whose index has outgrown the pod's page cache stays pinned at whatever
+    # ceiling it was born with. Pointing the PVC at a VolumeAttributesClass drives
+    # ec2:ModifyVolume against the live volume instead -- online, no pod restart, no
+    # rebind, so it does not disturb the zone-locked PVC.
+    #
+    # The chart's PVC template has no field for this (meilisearch-kubernetes 0.38.0),
+    # hence a patch against the Helm-created claim rather than a values entry.
+    # Emitted only where configured, so the stacks that have not been sized for it
+    # render unchanged.
+    if volume_attributes_class := meilisearch_config.get("volume_attributes_class"):
```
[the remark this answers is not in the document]
feoh: Fixed in afd399a6d: added `ebs-gp3-throughput-125` as an explicit rollback target, and documented the procedure in MEILISEARCH.md — point the claim at the baseline class, confirm `Throughput` reads 125 on the volume, then drop the reference. Confirmed against the API contract (`kubectl explain pvc.spec.volumeAttributesClassName`): an empty or nil value "indicates that no VolumeAttributesClass will be applied", which is not the same as restoring the prior geometry, so a plain revert would have stranded production at 500.

src/ol_infrastructure/infrastructure/aws/eks/__main__.py
feoh: I don't think this one collides, and I'd rather you double-check my reasoning than take my word for it.

`cluster_name` here is `f"{stack_info.env_prefix}-{stack_info.env_suffix}"` (line 128), so it's the cluster, not the deployment — and the EKS project has exactly 12 stacks for 12 clusters, so this creates one object per cluster. The `ebs-gp3-sc` StorageClass 40 lines up uses the identical shape (`resource_name=f"{cluster_name}-..."` with a fixed cluster-scoped `name`) and has run 638 days across every cluster; live labels show single ownership, e.g. `ebs-gp3-sc` in applications-production is owned by `infrastructure.aws.eks.applications.Production` alone.

The pairings you name do share clusters, but at the edxapp layer — xpro and mitxonline are both `applications_vpc`, mitx and mitx-staging both `residential_mitx_vpc`. My per-deployment resource there is the namespaced PersistentVolumeClaimPatch, which can't collide.

That said, the collision you're describing is exactly what happens if this class gets moved into `meilisearch.py` next to the patch, which is where someone would naturally put it. Happy to add a comment pinning it to the cluster stack if you want that guard.

blarghmatey: Roger that, I was conflating this with the edxapp deploy. Ship it 👍

blarghmatey (approved):

PR #5791, opened 2026-09-08
chore: clear the remaining deptry findings

Ardiea:

N/A — follow-up to #5789, which added the deptry config. With its exclusions in place the remaining findings were all real, so this clears them at the source rather than by widening the config.

- Declares three packages we import but never declared: `pulumi-kubernetes`, `pyyaml`, `botocore`. All three were arriving transitively.
- Removes three declared packages that nothing imports: `pulumi-docker`, `pulumi-aws-native`, `pulumi-terraform`.
- Removes `[tool.deptry.package_module_name_map]`, which turns out to be dead config.
- `deptry` now exits 0 (was 1711 findings before #5789, 130 after it).

<details>
<br>

**The declarations.** `pulumi_kubernetes` is imported by 98 files and reached us only through `pulumi-eks`, a thin wrapper we may well drop — that one is the actual fragility. `yaml` and `botocore` are lower stakes (three declared deps supply PyYAML; botocore comes from boto3) but were equally undeclared. Adding all three changes **no resolved version** — the lock diff contains zero `+version`/`-version` lines apart from the three removals below. `botocore`'s floor is deliberately loose (`>=1.24`) so it does not fight boto3, which re-pins it to a narrow range (`>=1.43.85,<1.44.0`) on every bump.

**The removals.** These are the only part of this PR with any deploy-time surface, since nothing else in the tree requires them — they leave the venv entirely rather than merely becoming undeclared.

| Package | Imported anywhere? | Last use |
|---|---|---|
| `pulumi-terraform` | Never, in the entire repo history (`git log -S` returns no commits) | — |
| `pulumi-aws-native` | No | `iam_native.ManagedPolicy` in `learn_ai/__main__.py`, removed in ada09546e (Feb 2025) |
| `pulumi-docker` | No | `docker.Image` in the kubewatch webhook handler, removed in 15977ad97 (Dec 2025) |

`pulumi-terraform` is unrelated to the `pulumi-terraform-bridge` that generated the vendored `sdks/` — the config comment makes those easy to conflate, but the Python SDK was never used here.

**The dead map.** deptry infers `pyjwt` → `jwt` and `ol-parliament` → `parliament` from each distribution's `top_level.txt`, so the explicit mapping was doing nothing. Verified by deleting it: output is unchanged, and if it were load-bearing both packages would surface as DEP002.

</details>

### How can this be tested?

Verified on this branch:

- `uv run --with deptry deptry .` → `Success! No dependency issues found.` (exit 0)
- `uv run pre-commit run --files pyproject.toml uv.lock` → passes
- `git diff uv.lock` → only the three removed package blocks; no other package changed version

**Not verified, and the one thing worth a reviewer's time:** I could not reach `s3://mitol-pulumi-state/`, so I have no direct evidence about what the live stacks hold. A `pulumi preview` on **learn_ai** and **kubewatch_webhook_handler** — the only two stacks that ever used the removed providers — would confirm nothing is stranded. I expect no diff, but that is an expectation, not an observation.

### Additional Context

The removals are the only risky hunk; the declarations and the map deletion are inert by comparison. If the preview turns up anything unexpected, dropping the three removals and keeping the rest still gets deptry to a clean exit for the DEP003 class.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -19,7 +19,6 @@ dependencies = [
   "hvac[parser]>=2.0.0,<3",
   "pulumi>=3.39.1,<4",
   "pulumi-aws>=7.1,<8",
-  "pulumi-docker>=5,<6",
   "pulumi-fastly>=12,<13",
   "pulumi-gcp>=9.34,<10",
   "pulumi-github>=6.0.0,<7",
@@ -35,7 +34,6 @@ dependencies = [
   "pydantic-settings>=2.0.1,<3",
   "pulumiverse-heroku>=1.0.3,<2",
   "pulumi-eks>=4,<5",
-  "pulumi-aws-native>=1.25.0,<2",
   "packaging>=24.2",
   "kubernetes>=34.1.0",
   "ol-parliament>=1.7.0",
@@ -44,12 +42,18 @@ dependencies = [
   "requests-aws4auth>=1.3.1",
   "pulumi-qdrant-cloud",
   "pulumi-rootly",
-  "pulumi-terraform>=6.0.1",
   "pulumi-command>=1.0,<2",
   "ol-concourse>=0.18,<0.19",
   "pulumiverse-grafana>=2.0.0,<3",
   "pulumiverse-sentry==0.0.9",
   "pyjwt[crypto]>=2.9,<3",
+  # Imported directly across the codebase; declared here rather than inherited.
+  # pulumi-kubernetes reaches us only through pulumi-eks, a thin wrapper we may
+  # well drop -- 98 files import it. botocore's floor stays loose on purpose so
+  # it does not fight boto3, which pins it to a narrow range on every bump.
+  "pulumi-kubernetes>=4.34,<5",
+  "pyyaml>=6,<7",
+  "botocore>=1.24",
 ]
 
 [project.urls]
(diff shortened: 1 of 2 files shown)


Ardiea:
Closing in favour of #5792, which covers the same three declarations and is correct in a case this PR gets wrong.

**#5792 is right about `package_module_name_map`, and this PR was wrong to delete it.** I measured the block as inert, but only under `uv run --with deptry deptry .`, where the project's dependencies are installed and deptry can read `pyyaml`'s `top_level.txt` to infer `pyyaml` → `yaml`. Under `uvx deptry .` the project's packages are not installed, deptry cannot read that metadata, and it falls back to assuming the module name equals the package name. Measured on both branches:

| | `uvx deptry .` | `uv run --with deptry deptry .` |
|---|---|---|
| this PR | **29** (26 DEP001 + 3 DEP002) | 0 |
| #5792 | 3 | 3 |

This branch's 29 are exactly what #5792's description predicts: 23 × `DEP001 'yaml'` plus `DEP002 'pyyaml'`, and the same shape for `pyjwt` and `ol-parliament`. So the mapping is inert when the metadata is reachable and load-bearing when it is not — keeping it is correct either way, and #5792 produces the same answer in both modes while this one does not.

Worth capturing separately: **the issue count depends on how deptry is invoked**, which neither PR's config comments mention. Whenever deptry gets wired into CI (it is not wired anywhere today), the invocation should be pinned, since `uvx` and `uv run` disagree.

**Still outstanding after #5792 merges** — the three unused Pulumi SDKs, which #5792 correctly defers pending a stack-state check. Evidence gathered here, to save repeating it:

| Package | Imported anywhere? | Last use |
|---|---|---|
| `pulumi-terraform` | Never, in the entire repo history (`git log -S` returns no commits) | — |
| `pulumi-aws-native` | No | `iam_native.ManagedPolicy` in `learn_ai/__main__.py`, removed in ada09546e (Feb 2025) |
| `pulumi-docker` | No | `docker.Image` in the kubewatch webhook handler, removed in 15977ad97 (Dec 2025) |

None is required by anything else in the tree, so removing them drops them from the venv rather than merely undeclaring them. `pulumi-terraform` is unrelated to the `pulumi-terraform-bridge` that generated the vendored `sdks/`. The open question is whether any live stack still holds resources from these providers — a `pulumi preview` on **learn_ai** and **kubewatch_webhook_handler** is the check.

Branch `md/202609_deptry_dependency_cleanup` is left in place for whoever picks that up.

PR #5797, opened 2026-09-08
feat(local-dev): make the Keycloak image a per-developer knob

ChristopherChudzicki:
N/A

Lets a developer run local-dev Keycloak on an image built from an [ol-keycloakify](https://github.com/mitodl/ol-keycloakify) checkout, so an unreleased theme branch can be reviewed in the k3d stack before it ships. One command:

```bash
./local-dev/scripts/kc-theme-image.sh          # ../ol-keycloakify, or pass a path
./local-dev/scripts/kc-theme-image.sh --reset  # back to the published image
```

- New `keycloak_image` knob: `tilt_config.json` → `LOCAL_DEV_KEYCLOAK_IMAGE` → core Pulumi stack → Keycloak CR `spec.image`. Unset, it falls back to the published `mitodl/keycloak` digest the stack already pins, so nothing changes by default.
- New `local-dev/keycloak/Dockerfile`, multi-stage: a native-arch node stage builds the theme jar from the checkout (yarn + Maven, no host toolchain needed), and the `linux/amd64` Keycloak stage bakes it in and reruns `kc.sh build` with the same flags as ol-keycloak's `Dockerfile.hosted`. The base digest is read from `local-dev/infra/core/__main__.py` at build time, so there is one copy.
- New `local-dev/scripts/kc-theme-image.sh` (140 lines, a third of them comments): build, push to the k3d registry, verify the pushed manifest, write the knob. The tag hashes the checkout's HEAD, its uncommitted changes, and the Dockerfile, so each change rolls Keycloak once and an unchanged re-run is a cached build.
- README section "Testing a local ol-keycloakify build".
- Enables the `UPDATE_EMAIL` required action in the local olapps realm, mirroring production `olapps.py`. The module skipped it behind a comment saying the alias was removed in Keycloak 26; it exists in 26.5.7 and was only disabled.

<details>
<br>

- The CR runs Keycloak with `--optimized`, which refuses to start when a jar in `providers/` differs from the one recorded at build time, so a theme jar cannot be copied or mounted into the running pod. The image has to be rebuilt.
- Same env-var pattern as `log_retention_period`: the value is deliberately not pinned in `Pulumi.local-dev.core.Dev.yaml`, since Pulumi config would beat the env var.
- The published base image is `linux/amd64` only and a plain single manifest, which is why it runs on the arm64 k3d nodes at all. BuildKit's default OCI index (image + attestation) makes containerd platform-match and reject the pull, so the script builds with `--provenance=false --sbom=false` and checks the registry's stored media type after pushing.
- Pushes go to `localhost:5001` (the daemon insists on TLS for the `k3d-registry.localhost` name); the config gets the in-cluster name `k3d-registry.localhost:5000/...`, the same host Tilt's own images use.
- `Dockerfile.dockerignore` keeps `node_modules` and build output out of the context (10 MB instead of >1 GB). hadolint is told not to parse it as a Dockerfile.
- No Tilt-driven rebuild on save, on purpose: every Keycloak roll logs every local session out, and the Maven + `kc.sh build` cycle is 1–2 minutes, so this stays a deliberate command.
</details>

### How can this be tested?
Ran end to end on this branch with the theme from https://github.com/mitodl/ol-keycloakify/pull/196:

1. `./local-dev/scripts/kc-theme-image.sh` against a dirty ol-keycloakify checkout. Full build about 2.5 min (`kc.sh build` under emulation is 54 s of it); an unchanged re-run is 1.5 s (all stages cached); editing the Dockerfile produced a new tag.
2. Tilt reran `local-infra-core`, the operator rolled `keycloak-0`, and the pod came up on the new tag with the built jar as the only theme jar in `providers/`. The jar in the pulled image contained the checkout's uncommitted edits.
3. Triggered `UPDATE_EMAIL` from the account console as a seeded user. Mailpit received the new `email-update-confirmation` template and the confirm link completed the change.
4. `pulumi preview` on the core stack with the knob unset shows no Keycloak diff; set, only `spec.image` changes.
5. `--reset`, `--help`, and an unknown flag behave as documented. pre-commit (shellcheck, hadolint, ruff, mypy) passes.

The `UPDATE_EMAIL` change was applied to a running realm via Tilt's `local-infra-apps`; the provider adopted the built-in alias in place.

### Additional Context
- The pinned digest is Keycloak 26.5.7 (April build) while `Dockerfile.hosted` is on 26.7.1. Bumping it is now a one-place change in `local-dev/infra/core/__main__.py` plus rechecking the `kc.sh build` flags. Left for a separate PR.
- Two Claude subagent reviews (correctness and developer experience) were run on the branch; their accepted findings are folded in. Deliberately not taken, to keep the script small: a `sha256sum` fallback for hosts without `shasum`, and printing the root-domain account-console URL from the script.

🤖 Co-authored with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -140,3 +140,6 @@ repos:
   rev: v2.15.1
   hooks:
   - id: hadolint-docker
+    # BuildKit reads <Dockerfile>.dockerignore beside a Dockerfile; it is not
+    # Dockerfile syntax.
+    exclude: \.dockerignore$
--- a/Tiltfile
+++ b/Tiltfile
@@ -14,6 +14,7 @@ config.define_string_list("prebuilt_tags", usage="Prebuilt image tag overrides p
 config.define_string("disk_keep_tags", usage="Newest tilt-built image tags kept per repo by the disk janitor (default: 3). Overrides LOCAL_DEV_DISK_KEEP_TAGS env var.")
 config.define_string("disk_buildcache_max_gb", usage="Docker build-cache size cap in GB (default: 10% of total disk; 0 disables). Overrides LOCAL_DEV_BUILDCACHE_MAX_GB env var.")
 config.define_string("log_retention_period", usage="How long Grafana/Loki keeps local-dev logs, as a whole number of days, e.g. 72h or 3d (default: 168h). Overrides LOCAL_DEV_LOG_RETENTION env var.")
+config.define_string("keycloak_image", usage="Keycloak server image for the core stack (default: the published mitodl/keycloak digest). Set and cleared by local-dev/scripts/kc-theme-image.sh to test an unreleased ol-keycloakify theme. Overrides LOCAL_DEV_KEYCLOAK_IMAGE env var.")
 cfg = config.parse()
 
 enabled_apps = cfg.get("enabled_apps", ["mit-learn", "learn-ai", "mitxonline", "odl-video-service"])
@@ -54,6 +55,12 @@ prebuilt_tags = {
 # not pinned in Pulumi.local-dev.core.Dev.yaml, since Pulumi config would win
 # over the environment and silently override this.
 log_retention_period = cfg.get("log_retention_period") or os.environ.get("LOCAL_DEV_LOG_RETENTION", "")
+
+# Keycloak server image. Empty means the Pulumi program's own default (the
+# published mitodl/keycloak digest). Set it to a locally built image to test an
+# unreleased ol-keycloakify theme -- see "Testing a local ol-keycloakify build"
+# in local-dev/README.md. Same not-pinned-in-Pulumi-config caveat as above.
+keycloak_image = cfg.get("keycloak_image") or os.environ.get("LOCAL_DEV_KEYCLOAK_IMAGE", "")
 
 # Workspace root: directory that contains ol-infrastructure and sibling app repos.
 # Override with MITOL_WORKSPACE_ROOT environment variable.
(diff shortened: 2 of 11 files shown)


daniellefrappier18 (approved):
Manually tested this end to end against a real ol-keycloakify checkout 

Verified working:

- kc-theme-image.sh builds, pushes, and writes keycloak_image; Tilt rolls keycloak-0 onto the new tag and the built jar is the only theme jar in providers/.
- `--reset` correctly rolls back to the published image.
- pulumi preview on the core stack: with the knob unset vs. set, the only diff on the Keycloak CR is spec.image — nothing else moves.
- UPDATE_EMAIL end to end: forced the required action on a seeded user, logged in, got the new "Confirm your email change" email in Mailpit, clicked the confirmation link, and the user's email/username were actually updated in Keycloak.

Found and worked around (not a blocker on this PR):

- The account-console client in the seeded olapps realm has webOrigins: [] instead of Keycloak's built-in ["+"], so every browser-based account-console login fails token exchange with {"error":"Invalid origin"}. Reproduced identically on the published image, so it's unrelated to this PR's image-swap mechanism — identity_core.py doesn't touch this client at all. Worth a follow-up fix (webOrigins: ["+"] on that client) since it currently blocks anyone testing account-console flows in local-dev at all.
- Account console UI has some rendering glitches (raw {{givenName}} {{familyName}} placeholders, broken icon glyphs, no visible "Update email" button despite updateEmailFeatureEnabled: true). 

Image

ChristopherChudzicki:
@daniellefrappier18

> Found and worked around (not a blocker on this PR) ...

Thanks. Those are logged as tk-local-dev-keycloak-account-console-fails-token-e-f7fe30 in shared witan.

PR #5862, opened 2026-09-14
feat(local-dev): per-developer kustomize overlays for app manifests

ChristopherChudzicki:
N/A

Lets a developer change anything in an app's local-dev manifests (memory limits, probes, replicas, an image) without editing committed files.

- Tracked manifests move to `local-dev/apps/<app>/base/` with a `kustomization.yaml` listing them (pure renames).
- `local-dev/apps/<app>/local/` is a gitignored kustomize overlay of `base/`, with committed `.example` templates for its two files:
    - `kustomization.yaml` (the overlay; Tilt builds it instead of `base/` whenever it exists, and `patches:` go here)
    - and `app-env.local.yaml` (the existing env-override ConfigMap, moved from `configmaps/`, now listed as a resource of the overlay).
- `pg_memory_limit` in `tilt_config.json` sets the shared Postgres pod's memory limit, forwarded to the core Pulumi stack like `log_retention_period`.

**Why so many files changed?:** kustomize has two rules—(1) one kustomization per directory and (2)  an overlay cannot live inside its own base—which together constrain the layout. We either build a parallel kustomization app tree sibling to `local-dev/apps`, or we move a bunch of files and have `apps/<app>/base + apps/<app>/local`. The latter is nicer: It's what kuztomize's docs use, and it keeps all the per-app stuff in a single dir rather than two.

Committed defaults are unchanged; raising them is a separate question.

<details>
<br>

- `tiltlib.star`: `k8s_yaml_local(paths, local_overrides)` becomes `k8s_yaml_app(app_dir)`: kustomize-build `base/` or `local/`, root-domain substitution, fingerprint, stamp the `ol.mit.edu/config-hash` pod-template annotation. The Starlark that used to read, validate, and merge the env-override file is gone; the ConfigMap reaches the cluster as an overlay resource like every other manifest. Its position in `envFrom` (last, `optional: true`) is unchanged, so overrides still beat both the tracked ConfigMap and the tracked Secret.
- Both `local/` files are required together: kustomize fails if the overlay lists a missing `app-env.local.yaml`, and an `app-env.local.yaml` with no overlay beside it fails the Tiltfile with a pointer, since on its own it would silently do nothing. Copying both templates untouched is a no-op (the template's memory patch ships commented out).
- Tilt's `kustomize()` uses a standalone `kustomize` if present, else `kubectl kustomize`, so no new prerequisite. Tilt watches the inputs its kustomization parser knows about (`resources`, patch files, `crds`, `configMapGenerator` `files`), which covers `app-env.local.yaml`; `configMapGenerator` `envs` and `secretGenerator` sources are not watched, and the README says so.
- The overlay must live outside `base/` (kustomize rejects an overlay nested in its own base), which is why the tracked files moved down a level rather than the overlay moving elsewhere.
- Strategic-merge patches must name the Deployment's `namespace`; the `.example` files carry that and match containers by name.
- Overlay detection uses `read_file(default="")` rather than `os.path.exists` so creating the file mid-session re-evaluates the Tiltfile.
- `.gitignore` covers everything under `local/` except `*.example`, so patch files dropped beside the kustomization are ignored too. The broad `*.local.yaml` pattern stays, so an `app-env.local.yaml` left in the old location is still never tracked.
- Temporary migration, marked `DELETE ANYTIME AFTER 2026-11-14` in both places: `start.sh` runs `local-dev/scripts/migrate-local-overrides.sh`, which prints the `mv`/`cp` it would do and asks before doing it (moves `configmaps/app-env.local.yaml` into `local/`, copies the overlay template beside any env file lacking one; never overwrites). A matching check in `tiltlib.star` fails the Tiltfile on a file still in `configmaps/`, since nothing reads it there and the overrides would otherwise vanish silently. The README does not mention any of this.
</details>

### How can this be tested?

1. Start on `main` with local-dev running.
2. Switch to this branch. You should see an error in the Tilt UI (not the terminal, which does not stream logs by default) on the `(Tiltfile)` resource:
    > Error in fail: <repo>/local-dev/apps/mit-learn/configmaps/app-env.local.yaml is no longer applied; it belongs in <repo>/local-dev/apps/mit-learn/local/. Run local-dev/scripts/start.sh (or local-dev/scripts/migrate-local-overrides.sh) to move it and create the overlay beside it.
3. Stop local-dev. Re-start with `./local-dev/scripts/start.sh`. You should see it prompt for permission akin to below. Say yes.
    ```
    Per-developer override files need to move into each app's local/ overlay:
    mv local-dev/apps/mit-learn-nextjs/configmaps/app-env.local.yaml  →  local-dev/apps/mit-learn-nextjs/local/app-env.local.yaml
    mv local-dev/apps/mit-learn/configmaps/app-env.local.yaml  →  local-dev/apps/mit-learn/local/app-env.local.yaml
    mv local-dev/apps/mitxonline/configmaps/app-env.local.yaml  →  local-dev/apps/mitxonline/local/app-env.local.yaml
    mv local-dev/apps/odl-video-service/configmaps/app-env.local.yaml  →  local-dev/apps/odl-video-service/local/app-env.local.yaml
    cp local-dev/apps/mit-learn-nextjs/local/kustomization.yaml.example  →  local-dev/apps/mit-learn-nextjs/local/kustomization.yaml
    cp local-dev/apps/mit-learn/local/kustomization.yaml.example  →  local-dev/apps/mit-learn/local/kustomization.yaml
    cp local-dev/apps/mitxonline/local/kustomization.yaml.example  →  local-dev/apps/mitxonline/local/kustomization.yaml
    cp local-dev/apps/odl-video-service/local/kustomization.yaml.example  →  local-dev/apps/odl-video-service/local/kustomization.yaml
    ```
4. local-dev should work (after you wait for pods, etc).
5. Env var changes in the `app-env.local.yaml` files should still work: they should roll the pods and be visible inside them.
6. **Check `pg_memory_limit`**:
    - Run `kubectl get pod -n local-infra local-pg-1 -o jsonpath='{.spec.containers[0].resources.limits.memory}'` to check the Postgres memory limit; the default is `512Mi`.
    - Set `"pg_memory_limit": "2Gi"` (or whatever) in your `tilt_config.json` and watch `local-infra-core` re-run in Tilt.
    - The operator restarts `local-pg-1` (a few seconds of Postgres downtime).
    - Afterwards `kubectl get pod -n local-infra local-pg-1 -o jsonpath='{.spec.containers[0].resources.limits.memory}'` prints the new value.

🤖 Co-authored with [Claude Code](https://claude.com/claude-code)


ChristopherChudzicki:
@blarghmatey To your slack comment about alternatives,
> Kustomize is definitely the more widely adopted approach. The alternative approach that I would be thinking of is **whether/how we can make Pulumi own more of that stack so that we don't introduce a lot of drift between local-dev and our deployed environments**. There's already a lot of variance between them, so I just wanted to think through that piece.

(Emphasis mine). Yes, lots of things are already pinned in two places and it's unnerving.

With that in mind, I took a look at the unification with Fable. Our summary, below.

### Make Local-dev parallel RC/Prod

Making `OLApplicationK8s` local-capable is a refactor of code that deploys eleven apps across CI/RC/Production:
- make all EKS-specific stuff optional (e.g. the ECR registry rewrite, the AWS security group policy and its required inputs, Vault-sourced secrets, the VPA/KEDA/HPA/PodMonitor objects).
- Then a local-dev Pulumi program instantiates it per app with DockerHub images and plain Secrets. Env vars still come local-dev.

How it would hook into Tilt:
- **Option 1**: `k8s_custom_deploy` with `pulumi up` as the apply command.
    - **Same apply path as rc/prod** (same Pulumi k8s provider, diffing, errors), not just the same object definitions. Stable, non-beta. Needs some glue. [^glue]
    - **Code edits stay fast**: live update syncs into the running container as today. What pays a `pulumi up` (tens of seconds, unmeasured per app) is image rebuilds, env/manifest changes, and each app at `tilt up`.
    - **Per-dev overrides**: one gitignored file per app that the local Pulumi program reads. The program merges env-var overrides into the env dict it hands the component (last-wins, as today) and passes memory, replicas, probes, extra sidecars/volumes to existing component fields, so no component changes for the common cases; a field the component doesn't expose would need a Pulumi transformation. Same knobs as prod, one place for overrides; an env change costs a `pulumi up` rather than a `kubectl apply`. The kustomize overlay goes away.
- **Option 2**: Pulumi's `renderYamlToDirectory` provider mode. Pulumi renders manifests to a directory and only re-runs when the program changes.
    - **Fast and Tilt-native**: Tilt applies the rendered YAML itself, so image injection, live update, and redeploys work exactly as today.
    - **Pulumi labels this feature BETA** / developer preview. Its other caveats, undefined cluster-computed outputs and plaintext secrets in the output, are non-issues for a local render.
    - **Apply path differs from rc/prod** (`kubectl apply` via Tilt, not the Pulumi provider).
    - **Per-dev overrides**: the kustomize overlay from this PR, unchanged. `base/` becomes Pulumi's rendered output, and the env-override ConfigMap and patches sit on top of it exactly as they do on hand-written YAML.

[^glue]: the apply command must print the applied YAML and be idempotent. Pulumi is idempotent but doesn't print YAML, so a small wrapper dumps the objects after `pulumi up`.

blarghmatey (approved):

ChristopherChudzicki:
Follow-up: the Pulumi-unification idea above is now tracked in witan as project `wp-pulumi-owned-local-dev-app-manifests-olapplicati-bde4c5` (two tasks: make `OLApplicationK8s` local-capable, then use it for local-dev via Tilt). Option 1 is the working preference there, not a settled decision; whoever picks it up should check the choice with devops first.

Closing this PR rather than merging. Under Option 1 the kustomize overlay and the `base/` vs `local/` split go away, so the layout change here isn't worth landing. The memory-limit problem that motivated it is being handled separately by running pytest on the host against the cluster databases.

🤖 Co-authored with [Claude Code](https://claude.com/claude-code)

PR #5876, opened 2026-09-15
fix(starrocks): give Pulumi the operator CRD so upgrades actually apply

Ardiea:
N/A — no GitHub issue. Tracked in witan as `tk-starrocks-crd-is-frozen-at-v1-9-10-while-the-ope-6d56d2`.

Helm's `crds/` directory is install-only: `helm upgrade` never touches it. The StarRocks operator chart ships its CRD there, so the CRD froze at whatever chart version first landed on each cluster while the operator Deployment was upgraded repeatedly.

- All three data clusters run operator 1.11.7 against a CRD annotated `version: v1.9.10`.
- The live schema is **37 properties behind** chart 1.11.7 on `data-qa` / `data-production`, and **76 behind** on `data-ci`.
- Pruned fields include `persistentVolumeClaimRetentionPolicy`, `minReadySeconds`, `podManagementPolicy`, `sysctls`, `storageVolumes[].csi`, `service.externalTrafficPolicy` and (on `data-ci`) the whole `disasterRecovery` block.
- Nothing in the repo sets any of them today, so there is no active breakage — but setting one is silently useless: Pulumi accepts it, the API server accepts it, and schema pruning drops it with no error anywhere.

Pulumi now applies the CRD as its own resource ahead of the release, and `skip_crds=True` keeps Helm out of it on fresh installs too.

This is what blocks `persistentVolumeClaimRetentionPolicy` on the StarRocks FE StatefulSet — only the stale schema was in the way. The running operator already honours it: [`pkg/k8sutils/templates/statefulset/spec.go#L99-L113`](https://github.com/StarRocks/starrocks-kubernetes-operator/blob/v1.11.7/pkg/k8sutils/templates/statefulset/spec.go#L99-L113) forces `WhenScaled: ""` and passes `WhenDeleted` through.

<details>
<br>

**Adoption needs two flags, not one.** The CRD already exists on every cluster and its fields are owned by Helm's field manager, so handing it to Pulumi fails twice before it works. Both were confirmed necessary by reproducing the frozen CRD on a throwaway k3d cluster (install chart 1.9.10 → `helm upgrade` to 1.11.7 → CRD still `v1.9.10`) and then adopting it:

1. `upsert_existing_objects=True` on the provider. Without it: `customresourcedefinitions.apiextensions.k8s.io "starrocksclusters.starrocks.com" already exists`.
2. `pulumi.com/patchForce: "true"` in the CRD's annotations. Without it: `Apply failed with 2 conflicts: conflicts with "helm" using apiextensions.k8s.io/v1`.

`upsert_existing_objects` is set on a **CRD-scoped provider**, not the cluster-wide one, where it would let any resource silently take over an unrelated pre-existing object that Pulumi never created. It requires pulumi-kubernetes >= 4.29.0; this repo locks 4.34.0.

**Why `deploy/` and not the chart's `crds/` path.** The chart's `crds/starrocks.com_starrocksclusters.yaml` is a **symlink** to `deploy/starrocks.com_starrocksclusters.yaml`, and raw.githubusercontent.com serves a symlink's target *path* as the response body instead of following it — fetching the chart path yields a one-line manifest, not a CRD. The `deploy/` manifest at tag `v1.11.7` parses equal to the `crds/` manifest in the packaged chart (verified by pulling the chart and comparing the parsed YAML).

**Version pinning.** The URL is derived from `STARROCKS_OPERATOR_CHART_VERSION`, whose git tag is the chart version prefixed with `v`. [`versions_map.py`](https://github.com/mitodl/ol-infrastructure/blob/main/src/ol_concourse/pipelines/versions_map.py#L159-L171) already tracks that constant for `substructure/aws/eks/`, so a renovate bump redeploys this stack and carries the CRD along. No new version pin.

**Audit of the other Helm-installed operators.** All 31 Helm charts in the repo were pulled at their pinned versions and checked for a top-level `crds/` directory, then every resulting CRD was schema-diffed against all 12 live EKS clusters.

Clean, as their mechanisms predict:
- **ClickHouse (Altinity)** — applied via `yaml.v2.ConfigGroup` from the raw install bundle; exact match to `release-0.26.0` on all 3 data clusters. (Opik bundles its own altinity subchart, but it is `enabled: False`.)
- **Keycloak** — `ConfigGroup` from raw manifests; exact match to 26.7.3 on all 3 operations clusters.
- **typesense** — ships its CRD in `templates/`, not `crds/`, so Helm re-renders it every upgrade; exact match to 0.4.1.
- **KEDA / VPA** — same, `templates/crds/` is templated and upgrades normally.
- **karpenter / toolhive** — already dodge this via dedicated `-crds` charts.

Drifting, filed as follow-ups rather than fixed here since each has a different blast radius:
- `tk-traefik-crds-frozen-by-helm-s-install-only-crds--4dfd71` — 25 CRDs, all 12 clusters; `middlewares.traefik.io` is 50 properties behind and `Middleware` **is** used by `components/services/traefik.py`.
- `tk-apisix-ingress-controller-crds-frozen-by-helm-s--40a5b3` — `apisixupstreams` still serves `v2beta3`, which chart 2.17.0 no longer ships. The only served-version discrepancy in the whole audit, so not a straight copy of this fix.
- `tk-aws-load-balancer-controller-crds-frozen-by-helm-300fd5` — `targetgroupbindings` missing `spec.targetGroupProtocol` and the entire `status.conditions[]` block, which the controller itself writes.
- `tk-orphaned-starrocksclusters-crd-on-applications-c-efd6f0` — `applications-ci` carries a StarRocks CRD though the operator is only enabled on `data-*`.
</details>

### How can this be tested?
What I ran:

- **Reproduced the bug on a throwaway k3d cluster**: `helm install` operator chart 1.9.10, then `helm upgrade` to 1.11.7. Release reports `operator-1.11.7`; CRD still annotated `v1.9.10`; `starRocksFeSpec.persistentVolumeClaimRetentionPolicy` absent.
- **Applied this mechanism to that cluster**: CRD annotation moved `v1.9.10` → `v1.11.7`, and `persistentVolumeClaimRetentionPolicy`, `minReadySeconds`, `podManagementPolicy` and `sysctls` are present on all of `starRocksFeSpec` / `starRocksBeSpec` / `starRocksCnSpec`. A full schema-property diff against chart 1.11.7 came back with zero differences.
- **Idempotency**: a second `pulumi up` reported `5 unchanged`.
- **Helm non-interference**: a subsequent `helm upgrade --skip-crds` left the CRD at `v1.11.7`.
- **`pulumi preview`** on `data.CI`, `data.QA` and `data.Production`: all three clean, `+3 to create` (CRD-scoped provider, ConfigGroup, CRD) and `~2 to update` — the Helm release flipping `skipCrds: false => true`, plus an unrelated pre-existing `EC2NodeClass` AMI-alias drift (`al2023@v20260903 => al2023@v20260911`) that is not from this change.
- `pre-commit run` passes on the changed file (ruff format, ruff check, mypy).

Not yet applied to the real clusters — that happens on the next `substructure/aws/eks` Concourse deploy after merge.

To validate after deploy, on each of `data-ci` / `data-qa` / `data-production`:

```
kubectl --context data-ci get crd starrocksclusters.starrocks.com \
  -o jsonpath='{.metadata.annotations.version}'
# expect: v1.11.7

kubectl --context data-ci get crd starrocksclusters.starrocks.com -o json \
  | jq '.spec.versions[0].schema.openAPIV3Schema.properties.spec
        .properties.starRocksFeSpec.properties
        | has("persistentVolumeClaimRetentionPolicy")'
# expect: true
```

### Additional Context
The first apply on each cluster force-takes the CRD's schema fields from Helm's field manager. That is the intended effect and it is why `patchForce` is here, but it does mean the first deploy is the moment the schema actually changes — worth watching the `data-ci` deploy before QA/Production. The change is additive to the schema (no served apiVersion is removed and no property is dropped), so existing `StarRocksCluster` objects are unaffected.

`skip_crds=True` means a brand-new cluster now depends on the `ConfigGroup` for the CRD rather than Helm; `depends_on` orders it ahead of the release so the operator never starts against a missing CRD.


blarghmatey (approved):

Ardiea:
### Out-of-band cleanup: removed an orphaned StarRocks CRD from `applications-ci`

Recording this here because it is a manual `kubectl` action with no code change, and it should not read as unexplained drift later.

The CRD audit behind this PR turned up a **fourth** cluster carrying `starrocksclusters.starrocks.com` — `applications-ci` — even though the operator is only ever installed on the `data` clusters. `setup_starrocks` is gated on `stack_info.env_prefix == "data"` (`src/ol_infrastructure/substructure/aws/eks/__main__.py:384`) and only the three `Pulumi.data.*.yaml` stacks set `starrocks:enable_operator: true`, so nothing in this repo installs or reconciles that CRD on `applications-ci`.

Verified before deleting:

| check | result |
| --- | --- |
| `kubectl --context applications-ci get starrockscluster -A` | `No resources found` |
| `starrocks` / `kube-starrocks` namespace | absent |
| any Deployment matching `starrocks` | none |
| helm release owning it | none (no helm ownership labels or `meta.helm.sh/*` annotations on the CRD) |
| `pulumi stack --stack applications.CI export` (990 KB) | zero `starrocks` matches |

CRD metadata: created `2026-02-02`, annotated `version: v1.11.4`, no labels. A leftover from a one-off install.

Deleted:

```
$ kubectl --context applications-ci delete crd starrocksclusters.starrocks.com
customresourcedefinition.apiextensions.k8s.io "starrocksclusters.starrocks.com" deleted
```

The three data clusters are untouched and still carry the CRD this PR takes ownership of:

```
data-ci:         starrocksclusters.starrocks.com  2025-01-28T22:41:24Z
data-qa:         starrocksclusters.starrocks.com  2026-01-20T17:10:33Z
data-production: starrocksclusters.starrocks.com  2026-02-09T15:08:20Z
```

The motivation is the trap rather than the cost: an unreferenced CRD is free at runtime, but with no controller behind it a `StarRocksCluster` created on `applications-ci` would be accepted by the API server and then sit there forever with nothing reconciling it.

**While there — swept all 12 clusters for the same pattern** (CRDs installed for an operator the cluster does not run), by diffing the CRD set within each `ci`/`qa`/`production` family:

- `globalaccelerators.aga.k8s.aws`, `listenerruleconfigurations.gateway.k8s.aws`, `albtargetcontrolconfigs.elbv2.k8s.aws` — `aws-load-balancer-controller` chart-version skew across envs. The controller is running everywhere; these resolve on the next chart bump. Not orphans.
- `apisixclusterconfigs.apisix.apache.org` on `operations-ci` only — looked like a candidate, but the legacy `apache-apisix-ingress-controller` Deployment is running (3/3) on all three operations clusters and the legacy CRs are in active use (`apisixroutes`, `apisixtlses`, `apisixpluginconfigs`). This is the same install-only `crds/` freeze this PR fixes for StarRocks, not an absent operator.

`starrocksclusters` on `applications-ci` was the only genuine orphan.

PR #5882, opened 2026-09-15
feat(local-dev): add ocw-studio with RustFS object storage and Concourse

gumaerc:

Closes mitodl/hq#13381 (together with mitodl/ocw-studio#3207)

Depends on mitodl/ocw-studio#3207, which makes the compose-era addresses configurable. Without it the app cannot reach the object store.

Adds ocw-studio to local-dev, opt-in and absent from the default `enabled_apps` because it brings an object store and a privileged Concourse worker with it.

`enabled_apps` previously only reached Tilt: the Pulumi stacks provisioned every app's namespace, database and OIDC client regardless, which the README said outright. The Tiltfile now forwards it as `LOCAL_DEV_ENABLED_APPS`, and namespaces listed in the new `OPTIONAL_APP_NAMESPACES` are created only when their app is on. The four original apps stay unconditional, so nobody's existing stack churns.

RustFS is the S3 stand-in, deployed only when an app that needs it is enabled. Minio's community server is effectively frozen. One thing worth knowing: RustFS does not implement S3 ACLs, so the bootstrap Job grants anonymous read with a bucket policy. Published sites are served with no credentials, so without that every page would 403.

Concourse runs web plus worker in one pod. The TSA authenticates workers with a keypair both halves must agree on, and a single pod gives a shared emptyDir and a loopback TSA connection without needing RBAC to mint a Secret. The worker is the only privileged workload here; it runs its own containerd inside the k3s node container.

Five hostnames: studio, plus draft/live/test for the published sites and the Concourse UI.

### How can this be tested?

Add `ocw-studio` to `enabled_apps` in `tilt_config.json`, then re-run `./local-dev/scripts/setup.sh` for the hostnames and the `*.ocw` cert wildcard. Publishing needs a git host you can push to, so set `GIT_ORGANIZATION` and `GIT_TOKEN` in the gitignored `app-env.local.yaml`, or set `CONTENT_SYNC_PIPELINE_BACKEND` empty to edit content without publishing.

Verified end to end against a copy of the production database (2904 sites):

- Log in through Keycloak, publish a site from the CMS.
- The pipeline pushes to GitHub, Concourse builds it, and the result serves from the live host.
- The theme assets pipeline builds and uploads (212 objects per bucket).
- `pulumi preview` on the core stack with no optional app enabled is a no-op, so existing stacks are unaffected.

### Additional Context

Two things sized from that run rather than guessed. RustFS idles around 850Mi, so the original 1Gi limit OOMKilled it under a mass build and Hugo's instructor fetches timed out. And every parallel build runs inside the one worker pod, so its limit has to match `OCW_MASS_BUILD_MAX_IN_FLIGHT`.

The mit-learn cookie change is included because it is the same bug surfacing elsewhere: its CSRF and session cookies were scoped to the whole root domain under Django's default names, so every sibling app received them and the duplicate broke ocw-studio's CSRF checks intermittently. Narrowed to `.learn`, which still covers its own hosts.


local-dev/infra/core/__main__.py
```diff
@@ -78,6 +79,20 @@
     or "168h"
 )
 
+# Which apps the developer has switched on, forwarded by the Tiltfile from
+# tilt_config.json's enabled_apps. Only the app-specific resources added since
+# per-app gating existed consult this -- the four original apps' namespaces and
+# databases are still provisioned unconditionally, so an existing stack does
+# not churn when this is set. Same not-pinned-in-Pulumi-config caveat as
+# log_retention_period below.
+enabled_apps = tuple(
```
umar8hassan: `enabled_apps` falls back to an empty list when `LOCAL_DEV_ENABLED_APPS` is unset, and the optional namespaces and the object store are then left out of the program. README lines 606-607 tell people to run `pulumi up --stack local-dev.core.Dev` by hand without that variable, so following those instructions deletes the `ocw-studio` namespace and the RustFS store along with their contents. Either stop when the variable is missing, or read the list from Pulumi config so a hand-run update keeps it.

local-dev/infra/modules/objectstore.py
```diff
@@ -0,0 +1,359 @@
...
+
+# Local-dev credentials. These are not secret in any meaningful sense: the
+# whole cluster is local and every app manifest carries the same literal pair.
+# They are duplicated in local-dev/apps/ocw-studio/secrets.yaml.
+ACCESS_KEY = "localdevaccess"
+SECRET_KEY = "localdevsecret123"  # noqa: S105  # pragma: allowlist secret
+
+# Buckets created at bootstrap, mirroring ocw-studio's .env.example names so a
+# developer's existing local settings keep working. Every one is granted
+# anonymous read; ocw-studio's published output is public content.
```
umar8hassan: Every bucket here is given public read, including `ol-ocw-studio-app-local`, `ol-ocw-studio-app-import-local` and `ol-eng-artifacts-local`. The reasoning in this comment covers published course pages, but those three hold uploads, imports and build output rather than published pages. I checked it against a running store: an anonymous request for a missing key in `ol-ocw-studio-app-local` comes back `NoSuchKey`, not `AccessDenied`, so anything in it can be read without credentials. Limiting `PUBLIC_BUCKETS` to the `ocw-content-*` buckets would match both the comment and the deployed setup.

local-dev/infra/modules/objectstore.py
```diff
@@ -0,0 +1,359 @@
...
+        "rustfs-svc",
+        metadata={"name": "rustfs", "namespace": "local-infra"},
+        spec={
+            "selector": {"app": "rustfs"},
+            "ports": [{"name": "s3", "port": 9000, "targetPort": 9000}],
+        },
+        opts=_k8s(parent=stateful_set),
+    )
+
+    bootstrap_job = k8s.batch.v1.Job(
```
umar8hassan: The Job mounts `rustfs-bootstrap-script`, but nothing orders it after that ConfigMap: `parent=` does not create an ordering edge, and the ConfigMap is declared further down. On a first run the Job's pod can start before the ConfigMap exists and then sits unable to start its container, which holds up the update until it times out and leaves the buckets and their policies uncreated. Add the ConfigMap to `depends_on` next to the service.

local-dev/infra/modules/objectstore.py
```diff
@@ -0,0 +1,359 @@
...
+        metadata={
+            "name": "rustfs-bootstrap-buckets",
+            "namespace": "local-infra",
+        },
+        spec={
+            # The Job body is idempotent but its spec is immutable, so bumping
+            # the bucket list means Pulumi replaces the Job rather than
+            # patching it.
+            "backoffLimit": 6,
+            "ttlSecondsAfterFinished": 600,
```
umar8hassan: `ttlSecondsAfterFinished` removes the Job ten minutes after it finishes while Pulumi still has it in state. Tilt refreshes first so this stays hidden there, but the hand-run command in README lines 606-607 does not, and a later change touching the Job then acts on something that is no longer there. Drop the TTL, or mark the Job so it is replaced on each update.

local-dev/apps/ocw-studio/deployment.yaml
```diff
@@ -0,0 +1,399 @@
...
+          until $PSQL "SELECT 1" >/dev/null 2>&1; do
+            echo "waiting for postgres..."
+            sleep 2
+          done
+          for db in ocwstudio concourse; do
+            if [ "$($PSQL "SELECT 1 FROM pg_database WHERE datname='${db}'")" = "1" ]; then
+              echo "database exists: ${db}"
+            else
+              echo "creating database: ${db}"
+              $PSQL "CREATE DATABASE ${db} OWNER app"
```
umar8hassan: This `CREATE DATABASE` has no `|| true`, unlike the same step in `concourse.yaml` line 87, and the script runs under `set -eu`. When this pod and the Concourse pod start together on a fresh enable, both can check for `concourse` before either creates it, and the one that loses exits non-zero and puts the web pod into a restart loop. Use the same guard as `concourse.yaml`.

local-dev/apps/ocw-studio/Tiltfile
```diff
@@ -0,0 +1,92 @@
...
+k8s_resource('ocwstudio-celery', labels=['ocw-studio'], resource_deps=['local-infra-apps'])
+
+# objects= is explicit on purpose. Left to its own devices Tilt groups every
+# ConfigMap in the applied YAML onto the first workload that references one,
+# which put the site-host nginx config behind the Studio image build -- so the
+# published sites stayed down until an unrelated build finished. Naming them
+# here keeps the two lifecycles genuinely independent.
+k8s_resource(
+    'ocwstudio-sitehost',
+    objects=[
```
umar8hassan: `ocwstudio-site-plugins`, `ocwstudio-shared-plugins` and `ocwstudio-tls` are not named in any `objects=` list, so Tilt assigns them to `ocwstudio-webapp`. The routes for the published sites and for Concourse point at those plugin configs and that certificate, so none of them can be served until the webapp image has finished building, which is the wait this comment says was removed. Add them to the `ocwstudio-sitehost` and `ocwstudio-concourse` lists.

local-dev/apps/mit-learn/ocw-hugo-dev.yaml
```diff
@@ -0,0 +1,102 @@
...
+addressType: IPv4
+ports:
+- name: http
+  port: 3000
+  protocol: TCP
+endpoints:
+# Substituted by tiltlib.star with the k3d bridge gateway, which is how the
+# cluster reaches the host. Set LOCAL_DEV_HOST_GATEWAY to override.
+- addresses:
+  - "__LOCAL_DEV_HOST_GATEWAY__"
```
umar8hassan: This is the Docker bridge gateway, which is the developer's machine only on Linux. On Docker Desktop for Mac it points inside the Docker virtual machine, so the dev server running on the machine is unreachable and the page returns 502 — the same result the comment attributes to the server being stopped. README lines 542-544 already give the rule: `host.docker.internal` on Docker Desktop, the bridge address only on Linux.

local-dev/apps/mit-learn/ocw-hugo-dev.yaml
```diff
@@ -0,0 +1,102 @@
...
+---
+apiVersion: apisix.apache.org/v2
+kind: ApisixRoute
+metadata:
+  name: ocw-hugo-dev-route
+  namespace: mit-learn
+spec:
+  ingressClassName: apache-apisix
+  http:
+  - name: hugo-dev
```
umar8hassan: The route does not set `websocket: true`, so the live-reload connection is never upgraded and theme edits need a manual refresh. `learn-ai/apisix-routes.yaml` line 95 and `mit-learn-nextjs/apisix-routes.yaml` line 39 both set it.

local-dev/tiltlib.star
```diff
@@ -22,6 +26,24 @@
 
 _ROOT_DOMAIN_DEFAULT = "mit.dev"
 _CONFIG_HASH_ANNOTATION = "ol.mit.edu/config-hash"
+_HOST_GATEWAY_TOKEN = "__LOCAL_DEV_HOST_GATEWAY__"
+_CLUSTER_NAME_DEFAULT = "local-dev"
+
+def _host_gateway():
```
umar8hassan: This runs `docker network inspect` on every Tiltfile load and uses the result unchecked. The template loops over each address block, so a network with both IPv4 and IPv6 returns two gateways run together and a network without one returns nothing; either value goes straight into `addresses` and the manifest is refused. `local()` also stops the whole Tiltfile when the network is missing, which breaks `tilt up` for people not using the dev server. Take the first address only, and skip this resource when the lookup gives nothing usable.

local-dev/scripts/seed.sh
```diff
@@ -55,6 +57,14 @@ python manage.py configure_instance"
 
     [odl-video-service]="python manage.py migrate --noinput
 python manage.py createpresets"
+
+    # No migrate: the ocw-studio deployment runs it in an init container.
+    # The superuser is what a Keycloak login associates to by email, so
+    # without it the first login lands as an unprivileged new user.
+    [ocw-studio]="python manage.py createsuperuser --noinput --username admin --email admin@odl.local --name Admin
```
umar8hassan: `createsuperuser` fails once the account exists, and this script runs under `set -euo pipefail`, so a second `seed.sh --app ocw-studio` stops here and never reaches `backpopulate_groups`, `import_website_starters` or `upsert_theme_assets_pipeline`. Every other app's list is made up of commands that can be run again safely. Add `|| true`, or move the superuser step last.

local-dev/apps/ocw-studio/concourse.yaml
```diff
@@ -0,0 +1,243 @@
...
+      labels:
+        app: ocwstudio
+        component: concourse
+    spec:
+      initContainers:
+        # Same keys.sh the compose stack ran, inlined. Regenerated on every
+        # pod start: they authenticate the worker to the web in this pod only,
+        # and nothing outside it ever sees them.
+      - name: concourse-keys
+        image: gotechnies/alpine-ssh:latest
```
umar8hassan: `gotechnies/alpine-ssh:latest` is the only image in this PR without a fixed version, and it is the one that makes the key pair the worker uses to reach the web node. A change to that tag upstream changes this behaviour with nothing changing here, and two people setting up on different days can get different results. `ssh-keygen` is already available in images pinned elsewhere in this pod.

local-dev/apps/ocw-studio/concourse.yaml
```diff
@@ -0,0 +1,243 @@
...
+          mountPath: /worker-state
+        # Sized for the configured build concurrency, not for an idle
+        # worker. Every pipeline task runs in this one pod's containerd, so
+        # OCW_MASS_BUILD_MAX_IN_FLIGHT parallel Hugo builds all draw on this
+        # single limit -- at the default of 40, a few hundred MB each adds up
+        # well past a handful of GB. The host this targets has far more to
+        # spare than the old 4Gi cap assumed, and an under-sized cap here shows
+        # up as builds failing on template timeouts rather than as an obvious
+        # OOM, so it is worth being generous.
+        resources:
```
umar8hassan: The comment ties this limit to the build concurrency, but `OCW_MASS_BUILD_MAX_IN_FLIGHT` is set in `configmaps/app-env.yaml` and can be overridden per developer, so raising it leaves this at 24Gi. README line 71 also asks for a 12-16 GB Docker virtual machine, so the limit cannot be reached and the worker is killed by the machine first; I hit this on a 15 GB machine. Derive both from one value, or say what ceiling the limit assumes.

local-dev/apps/ocw-studio/concourse.yaml
```diff
@@ -0,0 +1,243 @@
...
+          requests:
+            cpu: "2"
+            memory: 2Gi
+          limits:
+            memory: 24Gi
+
+      volumes:
+      - name: concourse-keys
+        emptyDir: {}
+      - name: worker-state
```
umar8hassan: `worker-state` has no `sizeLimit`, and with the `naive` volume driver each parallel build copies its inputs here instead of sharing them. At 40 builds that is tens of gigabytes on the node's own disk, and passing the eviction threshold set in `k3d-config.yaml` removes pods across the whole cluster, not only Concourse. Set a `sizeLimit` in line with the build concurrency.

local-dev/apps/mit-learn/configmaps/app-env.yaml
```diff
@@ -37,11 +37,28 @@ data:
...
+  # logout POSTs failed CSRF intermittently because of exactly that.
+  #
+  # .learn still covers both of mit-learn's own hosts, learn and api.learn
+  # (and ai.learn, as before), which are also its only CSRF_TRUSTED_ORIGINS.
+  # Nothing needs these at mitxonline: that integration is served under
+  # api.learn.mit.dev/mitxonline/* and keeps its session on the learn host.
+  #
+  # MITOL_COOKIE_DOMAIN below is deliberately left alone -- it carries a
+  # distinctly named cookie (mitlearn-auth), so it collides with nothing.
+  CSRF_COOKIE_DOMAIN: ".learn.mit.dev"
```
umar8hassan: Narrowing the domain moves this rather than settling it: `learn-ai` serves from `ai.learn.<domain>`, which is inside the new scope, and it sets no cookie name or domain of its own, so a browser there still gets two `csrftoken` cookies. Setting `CSRF_COOKIE_NAME` and `SESSION_COOKIE_NAME` gives each app its own cookie, which is what `MITOL_COOKIE_NAME` already does for the cookie this comment says collides with nothing.

local-dev/infra/modules/keycloak.py
```diff
@@ -780,3 +799,88 @@ def _make_oidc_secret(
...
+            realm_id=realm.realm,
+            client_id=ocw_studio_client.id,
+            name="audience",
+            included_client_audience=ocw_studio_client.client_id,
+            add_to_id_token=True,
+            add_to_access_token=True,
+            opts=kc_opts,
+        )
+
+        realm_keys = keycloak.get_realm_keys_output(
```
umar8hassan: `realm_keys` is already fetched at line 633 and read at line 645. This repeats the same lookup, and passing it through `Output.all()` turns the typed keys into plain dictionaries, which is the only reason `_first_public_key` exists. Reusing the existing value and its `.apply` removes both the second lookup and the helper. `_first_public_key` also returns an empty string when the list is empty, which would write an empty public key into the secret and fail login with nothing pointing at the cause.

local-dev/apps/ocw-studio/configmaps/nginx.yaml
```diff
@@ -0,0 +1,183 @@
...
+# policy, applied by the RustFS bootstrap Job.
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: ocwstudio-sitehost-conf
+  namespace: ocw-studio
+data:
+  default.conf: |
+    server {
+        listen 8044 default_server;
```
umar8hassan: These three server blocks are the same apart from the port and the bucket name, so any later fix has to be made three times. A `map` from `$server_port` to the bucket name, with one server block listening on all three ports, does the same job.

local-dev/EXTENDING.md
```diff
@@ -35,7 +35,14 @@ In `local-dev/infra/modules/database.py`, add to the `postInitSQL` list:
...
+
+Two things not to do:
+
+- Don't add the app's database to the CNPG cluster's `postInitSQL` (step 2) if the app is optional. That block only runs when initdb bootstraps an empty data directory, so it does nothing for anyone who already has a cluster, while still producing a spec diff Pulumi will try to apply to an immutable `initdb` section. Create the database from an init container instead — `local-dev/apps/ocw-studio/deployment.yaml` shows the idempotent pattern.
+- Don't retrofit gating onto the four original apps. Their namespaces and databases are provisioned unconditionally on purpose: changing that would churn Pulumi state on every existing developer's cluster to save a namespace and an empty database.
 
 ### 4. Add the Keycloak OIDC client (if needed)
```
umar8hassan: Step 4 says to call `_make_oidc_secret()`, but the ocw-studio client added here writes its own secret instead, because that helper produces the keys APISIX reads and this app needs `SOCIAL_AUTH_KEYCLOAK_*`. I compared the two on a running cluster and they have no keys in common. Since this PR already revises steps 2 and 3 for the optional-app pattern, step 4 should say which of the two shapes to use and when.

local-dev/infra/core/__main__.py
umar8hassan, line 88: `enabled_apps` falls back to an empty list when `LOCAL_DEV_ENABLED_APPS` is unset, and the optional namespaces and the object store are then left out of the program. README lines 606-607 tell people to run `pulumi up --stack local-dev.core.Dev` by hand without that variable, so following those instructions deletes the `ocw-studio` namespace and the RustFS store along with their contents.

Either stop when the variable is missing, or read the list from Pulumi config so a hand-run update keeps it.

local-dev/infra/modules/objectstore.py
umar8hassan: Every bucket here is given public read, including `ol-ocw-studio-app-local`, `ol-ocw-studio-app-import-local` and `ol-eng-artifacts-local`. The reasoning in this comment covers published course pages, but those three hold uploads, imports and build output rather than published pages. I checked it against a running store: an anonymous request for a missing key in `ol-ocw-studio-app-local` comes back `NoSuchKey`, not `AccessDenied`, so anything in it can be read without credentials. Limiting `PUBLIC_BUCKETS` to the `ocw-content-*` buckets would match both the comment and the deployed setup.

local-dev/infra/modules/objectstore.py
umar8hassan: The Job mounts `rustfs-bootstrap-script`, but nothing orders it after that ConfigMap: `parent=` does not create an ordering edge, and the ConfigMap is declared further down. On a first run the Job's pod can start before the ConfigMap exists and then sits unable to start its container, which holds up the update until it times out and leaves the buckets and their policies uncreated. Add the ConfigMap to `depends_on` next to the service.

local-dev/infra/modules/objectstore.py
umar8hassan: `ttlSecondsAfterFinished` removes the Job ten minutes after it finishes while Pulumi still has it in state. Tilt refreshes first so this stays hidden there, but the hand-run command in README lines 606-607 does not, and a later change touching the Job then acts on something that is no longer there. Drop the TTL, or mark the Job so it is replaced on each update.

local-dev/apps/ocw-studio/deployment.yaml
umar8hassan, line 63: This `CREATE DATABASE` has no `|| true`, unlike the same step in `concourse.yaml` line 87, and the script runs under `set -eu`. When this pod and the Concourse pod start together on a fresh enable, both can check for `concourse` before either creates it, and the one that loses exits non-zero and puts the web pod into a restart loop. Use the same guard as `concourse.yaml`.

local-dev/apps/ocw-studio/Tiltfile
umar8hassan: `ocwstudio-site-plugins`, `ocwstudio-shared-plugins` and `ocwstudio-tls` are not named in any `objects=` list, so Tilt assigns them to `ocwstudio-webapp`. The routes for the published sites and for Concourse point at those plugin configs and that certificate, so none of them can be served until the webapp image has finished building, which is the wait this comment says was removed. Add them to the `ocwstudio-sitehost` and `ocwstudio-concourse` lists.

local-dev/apps/mit-learn/ocw-hugo-dev.yaml
umar8hassan, line 59: I ran this on Docker Desktop for Mac: with a listener running on the machine, a pod reaches it through `host.k3d.internal` and through `host.docker.internal`, but a connection to `172.18.0.1` is refused because that address is the Docker virtual machine's own gateway and not the machine. `https://ocw-dev.learn.<domain>` returns 502 there.

`host.k3d.internal` is the better of the two: k3d injects it into CoreDNS on every runtime, Linux included, so one manifest works everywhere and `_host_gateway()` in `tiltlib.star` is no longer needed.

local-dev/apps/mit-learn/ocw-hugo-dev.yaml
umar8hassan, line 84: The route does not set `websocket: true`, so the live-reload connection is never upgraded and theme edits need a manual refresh. `learn-ai/apisix-routes.yaml` line 95 and `mit-learn-nextjs/apisix-routes.yaml` line 39 both set it.

local-dev/tiltlib.star
umar8hassan: This runs `docker network inspect` on every Tiltfile load and uses the result unchecked. The template loops over each address block, so a network with both IPv4 and IPv6 returns two gateways run together and a network without one returns nothing; either value goes straight into `addresses` and the manifest is refused. `local()` also stops the whole Tiltfile when the network is missing, which breaks `tilt up` for people not using the dev server. Take the first address only, and skip this resource when the lookup gives nothing usable.

local-dev/scripts/seed.sh
umar8hassan, line 64: I actually hit that on a clean setup. I opened Studio, signed in, and the seed could then neither create the account nor raise the one that now existed, which is the outcome the comment above warns about, with nothing in the tooling to get out of it. I had to exec and promote my user. 

It'd be great if we could force-promote the admin user if found.

local-dev/apps/ocw-studio/concourse.yaml
umar8hassan: `gotechnies/alpine-ssh:latest` is the only image in this PR without a fixed version, and it is the one that makes the key pair the worker uses to reach the web node. A change to that tag upstream changes this behaviour with nothing changing here, and two people setting up on different days can get different results. `ssh-keygen` is already available in images pinned elsewhere in this pod.

local-dev/apps/ocw-studio/concourse.yaml
umar8hassan: The comment ties this limit to the build concurrency, but `OCW_MASS_BUILD_MAX_IN_FLIGHT` is set in `configmaps/app-env.yaml` and can be overridden per developer, so raising it leaves this at 24Gi. README line 71 also asks for a 12-16 GB Docker virtual machine, so the limit cannot be reached and the worker is killed by the machine first; I hit this on a 15 GB machine. Derive both from one value, or say what ceiling the limit assumes.

local-dev/apps/ocw-studio/concourse.yaml
umar8hassan: `worker-state` has no `sizeLimit`, and with the `naive` volume driver each parallel build copies its inputs here instead of sharing them. At 40 builds that is tens of gigabytes on the node's own disk, and passing the eviction threshold set in `k3d-config.yaml` removes pods across the whole cluster, not only Concourse. Set a `sizeLimit` in line with the build concurrency.

local-dev/apps/mit-learn/configmaps/app-env.yaml
umar8hassan: Narrowing the domain moves this rather than settling it: `learn-ai` serves from `ai.learn.<domain>`, which is inside the new scope, and it sets no cookie name or domain of its own, so a browser there still gets two `csrftoken` cookies. Setting `CSRF_COOKIE_NAME` and `SESSION_COOKIE_NAME` gives each app its own cookie, which is what `MITOL_COOKIE_NAME` already does for the cookie this comment says collides with nothing.

local-dev/infra/modules/keycloak.py
umar8hassan, line 858: `realm_keys` is already fetched at line 633 and read at line 645. This repeats the same lookup, and passing it through `Output.all()` turns the typed keys into plain dictionaries, which is the only reason `_first_public_key` exists. Reusing the existing value and its `.apply` removes both the second lookup and the helper. `_first_public_key` also returns an empty string when the list is empty, which would write an empty public key into the secret and fail login with nothing pointing at the cause.

local-dev/apps/ocw-studio/configmaps/nginx.yaml
umar8hassan, line 78: These three server blocks are the same apart from the port and the bucket name, so any later fix has to be made three times. A `map` from `$server_port` to the bucket name, with one server block listening on all three ports, does the same job.

local-dev/EXTENDING.md
umar8hassan: Step 4 says to call `_make_oidc_secret()`, but the ocw-studio client added here writes its own secret instead, because that helper produces the keys APISIX reads and this app needs `SOCIAL_AUTH_KEYCLOAK_*`. I compared the two on a running cluster and they have no keys in common. Since this PR already revises steps 2 and 3 for the optional-app pattern, step 4 should say which of the two shapes to use and when.

umar8hassan (approved):
mostly Looks Good 👍 

I couldn't test the published website due to local issues but i have verified that Studio and concourse run fine and trigger the pipeline.

most of my commenst are nits.

PR #5888, opened 2026-09-16
fix(traefik): give Pulumi the chart CRDs so upgrades actually apply

Ardiea:
No GitHub issue. witan task `tk-traefik-crds-frozen-by-helm-s-install-only-crds--4dfd71`.

Stacked on #5887, which adds the helper. GitHub will retarget this to `main` once that merges.

Pulumi now owns the 25 CRDs the traefik chart ships in `crds/`, instead of Helm writing them once at install and never again.

- Applies them via `adopt_helm_chart_crds` ahead of the release.
- Flips the release's `skip_crds` from `False` to `True` — it only ever affected install, which is the bug.

`Middleware` is the reason this one matters. [`components/services/traefik.py:30-31`](https://github.com/mitodl/ol-infrastructure/blob/main/src/ol_infrastructure/components/services/traefik.py#L30-L31) creates `traefik.io/v1alpha1` Middleware resources, so anything in the missing schema — `spec.compress.encodings`, the `spec.encodedCharacters` block, `spec.errors.errorRequestHeaders`, `spec.errors.service.middlewares` — is accepted by Pulumi and by the API server and then silently dropped by schema pruning.

No CRD lost a served apiVersion, so nothing is broken today. This is the trap, not active breakage.

<details>
<br>

**Measured drift**, live CRDs vs chart 41.5.0, counted as schema property paths missing from the live copy, across all 12 EKS clusters:

| CRD | missing paths (range across 12) | clusters affected |
|---|---|---|
| `traefikservices.traefik.io` | 100–161 | 12/12 |
| `middlewares.traefik.io` | 19–50 | 12/12 |
| `ingressroutes.traefik.io` | 4–18 | 12/12 |

The spread is per-cluster because the CRDs date from different first-install times; residential-* are the low end.

**Pre-flight run before landing.** The only way this apply fails hard is if a cluster has a version in `status.storedVersions` that chart 41.5.0 no longer serves — the API server rejects that outright and it would need a storage migration rather than an apply. Checked all 25 CRDs across all 12 clusters: zero such cases.

Five `hub.traefik.io` CRDs are absent on most clusters and will show as creates rather than adopts. `upsert_existing_objects` creates as readily as it adopts, so this needs no special handling.

**The `hub.traefik.io` set rides along.** It is Traefik Hub, commercial and unused here, but it lives in the same `crds/` directory. Taking it was the lower-risk option: leaving it frozen would preserve exactly this bug for whenever it does get used. Filtering it out later is a one-line `include=`.

**These are not the Gateway API CRDs.** Those are installed separately at [`traefik.py:67`](https://github.com/mitodl/ol-infrastructure/blob/main/src/ol_infrastructure/infrastructure/aws/eks/traefik.py#L67) from the upstream experimental bundle and are untouched here. The two sets do not overlap.

</details>

### How can this be tested?

`pulumi preview --stack operations.CI` from `src/ol_infrastructure/infrastructure/aws/eks`:

```
Resources:
    + 27 to create
    ~ 1 to update
    28 changes. 153 unchanged
```

27 = 25 CRDs + the ConfigGroup + the scoped provider. The single update is the release, diffing on `~skipCrds` alone — no collateral churn.

Creates rather than adopts is expected and not a sign the adoption is wrong: Pulumi holds no state for these objects so it plans a create, and `upsert_existing_objects` does the adoption at apply time. That path was verified by reproducing a frozen CRD on a throwaway k3d cluster during the starrocks fix (d703e40ef); it has not yet been exercised against a real cluster here, which is what the CI apply will confirm.

After applying to CI, the check that this actually worked:

```
kubectl --context operations-ci get crd middlewares.traefik.io \
  -o jsonpath='{.spec.versions[0].schema.openAPIV3Schema.properties.spec.properties.compress.properties.encodings}'
```

Empty before, populated after.

### Additional Context

Stage CI → QA → Production.

Merge #5887 first.


Ardiea:

PR #5890, opened 2026-09-16
fix(eks): give Pulumi the load balancer controller CRDs

Ardiea:
No GitHub issue. witan task `tk-aws-load-balancer-controller-crds-frozen-by-helm-300fd5`.

Stacked on #5887, which adds the helper. GitHub will retarget this to `main` once that merges.

Pulumi now owns the 7 CRDs the aws-load-balancer-controller chart ships in `crds/crds.yaml` and `crds/gateway-crds.yaml`, instead of Helm writing them once at install and never again.

- Applies them via `adopt_helm_chart_crds` ahead of the release.
- Adds `skip_crds=True` to the release.

Of the three charts with this bug, this is the one closest to active degradation rather than a latent trap. `TargetGroupBinding` is written by the **controller itself**, not by this repo, and the frozen schema is missing the entire `status.conditions[]` block. A controller writing `status.conditions` against a schema that has no such property gets it pruned — which quietly removes the health/readiness signal the newer controller reports.

<details>
<br>

**Measured drift**, live CRDs vs chart 3.5.0, counted as schema property paths missing from the live copy, across all 12 EKS clusters:

| CRD | missing paths (range across 12) | clusters affected |
|---|---|---|
| `loadbalancerconfigurations.gateway.k8s.aws` | 59–64 | 12/12 |
| `targetgroupbindings.elbv2.k8s.aws` | 7–8 | 12/12 |

On `targetgroupbindings` specifically, every one of the 12 clusters is missing the whole `v1beta1.status.conditions` block — `lastTransitionTime`, `message`, `observedGeneration`, `reason`, `status`, `type` — and nine are additionally missing `v1beta1.spec.targetGroupProtocol`.

**One claim I could not verify, stated plainly:** whether the running v3.5.0 controller actually populates `status.conditions` today. It cannot be read off a live cluster, because the frozen schema prunes the field before it is ever persisted — absence of the data is what the bug predicts either way. So the schema gap is measured fact; "the controller is actively losing a health signal" remains inference from the chart shipping the field. It does not change the fix, only its urgency, and it is worth settling before sequencing the rollout.

**Three different CRD sets behind one identical controller image.** All 12 clusters run `public.ecr.aws/eks/aws-load-balancer-controller:v3.5.0`, yet:

- `applications-ci` and `data-qa` are missing `globalaccelerators.aga.k8s.aws` and `listenerruleconfigurations.gateway.k8s.aws` entirely.
- `albtargetcontrolconfigs.elbv2.k8s.aws` exists only on `residential-qa` and `residential-production`.

An absent CRD fails loudly at admission ("no matches for kind") rather than pruning silently, so it is less dangerous than the `status.conditions` case — but it means a manifest that applies cleanly in `applications-qa` fails in `applications-ci`, which is backwards from how CI is supposed to de-risk QA and production. Adoption fixes this, since `upsert_existing_objects` creates as readily as it adopts.

**`albtargetcontrolconfigs` was checked before converging**, to be sure "make all 12 match" would not spread a dead CRD. Chart 3.5.0 does still ship it, so it is a legitimate CRD that ten clusters are missing rather than an orphan of the class recently cleaned up on applications-ci. It gets created on the other ten.

**Pre-flight run before landing.** For all 7 CRDs across all 12 clusters, every version in `status.storedVersions` is still served by chart 3.5.0 — no stored-version drops, no storage migration needed.

</details>

### How can this be tested?

`pulumi preview --stack operations.CI` from `src/ol_infrastructure/infrastructure/aws/eks`:

```
Resources:
    + 9 to create
    ~ 1 to update
```

9 = 7 CRDs + the ConfigGroup + the scoped provider. The single update is the release, diffing on `~skipCrds` alone — no collateral churn.

Creates rather than adopts is expected: Pulumi holds no state for these objects, and `upsert_existing_objects` does the adoption at apply time. On `applications-ci` and `data-qa` the creates are genuine, since those clusters are missing CRDs outright.

After applying to CI, the check that this actually worked:

```
kubectl --context operations-ci get crd targetgroupbindings.elbv2.k8s.aws \
  -o jsonpath='{.spec.versions[?(@.name=="v1beta1")].schema.openAPIV3Schema.properties.status.properties.conditions}'
```

Empty before, populated after.

### Additional Context

This chart runs on every cluster and fronts every load balancer, so stage CI → QA → Production and give each stage time to settle.

Merge #5887 first.

The controller's `mservice` and `mtargetgroupbinding` webhooks are `failurePolicy: Fail` with no namespace selector, so this deployment going down blocks Service and TargetGroupBinding admission cluster-wide. Nothing here touches the Deployment — the release diff is `skipCrds` only — but it is the reason to watch the rollout rather than fire and forget.


Ardiea:

Issue #5592, opened 2026-08-24
Create End To End Simulated User Testing (Canary) Pipeline for MIT Learn


feoh:
### Description/Context

Create an end-to-end simulated user ("canary") testing pipeline for MIT Learn
that exercises the site's main user happy path.

All work should be done with a view towards branching this out by creating
pipelines for our other user-facing products.

### Plan/Design

- Use Claude to mine OTEL traces from production to map the most common paths taken through the MIT Learn application and write playwright tests accordingly
- Build the pipeline skeleton that just logs into the Learn website and clicks a
  button/tab.
- Model the flow outlined above in the pipeline.
- Pipeline should trigger after a new release to RC and also periodically on a schedule. Failures should block releases from going to production
- Build a SKILL to enable running the end to end tests in the local development environment
- Build another SKILL to add new tests to the end to end testing pipeline

feoh:
## Progress + findings (2026-09-01)

PR #5707 settles where canary specs live and lands a working Playwright harness. Everything below was measured, not assumed.

**Decision: specs live in this repo**, at `src/ol_concourse/pipelines/canaries/`, run against the stock `mcr.microsoft.com/playwright` image after a pinned `npm ci`. No canary image is built. Recorded as ADR 0011.

I initially chose a dedicated repo (`mitodl/end-user-test-canaries`, since created) and reversed it, because two arguments for separating didn't survive checking:

- *"Don't give spec authors commits in the repo that deploys production"* — `odl-engineering` already has `maintain` here.
- *"Spec edits become commits in the production-deploying repo"* — Concourse path-watching already scopes that, same as `PULUMI_WATCHED_PATHS`.

The deciding factor for keeping them together: `package.json`'s pin is the single source of truth and the pipeline **derives** the image tag from it, so version drift becomes impossible rather than merely guarded. That repo is now unused and should be archived or deleted.

**Playwright packaging facts:**

- `@playwright/test` is *not* installed globally in the official image (`/usr/lib/node_modules` has only `corepack`, `npm`, `yarn`), so an install step is mandatory — there is no zero-install path.
- With a tiny dependency set that install is 6 packages in ~400 ms and downloads no browsers, which is why building a custom canary image isn't justified.
- Runner and image versions **cannot** drift. Pinning 1.58.1 against the `v1.62.1-noble` image fails with `Executable doesn't exist at /ms-playwright/chromium_headless_shell-1208/...`, naming neither version.
[...]

PR #5946, opened 2026-09-18
feat(canaries): add a mitxonline canary for the staff dashboard

blarghmatey:

Part of https://github.com/mitodl/ol-infrastructure/issues/5592

mitxonline is moving its staff dashboard from Refine 3 to Refine 4, and nothing checks the deployed dashboard on RC. This adds `mitxonline` as a canary property against https://rc.mitxonline.mit.edu, so `canary-meta` will set a `canary-mitxonline` pipeline once this merges.

- Two anonymous journeys. `/staff-dashboard/` must land on the Keycloak email screen with MITx Online's `/login/` as the `redirect_uri`. A deep route (`/staff-dashboard/flexible_pricing`) must load the app, which requests `/api/v0/users/me` and sends the visitor to sign in. The first is a server-side redirect, so only the second fails when the bundle is broken.
- A read-only staff journey: the dashboard heading, the Flexible Pricing menu item, and the Flexible Pricing list rendering. It never clicks Create or Save, and there are no discount journeys because those need superuser.
- The staff journey signs in as the existing mit-learn canary account (`credential_secret="canary_mit_learn"`). That account has been granted `is_staff` (not superuser) on RC MITx Online.
- The Keycloak half of mit-learn's sign-in (the positive password-screen check and the no-retry guard against the realm's permanent lockout) moves to `specs/shared/` in its own commit, so both properties use one copy of it.

<details>
<br>

- MITx Online's homepage on RC redirects to rc.learn.mit.edu, so the staff sign-in starts at `/staff-dashboard/` instead of the homepage. MIT Learn's session is not MITx Online's.
- The Flexible Pricing list shows learners' names, emails and incomes, and failure artifacts are uploaded to `s3://ol-eng-artifacts/canary-results/`. Turning off trace, screenshot and video isn't enough on its own, because on any failure Playwright 1.63 writes an ARIA snapshot of the page into `error-context.md`. That snapshot has two sources. A failed assertion carries one in its matcher result, and the page fixture's teardown takes one unless `PLAYWRIGHT_NO_COPY_PROMPT` is set. So the staff spec rethrows each failed assertion as a plain error that names the step, and it sets `PLAYWRIGHT_NO_COPY_PROMPT` for its own file only.
- `signInThroughOlapps` takes a `startLogin` callback that gets from the property's pages to Keycloak and returns the application origin. mit-learn keeps its marker file name (`mit-learn-canary-credential-rejected`) and its onboarding check.
- `specs/mitxonline/README.md` records the measured redirect chain and the account decision.

</details>

### How can this be tested?

I ran the anonymous spec against RC in the pinned image (`mcr.microsoft.com/playwright:v1.63.0-noble`). It passed twice in Chromium and once in WebKit:

```
cd src/ol_concourse/pipelines/canaries
docker run --rm --ipc=host -v "$PWD":/specs -w /specs \
  -e CANARY_BASE_URL=https://rc.mitxonline.mit.edu \
  mcr.microsoft.com/playwright:v1.63.0-noble \
  bash -c 'npm ci && npx playwright test specs/mitxonline/staff-dashboard-anonymous.spec.ts --project=chromium --repeat-each=2 && npx playwright test specs/mitxonline/staff-dashboard-anonymous.spec.ts --project=webkit'
```

As a negative check, a throwaway copy of the deep-route journey that aborted every `.js` request failed at the `users/me` wait, so that journey really does depend on the bundle.

For the PII handling, I ran a copy of the staff spec against a routed fake page that had a sentinel learner row, with the import swapped so it needed no sign-in. It failed at the menu-item step. The sentinel was in no report or artifact file, apart from `error-context.md` quoting the injected route line from the copied test source. Before the fix, the same kind of failure put the row into `error-context.md` as a page snapshot.

Also run: `npm run typecheck`, `npx playwright test --list` (all six tests load, including mit-learn's signed-in ones through the moved helpers), `specs/mit-learn/homepage.spec.ts` against rc.learn.mit.edu, `uv run pytest tests/ol_concourse`, and pre-commit on the changed files. `pipeline.py mitxonline` renders a job that runs all of `specs/mitxonline` in Chromium, with `CANARY_USER_EMAIL` and `CANARY_USER_PASSWORD` from `((canary_mit_learn.email))` and `((canary_mit_learn.password))`. `meta.py` renders `create-mitxonline-pipeline`.

The signed-in journeys have not run against RC. That needs the live credential, and a mistake there counts toward the account's permanent lockout. So the staff journey's first real run will be the first `canary-mitxonline` build after merge. The first `canary-mit-learn` build after merge is likewise what shows the moved sign-in still works. Both should be watched. A red build that names a rejected credential means both pipelines should be paused before anything else, because of the lockout note below.

### Additional Context

The staff journey shares the mit-learn canary account on purpose, and the account was granted `is_staff` for this. The cost is lockout exposure. Both pipelines submit the same password every 10 minutes, and the rejected-credential marker only covers retries within one run. So a drifted password reaches the realm's `failureFactor=10` twice as fast, and the permanent disable arrives in about one hour instead of about two. Rotation is still one update: Keycloak plus the single SOPS value `infrastructure/canary_mit_learn`. Both pipelines are in the same Concourse team and resolve `((canary_mit_learn.*))` to the same Vault path. There is no pipeline-scoped copy, and the default lookup templates are unchanged.

`is_staff` lives on the MITx Online user record in RC, not in Keycloak or Pulumi. An RC database restore that drops it would turn the staff journey red at the dashboard heading.

https://claude.ai/code/session_01VcbiSRu3CsE4uyfr6Lqjdr


feoh:
Let's walk before we try to run :) Closing.

PR #5966, opened 2026-09-21
Build api.learn's two shared plugin configs from one plugin list

blarghmatey:
N/A. Stacked on #4759, which is where the second shared plugin config comes from.

#4759 gives `api.learn.mit.edu` a second shared plugin config so the `browser-*` routes can carry rate limiting that the Fastly-fronted routes must not. Both configs are written out by hand, so every non-rate-limiting plugin has to be repeated in both, and a one-sided edit changes behaviour by request `Origin` with nothing in review or CI to catch it.

That has already happened twice on this host:

1. #4759 as originally written duplicated `stale_session_cookie_cleanup_plugin` into the browser config but not `oidc_gateway_pre_function_plugin`, so a login callback arriving with an `Origin` header would have kept 500ing while the identical request without one got recovered. Fixed in 4fec587ca.
2. Rebasing #4759 onto main after #5417 auto-merged the recovery attachment out of the base list and into the browser one, because both lists ended with the same three lines of context. The full suite still passed, because nothing asserted on the base config's plugin list.

This adds `ol_apisix_shared_plugins_variants()`, which renders one `OLApisixSharedPluginsConfig` as several configs on a host. `OLApisixSharedPluginsVariant` carries only the fields a variant may legitimately differ on (its Pulumi resource name, `resource_suffix`, `enable_rate_limiting`) and sets `extra="forbid"`, so a per-variant `plugins` list is a `ValidationError` rather than a silent divergence. Going the other way, the factory raises if the shared config sets `resource_suffix` or `enable_rate_limiting`: converting a single-config application means moving `enable_rate_limiting` onto a variant, and left behind it would silently take rate limiting off every route on the host.

Each variant supplies its own Pulumi resource name rather than deriving one, because renaming an existing component replaces it and every child CRD under it.

Also drops the base config's explicit `enable_gzip=True`. The component default has been unconditionally `True` since the Production soak gate was retired, so the override rendered nothing extra and its comment described a per-stack default that no longer exists.

### How can this be tested?
This is a pure refactor, verified rather than asserted: `pulumi preview -s CI --json` before and after gives the same 163 resources, the same 3 creates / 7 updates / 1 replace / 152 same, and zero resources with a differing input.

Tests in `tests/ol_infrastructure/components/services/test_apisix.py`:

- both variants render the same plugin names apart from `limit-conn`/`limit-req`, asserted separately on the v2 `ApisixPluginConfig` and the v1alpha1 `PluginConfig` since those are built by different code paths, and each asserting the rate-limit plugins are present to be filtered rather than letting the filter become a no-op
- a duplicate `resource_suffix` raises before two CRDs collide on one `metadata.name`
- `resource_suffix` or `enable_rate_limiting` on the shared config raises
- an AST check over `src/ol_infrastructure/applications/` groups `OLApisixSharedPlugins` calls by `application_name` and fails if one application builds more than one by hand

The AST check is the only one that can carry the invariant at the application level; the render tests share a plugin list by construction, so they can only catch the component introducing asymmetry, never an application doing it. It groups by host rather than by file so that splitting one host's two configs into sibling modules is not a way out, and so that two configs for two different applications stay legal (`edxapp` and `meilisearch` are that case today). It resolves aliased imports and attribute-style calls, reports an unparseable module instead of raising, and asserts it actually scanned modules and found calls, so a moved test file or a renamed class cannot leave it permanently green.

I confirmed each of those by constructing the case and watching the check fail, rather than by reading it: same host across two files fails, aliased and attribute-style calls fail, two different hosts in one file stays green, an unparseable file is reported.

`ruff format`, `ruff check`, `mypy` and `pre-commit` pass on the changed files with no new findings over the baseline; the full suite is 1301 passed.

Two things I left alone and would rather name than bury: `model_copy(update=...)` is shallow, so all variants share one `plugins` list object. The component only reads that list, and `deep=True` would deep-copy any Pulumi `Output` a caller put in it, which is the worse hazard, so it is commented rather than changed. `update` also bypasses validation, which is bounded by the fact that the only values passed through it come from the already-validated variant model.

https://claude.ai/code/session_01MFSSEUvubSr23Fi3aFYwJz


feoh (approved):
Claude analyzed for correctness and blast radius and this checks out 100%. Approved!

PR #5973, opened 2026-09-21
feat(canaries): trigger the mit-learn canary on RC deploy, not just on a schedule

feoh:

Part of https://github.com/mitodl/ol-infrastructure/issues/5592

Stacked on #5972 — base is `canary-green-run-flake-evidence`, and this will
retarget to `main` when that merges. The `results.json` record this writes
`releaseRef` into comes from that PR.

Issue #5592 asked for a canary that runs "after a new release to RC and also
periodically on a schedule". Only the schedule half existed.

- Adds a deploy trigger as a **second trigger on the same job**, so a release to
  `rc.learn.mit.edu` runs the journeys immediately instead of up to 10 minutes later.
- Makes a canary run attributable: the release calver becomes the Concourse
  **resource version**, so it shows up in the build's own inputs, and it is recorded
  into the retained `results.json` as `config.metadata.releaseRef`.
- Keeps the schedule. Breakage that no deploy caused is the other half of a canary's job.
- Producer side is **opt-in per app** and enabled only for `mit-learn`. The other eight
  `k8s_apps` pipelines render byte-identical to `main`.

Onboarding the next property stays two fields, not new hand-written jobs:
`CanaryParams.deploy_trigger` here and `AppPipelineParams.publish_rc_deploy_marker`
on the app.

<details>
<br>

**Why a marker object in S3.** Concourse's `passed` constraint is pipeline-local and
there is no trigger-another-pipeline step, so a canary in `canary-mit-learn` can only
observe a deploy in `mit-learn-pipeline` through a resource both declare. The RC deploy
job now writes `s3://ol-eng-artifacts/deploy-markers/<app>/RC/<version>.json`, and the
canary watches that prefix with an `s3` resource.

The version is in the **object key**, not just the body, because that is where the s3
resource reads a version from. That is what makes the release calver the resource
version — a marker carrying the version only in its body would trigger builds that
could not be attributed to anything, which is exactly what the downstream
production-gating work needs.

**Alternatives rejected, with reasons:**

- *The app's GitHub Deployment.* The natural fit, but
  [`github_deployment()`](https://github.com/mitodl/ol-concourse/blob/main/src/ol_concourse/lib/resources.py)
  is a put-only resource pinned to `check_every: never`, so it emits no versions to
  trigger on. It is also only created by the release-resource pipeline shape, and
  `mit-learn` is not on it — as of this writing only `ol-analytics-api` sets
  `AppRegistration.release_resource_workflow`
  ([apps.py](https://github.com/mitodl/ol-infrastructure/blob/main/src/bridge/settings/apps.py)).
  There is no GitHub Deployment for mit-learn to watch.
- *The `mitodl/mit-learn-app` registry image.* Triggers when the image is **built**,
  which is before it is deployed. The canary would test the previous release and
  attribute it to the new one — worse than not triggering.
- *A `fly trigger-job` task in the deploy pipeline.* Needs Concourse credentials in the
  deploy pipeline and produces a build with no input identifying what it tested.

**One shared contract.** The key layout lives once in
`src/ol_concourse/pipelines/deploy_markers.py`; both ends build the resource from
`deploy_marker_resource()`. A producer writing keys the consumer's `regexp` does not
match fails silently at *both* ends — the deploy stays green and the canary quietly
stops deploy-triggering while still passing on its schedule. Both meta pipelines now
watch that file so a layout change re-renders them together.

**Deliberate choices worth reviewing:**

- **`initial_path` on the marker resource.** A `get` on a resource that has never had a
  version never becomes schedulable. Since both triggers feed one job, that would have
  stopped the *scheduled* runs too. The synthetic `0.0.0.0` initial version means the
  canary is degraded-but-running if the producer half is ever missing.
- **One job with `max_in_flight=1`, not two jobs plus a `serial_group`.** A deploy
  landing mid-run queues behind the in-flight run instead of interleaving on the live
  property, and the run history (including the flake record) stays in one place.
- **Marker `put` ordered after the Fastly purge.** A canary that woke before the edge
  cache was purged would test the previous release through Fastly and report it against
  the new version.
- **Task output named `rc-deploy-marker-file`, not the resource name.** A `put` step and
  a task output share one artifact namespace; identical names happen to work only
  because `no_get` suppresses the implicit get.
- **No IAM change.** Checked at the source rather than assumed: the s3 resource only
  sends `x-amz-acl` when an `acl` param is set
  ([aws-sdk-go-v2 `serializers.go`](https://github.com/aws/aws-sdk-go-v2/blob/main/service/s3/serializers.go),
  `if len(v.ACL) > 0`), so `s3:PutObjectAcl` is not required. The operations worker role
  already has `s3:PutObject`/`GetObject*`/`ListBucket*` on `ol-eng-artifacts`.

</details>

### How can this be tested?

What I ran:

- **Blast radius.** Rendered all nine `k8s_apps` pipelines on this branch and on a
  clean `origin/main` worktree and diffed them. Eight are byte-identical; only
  `mit-learn` changes, purely additively (one resource, two steps appended to the QA
  job). All nine pass `fly validate-pipeline`, as does the rendered
  `canary-mit-learn`.
- **The marker script, in a real container.** Ran the generated `sh` script in
  `alpine:latest`: with `RELEASE_VERSION=2026.9.21.1` it writes valid JSON to
  `rc-deploy-marker-file/2026.9.21.1.json`; with the var empty it exits 1 rather than
  writing a `.json` file no regexp would match.
- **The regexp, both directions.** `deploy-markers/mit-learn/RC/([0-9.]+)\.json`
  captures `2026.9.21.1` and the `0.0.0.0` initial path, and rejects `latest.json`,
  `.json`, `…json.bak`, and another app's prefix.
- **Release attribution end to end in Playwright.** Ran
  `specs/mit-learn/homepage.spec.ts` against `https://rc.learn.mit.edu` with
  `CANARY_RELEASE_REF=2026.9.21.1` set and unset. `results.json` came back with
  `config.metadata.releaseRef` of `"2026.9.21.1"` and `"unknown"` respectively.
- `uv run pytest tests/` (1368 passed), `pre-commit` on the touched files, `mypy` clean
  on all four Python files, and `npm run typecheck`.

To validate as a reviewer:

```bash
# Only mit-learn should differ from main.
for app in learn-ai micromasters mit-learn mit-learn-nextjs mitxonline \
           ocw-studio odl-video-service ol-analytics-api xpro; do
  uv run python src/ol_concourse/pipelines/infrastructure/k8s_apps/pipeline.py "$app" >/dev/null
  mv definition.json "/tmp/$app.json"
done
```

### Additional Context

**Not verified, and cannot be until this is set:** the live path — a real RC deploy
writing a marker, and that marker triggering a canary build. Nothing here has run in
Concourse. The first real `mit-learn` RC deploy after this merges is the actual test,
and the thing to watch is that `canary-mit-learn` picks up a `deploy-marker` version
that is the release calver rather than `0.0.0.0`.

Until the first real deploy, the canary will report `0.0.0.0` as the release it tested.
That is the initial version working as intended, not a bug.

This does **not** add gating, metrics, or alerting. Concourse build status remains the
sole canary result, per the standing decision in #5592. Blocking promotion to production
on a failing canary is a separate follow-up that this unblocks, since it needs runs that
can name the release they covered.


feoh:
Closing because we have decided to keep the canary time-scheduled only rather than triggering it from new mit-learn RC deployments. The independent per-run evidence work in #5972 remains open.

Issue #5562, opened 2026-08-21
Rate-limit NextJS SSR origin traffic (Traefik) for MIT Learn


blarghmatey:
### Description/Context
PR #4759 added per-client-IP rate limiting to the MIT Learn APISIX gateway, scoped to direct browser traffic only (matched on the `Origin` header). It does not, and by design cannot, limit a second class of load: bots hitting `learn.mit.edu` directly, causing Fastly cache misses that trigger NextJS server-side rendering, which in turn calls `api.learn.mit.edu`. That SSR-to-API traffic carries no `Origin` header for APISIX to key on.

Raised by @ChristopherChudzicki: https://github.com/mitodl/ol-infrastructure/pull/4759#issuecomment-5363603962

An earlier version of PR #4759 included `OLEKSGatewayRateLimitConfig`, a Traefik middleware-based rate limit on the NextJS gateway itself, which would cover this case by limiting bot traffic at the NextJS origin. It was dropped from that PR's final scope.

### Plan/Design
Restore Traefik middleware-based rate limiting on the `mit-learn-nextjs` gateway (`OLEKSGatewayRateLimitConfig`, see the removed implementation: https://github.com/mitodl/ol-infrastructure/commit/1c6f0e068d9fbc3a58ac472733ce0856db023bed), keyed on client IP.

Open question to resolve before implementing: which header carries the real end-client IP at the NextJS origin. Fastly always sets `Fastly-Client-IP` itself, which is likely the safer `sourceCriterion`. The Fastly service currently sets `xff="leave"`, so it's not confirmed that `X-Forwarded-For` at any fixed depth reliably resolves to the real client rather than a Fastly edge address -- verify against the live Fastly VCL config before picking a header.

PR #5977, opened 2026-09-21
Exclude the mit-learn SSR client from api.learn rate limiting on purpose

blarghmatey:
N/A. Third in the stack: #4759 adds the rate limiting, #5966 factors the shared plugin configs, this makes one of #4759's routing assumptions explicit.

The `browser-*` routes #4759 adds carry a per-client-IP rate limit sized for one browser (`limit-req` 50 r/s, `limit-conn` 100, keyed on `remote_addr`), and they are selected by a single predicate: an `Origin` matching the site. A first-party server-side caller is the wrong shape for that limit, because it aggregates every end user behind one address, so a browser-sized bucket throttles the whole population at once.

The Next.js SSR layer is that caller. It reaches api.learn as `axios/1.12.2` (`frontends/api/package.json`) from four AWS egress addresses, and on 2026-09-21 the busiest sustained 371-676 req/min, i.e. 11.3 req/s against the 50 req/s bucket:

| address | req/min | peak |
| --- | --- | --- |
| 34.193.218.132 | 371-676 | 11.3 req/s |
| 54.83.156.247 | 172-346 | |
| 52.203.147.167 | 158-385 | |
| 98.82.252.213 | 166-351 | |

It misses the rate-limited routes today only because Node axios sends no `Origin` header. Nothing records that, so a change to the SSR HTTP client, or anything that forwards browser headers, would move 11 req/s of aggregate traffic onto a per-browser bucket with no signal in review.

This adds a negative `User-Agent` clause beside the `Origin` match, through a new `browser_traffic_match_exprs()` helper, so the exclusion is the intent rather than a side effect of the client library.

**Why not the alternatives.** Both obvious options are wrong here, which is why they were measured first rather than reasoned about:

- Requiring a browser-shaped `Sec-Fetch-Site` does not work. Over 10 minutes it was absent on 55% of `Origin`-carrying requests (10,388 of 18,838) and *present* on the Amazon Bedrock crawler. It discriminates backwards.
- Requiring `^Mozilla/` would exempt crawlers along with SSR. DuckDuckBot, DuckAssistBot and an Amazon Bedrock knowledge base all replay our `Origin` (the last with a 1,466-byte cookie jar, sustaining 46-153 req/min from one address). Those are single actors that a per-IP limit *should* catch. Non-Mozilla is only 71 of 19,134 `Origin`-carrying requests (0.37%) and all of it is crawlers.

So the clause names the first-party client instead of excusing every non-browser.

**This adds a knob any client can turn**, not just a description of what already happens: anything sending a matching `User-Agent` opts itself out. That is not a regression for a non-browser caller, which could already opt out by omitting `Origin`, and dropping a header is easier than setting one. It does mean the clause is only as trustworthy as the limit it guards, which the component already describes as a blast-radius control rather than anti-abuse. The one genuinely new corner is browser-resident script: `Origin` is a forbidden request header so page JS cannot forge it, `User-Agent` no longer is. That needs XSS-equivalent access on learn.mit.edu, Chrome drops the header anyway, and it forces a CORS preflight that itself carries `Origin`.

`service_client_ua_regex` is required rather than defaulted, so a second host adopting the helper cannot inherit mit-learn's first-party client and silently exempt every stock-axios caller of its own.

### How can this be tested?
**Behaviour is unchanged today, at both levels.** `pulumi preview -s CI --json` touches only the two `ApisixRoute` resources, adding the clause to the three `browser-*` routes in each; no other resource differs and the change summary is unmoved at 3 creates / 7 updates / 1 replace / 152 same. At the traffic level, no request carrying the site `Origin` also carried an `axios` User-Agent in 60 minutes across 651,185 requests, so nothing in production moves off the rate-limited routes as a result of this.

**A request with no `User-Agent` at all still matches these routes**, rather than falling off or erroring. `exprs` are ANDed, so this was the one way the change could have broken real traffic. The ingress controller renders the clause as `["http_user_agent", "!", "~~", "^axios/"]`, and lua-resty-expr's `~~` comparator opens with `if not l_v then return false end`, which the negation turns into a pass. Bounded by measurement as well: across 30 minutes and 205,210 requests to this host exactly one had no User-Agent, and it was not carrying the site `Origin`. `RegexNotMatch` was confirmed against the op enum in the live v2 `ApisixRoute` CRD rather than assumed.

Tests assert the decision against real observed user agents rather than the shape of the dict: the SSR UA matches the exclusion, and the Chrome UA plus all three crawler UAs do not. They use `re.search`, not `re.match`, because APISIX matches with `ngx.re.find`, a substring search; `re.match` would silently supply an anchor the real matcher does not have and hide an unanchored value. I verified the anchor test has force by defeating the anchor with `^(.*)?axios/`, which still satisfies the `startswith("^")` check, and watching only that test fail.

`ruff`, `mypy` and `pre-commit` pass with no new findings over the baseline; full suite 1309 passed.

After deploy, the thing to watch is unchanged from #4759: any sustained non-zero `sum(rate(apisix_http_status{matched_host="api.learn.mit.edu", code="429"}[5m]))` means real traffic is hitting the limit. Worth noting that nothing currently alerts on that.

### Known limitation
This keys on axios's default `User-Agent`, which mit-learn does not set deliberately. Moving the SSR layer to native `fetch` or undici would lapse the exclusion again. The durable fix is for that client to send a stable first-party identifier, which is a mit-learn change and is tracked separately.

https://claude.ai/code/session_01MFSSEUvubSr23Fi3aFYwJz


blarghmatey:
Cross-reference for whoever reviews this: #5562 (Rate-limit NextJS SSR origin traffic (Traefik) for MIT Learn) covers the other half of this problem, and the two agree on the design rather than competing.

#5562's premise, in its own words, is that "SSR-to-API traffic carries no `Origin` header for APISIX to key on", so the SSR load gets limited at the NextJS origin with Traefik instead. This PR is the same fact from the other side: since SSR is deliberately outside the APISIX per-client-IP bucket, that exclusion should be stated rather than left resting on axios not sending an `Origin`.

  #5562  limit SSR-driven load at its source (Traefik, NextJS origin)
  #5977  keep SSR out of the APISIX per-IP browser bucket, on purpose

#5562 landing first would not remove the need for this. A Traefik limit bounds inbound bot traffic at the origin; it does not change which APISIX route the resulting SSR-to-API calls match, so the exclusion still has to hold on its own.

The known limitation in the description above (this keys on axios's default User-Agent, which mit-learn does not set deliberately) is tracked separately as a mit-learn change, with #5562 noted as required reading alongside it.

feoh (approved):

Ardiea (approved):

